PR #21, opened 2017-12-12
Enable kubectl logs

### LINKED ISSUE #2: kubectl logs support

ritazh:
Fixes #2

### DIFF
--- a/charts/virtual-kubelet/templates/deployment.yaml
+++ b/charts/virtual-kubelet/templates/deployment.yaml
@@ -15,29 +15,22 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         env:
         - name: AZURE_AUTH_LOCATION
-          value: /etc/virtual-kubelet/auth/credentials.json
+          value: /etc/virtual-kubelet/credentials.json
         - name: ACI_RESOURCE_GROUP
           value: {{ .Values.env.aciResourceGroup }}
         - name: ACI_REGION
           value: {{ default "westus" .Values.env.aciRegion }}
         - name: APISERVER_CERT_LOCATION
-          value: /etc/virtual-kubelet/apiservercert/cert.pem
+          value: /etc/virtual-kubelet/cert.pem
         - name: APISERVER_KEY_LOCATION
-          value: /etc/virtual-kubelet/apiserverkey/key.pem
+          value: /etc/virtual-kubelet/key.pem
         - name: VKUBELET_POD_IP
           valueFrom:
             fieldRef:
               fieldPath: status.podIP
         volumeMounts:
         - name: credentials
-          mountPath: "/etc/virtual-kubelet/auth"
-          readOnly: true
-        - name: apiservercert
-          mountPath: "/etc/virtual-kubelet/apiservercert"
-          readOnly: true
-        - name: apiserverkey
-          mountPath: "/etc/virtual-kubelet/apiserverkey"
-          readOnly: true
+          mountPath: "/etc/virtual-kubelet"
         command: ["virtual-kubelet"]
         args: ["--provider", "azure", "--namespace", "default", "--nodename", {{ default "virtual-kubelet" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
       volumes:
--- a/charts/virtual-kubelet/templates/deployment.yaml
+++ b/charts/virtual-kubelet/templates/deployment.yaml
@@ -39,9 +39,3 @@ spec:
       - name: credentials
         secret:
           secretName: {{ template "fullname" . }}
-      - name: apiservercert
-        secret:
-          secretName: {{ template "fullname" . }}
-      - name: apiserverkey
-        secret:
-          secretName: {{ template "fullname" . }}
--- a/charts/virtual-kubelet/templates/deployment.yaml
+++ b/charts/virtual-kubelet/templates/deployment.yaml
@@ -14,6 +14,8 @@ spec:
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         env:
+        - name: KUBELET_PORT
+          value: "10250"
         - name: AZURE_AUTH_LOCATION
           value: /etc/virtual-kubelet/credentials.json
         - name: ACI_RESOURCE_GROUP
--- a/charts/virtual-kubelet/templates/deployment.yaml
+++ b/charts/virtual-kubelet/templates/deployment.yaml
@@ -15,26 +15,38 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         env:
         - name: AZURE_AUTH_LOCATION
-          value: /etc/virtual-kubelet/credentials.json
+          value: /etc/virtual-kubelet/auth/credentials.json
         - name: ACI_RESOURCE_GROUP
           value: {{ .Values.env.aciResourceGroup }}
         - name: ACI_REGION
           value: {{ default "westus" .Values.env.aciRegion }}
-        - name: APISERVER_CERT
-          value: {{ .Values.env.apiserverCert | quote }}
-        - name: APISERVER_KEY
-          value: {{ .Values.env.apiserverKey | quote }}
+        - name: APISERVER_CERT_LOCATION
+          value: /etc/virtual-kubelet/apiservercert/cert.pem
+        - name: APISERVER_KEY_LOCATION
+          value: /etc/virtual-kubelet/apiserverkey/key.pem
         - name: VKUBELET_POD_IP
           valueFrom:
             fieldRef:
               fieldPath: status.podIP
         volumeMounts:
         - name: credentials
-          mountPath: "/etc/virtual-kubelet"
+          mountPath: "/etc/virtual-kubelet/auth"
+          readOnly: true
+        - name: apiservercert
+          mountPath: "/etc/virtual-kubelet/apiservercert"
+          readOnly: true
+        - name: apiserverkey
+          mountPath: "/etc/virtual-kubelet/apiserverkey"
           readOnly: true
         command: ["virtual-kubelet"]
         args: ["--provider", "azure", "--namespace", "default", "--nodename", {{ default "virtual-kubelet" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
       volumes:
       - name: credentials
+        secret:
+          secretName: {{ template "fullname" . }}
+      - name: apiservercert
+        secret:
+          secretName: {{ template "fullname" . }}
+      - name: apiserverkey
         secret:
           secretName: {{ template "fullname" . }}
--- a/charts/virtual-kubelet/templates/deployment.yaml
+++ b/charts/virtual-kubelet/templates/deployment.yaml
@@ -33,7 +33,7 @@ spec:
           mountPath: "/etc/virtual-kubelet"
           readOnly: true
         command: ["virtual-kubelet"]
-        args: ["--provider", "azure", "--namespace", "default", "--nodename", {{ default "aci-conn" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
+        args: ["--provider", "azure", "--namespace", "default", "--nodename", {{ default "virtual-kubelet" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
       volumes:
       - name: credentials
         secret:
--- a/charts/virtual-kubelet/templates/deployment.yaml
+++ b/charts/virtual-kubelet/templates/deployment.yaml
@@ -20,12 +20,20 @@ spec:
           value: {{ .Values.env.aciResourceGroup }}
         - name: ACI_REGION
           value: {{ default "westus" .Values.env.aciRegion }}
+        - name: APISERVER_CERT
+          value: {{ .Values.env.apiserverCert | quote }}
+        - name: APISERVER_KEY
+          value: {{ .Values.env.apiserverKey | quote }}
+        - name: VKUBELET_POD_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
         volumeMounts:
         - name: credentials
           mountPath: "/etc/virtual-kubelet"
           readOnly: true
         command: ["virtual-kubelet"]
-        args: ["--provider", "azure", "--nodename", {{ default "aci-conn" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
+        args: ["--provider", "azure", "--namespace", "default", "--nodename", {{ default "aci-conn" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
       volumes:
       - name: credentials
         secret:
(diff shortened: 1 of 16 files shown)


providers/azure/aci.go
```diff
@@ -31,6 +31,7 @@ type ACIProvider struct {
 	cpu             string
 	memory          string
 	pods            string
+	internalIP   	string
```
robbiezhang: nit: align with the other "string"

providers/azure/aci.go
```diff
@@ -193,6 +195,40 @@ func (p *ACIProvider) GetPod(namespace, name string) (*v1.Pod, error) {
...
+		if err != nil {
+			log.Println(err)
+			time.Sleep(5000 * time.Millisecond)
+		} else {
+			break
+			logContent = cLogs.Content
+			return logContent, nil
+		}
+	}
+	// create pod logs
```
robbiezhang: remove?

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -20,12 +20,20 @@ spec:
           value: {{ .Values.env.aciResourceGroup }}
         - name: ACI_REGION
           value: {{ default "westus" .Values.env.aciRegion }}
+        - name: APISERVER_CERT
+          value: {{ .Values.env.apiserverCert | quote }}
+        - name: APISERVER_KEY
+          value: {{ .Values.env.apiserverKey | quote }}
```
robbiezhang: As the apiserver key is a secret, can we mount the apiserver key and cert as secret volume?

providers/azure/aci.go
```diff
@@ -193,6 +195,40 @@ func (p *ACIProvider) GetPod(namespace, name string) (*v1.Pod, error) {
 	return containerGroupToPod(cg)
 }
 
+// GetPodLogs returns the logs of a pod by name that is running inside ACI.
+func (p *ACIProvider) GetPodLogs(namespace, name string) (string, error) {
+	logContent := ""
+	cg, err := p.aciClient.GetContainerGroup(p.resourceGroup, name)
```
robbiezhang: container group name should be "namespace-name"

vkubelet/provider.go
```diff
@@ -24,7 +24,10 @@ type Provider interface {
 	// GetPod retrieves a pod by name from the provider (can be cached).
 	GetPod(namespace, name string) (*v1.Pod, error)
 
-	// GetPodStatus retrievesthe status of a pod by name from the provider.
+	// GetPodLogs retrieves the logs of a pod by name from the provider.
+	GetPodLogs(namespace, name string) (string, error)
```
robbiezhang: namespace, podName, containerName string

ritazh:
hey @robbiezhang This is now ready for review.

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -15,18 +15,38 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         env:
         - name: AZURE_AUTH_LOCATION
-          value: /etc/virtual-kubelet/credentials.json
+          value: /etc/virtual-kubelet/auth/credentials.json
         - name: ACI_RESOURCE_GROUP
           value: {{ .Values.env.aciResourceGroup }}
         - name: ACI_REGION
           value: {{ default "westus" .Values.env.aciRegion }}
+        - name: APISERVER_CERT_LOCATION
```
robbiezhang: what's the requirement for this cert? how can the k8s admin generate the cert?

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -15,18 +15,38 @@ spec:
...
           readOnly: true
         command: ["virtual-kubelet"]
-        args: ["--provider", "azure", "--nodename", {{ default "aci-conn" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
+        args: ["--provider", "azure", "--namespace", "default", "--nodename", {{ default "virtual-kubelet" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
       volumes:
       - name: credentials
         secret:
-          secretName: {{ template "fullname" . }}
\ No newline at end of file
+          secretName: {{ template "fullname" . }}
```
robbiezhang: why do we create 3 volumes for the same secret? they are mounted in 3 pathes with the same content.

charts/virtual-kubelet/templates/secrets.yaml
```diff
@@ -4,4 +4,6 @@ metadata:
...
\ No newline at end of file
+  credentials.json: {{ printf "{ \"clientId\": \"%s\", \"clientSecret\": \"%s\", \"subscriptionId\": \"%s\", \"tenantId\": \"%s\", \"activeDirectoryEndpointUrl\": \"https://login.microsoftonline.com/\", \"resourceManagerEndpointUrl\": \"https://management.azure.com/\", \"activeDirectoryGraphResourceId\": \"https://graph.windows.net/\", \"sqlManagementEndpointUrl\": \"database.windows.net\", \"galleryEndpointUrl\": \"https://gallery.azure.com/\", \"managementEndpointUrl\": \"https://management.core.windows.net/\" }" (default "MISSING" .Values.env.azureClientId) (default "MISSING" .Values.env.azureClientKey) (default "MISSING" .Values.env.azureSubscriptionId) (default "MISSING" .Values.env.azureTenantId) | b64enc | quote }}
+  cert.pem: {{ printf "%s" (default "MISSING" .Values.env.apiserverCert) | quote }}
```
robbiezhang: no need to call "printf"

charts/virtual-kubelet/templates/secrets.yaml
```diff
@@ -4,4 +4,6 @@ metadata:
...
+  key.pem: {{ printf "%s" (default "MISSING" .Values.env.apiserverKey) | quote }}
```
robbiezhang: same here. "printf" is not necessary.

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -15,18 +15,38 @@ spec:
...
         - name: ACI_REGION
           value: {{ default "westus" .Values.env.aciRegion }}
+        - name: APISERVER_CERT_LOCATION
+          value: /etc/virtual-kubelet/apiservercert/cert.pem
+        - name: APISERVER_KEY_LOCATION
+          value: /etc/virtual-kubelet/apiserverkey/key.pem
+        - name: VKUBELET_POD_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
```
robbiezhang: 👍

charts/virtual-kubelet/templates/deployment.yaml
ritazh: There are 3 different files we need to mount. `credentials.json`, `cert.pem`, and `key.pem`.

robbiezhang: they are in the same secret. you just need to mount in one path, and you can find all 3 files in it.

vkubelet/apiserver.go
```diff
@@ -0,0 +1,43 @@
...
+	"strings"
+)
+var p Provider
+
+func ApiserverStart(provider Provider) error {
+	p = provider
+	http.HandleFunc("/", ApiServerHandler)
+	certFilePath := os.Getenv("APISERVER_CERT_LOCATION")
+	keyFilePath := os.Getenv("APISERVER_KEY_LOCATION")
+	err := http.ListenAndServeTLS(":10250", certFilePath, keyFilePath, nil)
```
robbiezhang: question: do we need to make the port configurable? it relates to some change in the node registration.

vkubelet/apiserver.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	return err
+}
+
+func ApiServerHandler(w http.ResponseWriter, req *http.Request) {
+	if req.Method == "GET" {
+		if strings.ContainsAny(req.RequestURI, "containerLogs" ) {
```
robbiezhang: not familiar with go, but this looks quite raw. @erikstmartin, is there any go package which helps to parse the URI and route to different webapi?

vkubelet/apiserver.go
ritazh: This is the default port kubelet serves on. `The port for the Kubelet to serve on. (default 10250)` https://kubernetes.io/docs/reference/generated/kubelet/

robbiezhang: Yes, 10250 is the default port. However, when you register the node, you can configure it to a different port by adding the following setting. 

  daemonEndpoints:
    kubeletEndpoint:
      Port: 10250

My question is whether we should allow the virtual-kubelet user to configure the port.

vkubelet/apiserver.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/http"
+	"os"
+	"strings"
+)
+var p Provider
+
+func ApiserverStart(provider Provider) error {
+	p = provider
+	http.HandleFunc("/", ApiServerHandler)
+	certFilePath := os.Getenv("APISERVER_CERT_LOCATION")
```
robbiezhang: for backward compatible, can we make sure the virtual-kubelet can still work (without exec logs support) if the environment variables are not set.

charts/virtual-kubelet/templates/deployment.yaml
ritazh: `openssl req -newkey rsa:4096 -new -nodes -x509 -days 3650 -keyout key.pem -out cert.pem` I will add this logic to the chart

vkubelet/apiserver.go
ritazh: ok got it. I will make this into an env var.

charts/virtual-kubelet/templates/deployment.yaml
ritazh: done

charts/virtual-kubelet/templates/deployment.yaml
robbiezhang: Don't quite understand. is it a self-signed cert? looks like it's used in the TLS. how does the ApiServer trust this cert?

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -14,19 +14,34 @@ spec:
...
         command: ["virtual-kubelet"]
-        args: ["--provider", "azure", "--nodename", {{ default "aci-conn" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
+        args: ["--provider", "azure", "--namespace", "default", "--nodename", {{ default "virtual-kubelet" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
       volumes:
       - name: credentials
         secret:
-          secretName: {{ template "fullname" . }}
\ No newline at end of file
+          secretName: {{ template "fullname" . }}
+      - name: apiservercert
```
robbiezhang: no longer needed.

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -14,19 +14,34 @@ spec:
...
+        secret:
+          secretName: {{ template "fullname" . }}
+      - name: apiserverkey
```
robbiezhang: no longer needed

providers/azure/aci.go
```diff
@@ -193,6 +195,37 @@ func (p *ACIProvider) GetPod(namespace, name string) (*v1.Pod, error) {
...
+		return logContent, err
+	}
+
+	if cg.Tags["NodeName"] != p.nodeName {
+		return logContent, nil
+	}
+	// get logs from cg
+	retry := 10
+	for i := 0; i < retry; i++ {
+		cLogs, err := p.aciClient.GetContainerLogs(p.resourceGroup, cg.Name, containerName, 10)
```
robbiezhang: The tail lines would be better to be an input parameter of the GetContainerLogs function. 
I assume it Is in the query string of the URI from ApiServer, correct?

robbiezhang: side question, what else parameters will be in the URI?

ritazh: > query string of the URI from ApiServer

yes

vkubelet/vkubelet.go
```diff
@@ -136,6 +139,7 @@ func (s *Server) registerNode() error {
 			Capacity:    s.provider.Capacity(),
 			Allocatable: s.provider.Capacity(),
 			Conditions:  s.provider.NodeConditions(),
+			Addresses:   s.provider.NodeAddresses(),
```
robbiezhang: guess you also need to set the Status.DaemonEndpoints.KubeletEndpoint.Port to "KUBELET_PORT" here.

ritazh: done

providers/azure/aci.go
ritazh: Just [two](https://github.com/ritazh/virtual-kubelet/blob/bcc5a33098f9a51839372e124e47eb928861a383/providers/azure/client/aci/logs.go#L17-L19)

Example:
https://management.azure.com/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ContainerInstance/containerGroups/{containerGroupName}/containers/{containerName}/logs?api-version=2017-10-01-preview&tail={tail}

vkubelet/apiserver.go
erikstmartin: https://golang.org/pkg/net/http/#Handle allows you to give it the full path you're looking for. We likely want to use this because there will be other endpoints and we can have different functions handle each route.

Also this current implementation would match a url that has containerLogs anywhere in it

providers/azure/aci.go
```diff
@@ -45,7 +47,7 @@ type AuthConfig struct {
 }
 
 // NewACIProvider creates a new ACIProvider.
-func NewACIProvider(config string, rm *manager.ResourceManager, nodeName, operatingSystem string) (*ACIProvider, error) {
+func NewACIProvider(config string, rm *manager.ResourceManager, nodeName, operatingSystem string, internalIP string, daemonEndpointPort int32) (*ACIProvider, error) {
```
erikstmartin: This parameter list is starting to get quite long. Maybe we should consider having things like ResourceManager passed in directly, but refactoring out all the configuration related things, (nodeName, OS, ip, daemon port, etc) out into like a ProviderConfig type. This will also prevent us from breaking all the providers when we add new params.

providers/azure/aci.go
```diff
@@ -193,6 +197,37 @@ func (p *ACIProvider) GetPod(namespace, name string) (*v1.Pod, error) {
 	return containerGroupToPod(cg)
 }
 
+// GetPodLogs returns the logs of a pod by name that is running inside ACI.
+func (p *ACIProvider) GetContainerLogs(namespace, podName, containerName string) (string, error) {
+	logContent := ""
+	cg, err := p.aciClient.GetContainerGroup(p.resourceGroup, fmt.Sprintf("%s-%s", namespace, podName))
+	if err != nil {
+		// Trap error for 404 and return gracefully
+		if strings.Contains(err.Error(), "ResourceNotFound") {
```
erikstmartin: We might want some shared values here so we don't have hard coded string matches. For example the net/http package has a bunch of constants defined https://golang.org/pkg/net/http/#pkg-constants so you can just do Status == http.StatusNotFound

providers/azure/aci.go
```diff
@@ -193,6 +197,37 @@ func (p *ACIProvider) GetPod(namespace, name string) (*v1.Pod, error) {
...
+			return logContent, nil
+		}
+		return logContent, err
+	}
+
+	if cg.Tags["NodeName"] != p.nodeName {
```
erikstmartin: I'm not sure what the k8s API Server expects here, but we may want to consider handing back an error, like a 404 or something. Getting empty logs could be misleading.

providers/azure/aci.go
ritazh: @erikstmartin do you mind if we do this in a separate PR so that this PR can focus on just enabling logging?

providers/azure/aci.go
robbiezhang: I mean parameters in the request query string from the ApiServer to kubelet. ACI only supports tail right now. I want to know the gap.

ritazh: ah ok. from what i can tell, these are there in ACI:
- `kubectl logs pod` :   `GET /containerLogs/{namespace}/{pod}/{container}`
- `kubectl logs pod --tails=2` :   `GET /containerLogs/{namespace}/{pod}/{container}?tailLines=2`

But the following options are not there in ACI
- `kubectl logs pod -f` :   `GET /containerLogs/{namespace}/{pod}/{container}?follow=true` 
- `kubectl logs pod --since=1m` :   `GET /containerLogs/{namespace}/{pod}/{container}?sinceSeconds=60`
- `kubectl logs pod -p=true` :   `GET /containerLogs/{namespace}/{pod}/{container}?previous=true`
- `kubectl logs pod --limit-bytes=1` :   `GET /containerLogs/{namespace}/{pod}/{container}?limitBytes=1`
- `kubectl logs pod --timestamps=true` :   `GET /containerLogs/{namespace}/{pod}/{container}?timestamps=true`

providers/azure/aci.go
ritazh: done

providers/azure/aci.go
ritazh: done

ritazh:
@erikstmartin @robbiezhang I have addressed all the comments. Can you please provide another review? Thanks!

robbiezhang:
LGTM

providers/azure/aci.go
robbiezhang: @erikstmartin are you fine with merge the PR and refactor the parameters later?

erikstmartin:
LGTM, merging

Issue #46, opened 2017-12-28
[Hyper] If the config file is not set, hyper provider will crash when pulling images


arapulido:
The AuthConfig is only set when there is a config file (and the env variables for key and secret are not set), but it is used by ensureImage nevertheless.

Steps to reproduce:

 * Set HYPER_ACCESS_KEY and HYPER_SECRET_KEY env variables
 * Create a pod using an image that you haven't pulled yet

Result:

The Hyper provider crashes:
```
/Users/ara/.kube/config
2017/12/28 11:50:21 Use AccessKey and SecretKey from HYPER_ACCESS_KEY and HYPER_SECRET_KEY
2017/12/28 11:50:21
 Host: tcp://eu-central-1.hyper.sh:443
 AccessKey: 4**********
 SecretKey: 4**********
 InstanceType: s2
2017/12/28 11:50:22 Node 'virtual-kubelet' with OS type 'Linux' registered
2017/12/28 11:50:22 receive GetPods
2017/12/28 11:50:22 found 0 pods
2017/12/28 11:50:37 receive GetPods
2017/12/28 11:50:37 found 0 pods
2017/12/28 11:50:37 Error retrieving pod 'nginx' from provider: Error: No such container: pod-nginx-nginx
2017/12/28 11:50:37 receive CreatePod "nginx"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1b48aa5]

goroutine 1 [running]:
github.com/virtual-kubelet/virtual-kubelet/providers/hypersh.(*HyperProvider).ensureImage(0xc42020d5c0, 0xc42045af80, 0xc, 0x0, 0x0)
	/Users/ara/go/src/github.com/virtual-kubelet/virtual-kubelet/providers/hypersh/util.go:337 +0x105
github.com/virtual-kubelet/virtual-kubelet/providers/hypersh.(*HyperProvider).CreatePod(0xc42020d5c0, 0xc420a0ea80, 0x0, 0x0)
	/Users/ara/go/src/github.com/virtual-kubelet/virtual-kubelet/providers/hypersh/hypersh.go:219 +0x652
github.com/virtual-kubelet/virtual-kubelet/vkubelet.(*Server).createPod(0xc4202104e0, 0xc420a0ea80, 0xc420979a40, 0x2)
[...]

PR #47, opened 2017-12-28
If the hyper config file is not set, the AuthConfig should be empty

arapulido:
This fixes issue #46 

When the hyper config file is not set, the AuthConfig should be empty to avoid a crash.


resouer:
lgtm, thanks for fixing this!

Issue #40, opened 2017-12-22
Pod scheduled at Windows kubelet should not mount any secret volume


sioakim:
[error.log](https://github.com/virtual-kubelet/virtual-kubelet/files/1583501/error.log)
I use the attached xml to deploy to Windows kubelet.

The windows kubelet should accept nodes for "beta.kubernetes.io/os": "windows"

From "describe node I get:"
Labels:             beta.kubernetes.io/os=windows

I get error:
No nodes are available that match all of the following predicates:: MatchNodeSelector (2), PodToleratesNodeTaints (1).

Full info is attached.

sioakim:
FYI - i just replaced the attached xml with nginx and Linux and worked instantly.

robbiezhang:
@sioakim, can you double check the Linux pod runs on the virtual-kubelet for Linux?

The virtual-kubelet node has the taints "azure.com/aci:NoSchedule". If you want to schedule a pod to the virtual-kubelet, you need to add the following tolerations to your pod spec:

  tolerations:
  - key: azure.com/aci
    effect: NoSchedule

Seems like the example is missing this section. I'll update it.

sioakim:
You are right - the linux one started on the k8s node. I had some other example apparently that had started it on the ACI.

Can you please provide an example for Windows as well?

The below doesn't work for example:

```
apiVersion: v1
kind: Pod
metadata:
  name: iis
spec:
  containers:
  - image: microsoft/iis
    imagePullPolicy: Always
    name: iis
    resources:
      requests:
        memory: 4G
        cpu: 2
    ports:
    - containerPort: 80
      name: http
      protocol: TCP
    - containerPort: 443
      name: https
  dnsPolicy: ClusterFirst
  nodeName: virtual-kubelet-aciconnector-win
  tolerations:
  - key: azure.com/aci
    effect: NoSchedule
```

sioakim:
Specifically the error for the above is:

2017/12/23 08:11:18 Error creating pod 'iis': api call to https://management.azure.com/subscriptions/9debe65e-7de1-48d8-9b1b-c22d8ef23c75/resourceGroups/SI-Containers/providers/Microsoft.ContainerInstance/containerGroups/default-iis?api-version=2017-12-01-preview: got HTTP response status code 400 error code "SecretVolumesNotSupportedInWindows": Secret volumes are not supported for Windows containers.

sioakim:
[...]

PR #60, opened 2018-01-18
[Azure] Filters service account secret volume mount for Windows

yolo3301:
Fixing issue: https://github.com/virtual-kubelet/virtual-kubelet/issues/40

Also fixing another small bug when kubelet port is not specified in env var. It will use the default port now.

### DIFF
--- a/providers/azure/aci.go
+++ b/providers/azure/aci.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


yolo3301:
@robbiezhang

providers/azure/aci.go
```diff
@@ -713,3 +719,37 @@ func aciContainerStateToContainerState(cs aci.ContainerState) v1.ContainerState
 		},
 	}
 }
+
+// Filters service account secret volumes for Windows.
```
robbiezhang: please add comments about why we do this.

providers/azure/aci.go
```diff
@@ -713,3 +719,37 @@ func aciContainerStateToContainerState(cs aci.ContainerState) v1.ContainerState
...
+	if strings.EqualFold(osType, "Windows") {
+		serviceAccountSecretVolumeName := ""
+
+		for index, container := range containerGroup.ContainerGroupProperties.Containers {
+			volumeMounts := make([]aci.VolumeMount, 0, len(container.VolumeMounts))
+			for _, volumeMount := range container.VolumeMounts {
+				if !strings.EqualFold(serviceAccountSecretMountPath, volumeMount.MountPath) {
+					volumeMounts = append(volumeMounts, volumeMount)
+				} else {
+					serviceAccountSecretVolumeName = volumeMount.Name
```
robbiezhang: will it result in multiple volumes?

providers/azure/aci.go
```diff
@@ -713,3 +719,37 @@ func aciContainerStateToContainerState(cs aci.ContainerState) v1.ContainerState
 		},
 	}
 }
+
+// Filters service account secret volumes for Windows.
+func filterServiceAccountSecretVolume(osType string, containerGroup *aci.ContainerGroup) {
+	if strings.EqualFold(osType, "Windows") {
+		serviceAccountSecretVolumeName := ""
```
robbiezhang: should it be an array?

yolo3301: Will there be multiple service accounts?

robbiezhang:
LGTM

robbiezhang:
@erikstmartin

PR #61, opened 2018-01-18
Add quotes for azure example.toml

yolo3301:
When not having quotes, I ran into the following errors:

* 2018/01/17 16:33:14 Near line 7 (last key parsed ''): expected a top-level item to end with a newline, comment, or EOF, but got 'G' instead
* 2018/01/17 16:41:20 toml: cannot load TOML value of type int64 into a Go string


robbiezhang:
LGTM

Issue #64, opened 2018-01-19
[Azure] EmptyDir volume cannot be used


robbiezhang:
If we use emptyDir volume in a pod, it will get the following error:

>  api call to https://management.azure.com/subscriptions/2ce4ca73-8bf4-45d5-aa55-b37121fd51a6/resourceGroups/allHandsDemo/providers/Microsoft.ContainerInstance/containerGroups/default-demo-fr-ir-aci-847b4c59f8-ck72s?api-version=2017-12-01-preview:
got HTTP response status code 400 error code "InvalidRequestContent": The request
    content was invalid and could not be deserialized: ''Could not find member ''medium''
    on object of type ''EmptyDirVolume''. Path ''properties.volumes[0].emptyDir.medium'',
    line 1, position 1052.''.'

The root cause is that the ACI provider added the "medium" and "sizeLimit" properties into the emptyDir, which are not supported by the ACI.

https://github.com/virtual-kubelet/virtual-kubelet/blob/2ca933fe27c06c5c5abc34a35ca75c936d61fffb/providers/azure/aci.go#L475-L476

PR #65, opened 2018-01-20
Fix the EmptyDir Volume Error in ACI Provider

robbiezhang:
Fix: https://github.com/virtual-kubelet/virtual-kubelet/issues/64

Add the missing vendor package.


rbitia:
thanks for the fix!! trying this out today

PR #72, opened 2018-01-25
Provider: Mock

nickschuch:
I thought it would be a great idea to write a mock provider.

That way running Kubernetes as a dependency in a pipeline like CircleCI becomes trivial.

This is a super rough bit of code, but I wanted to create this ticket and see if others were interested in this provider.

### DIFF
--- a/providers/mock/mock.go
+++ b/providers/mock/mock.go
@@ -4,6 +4,7 @@ import (
 	"log"
 	"time"
 
+	"fmt"
 	"github.com/virtual-kubelet/virtual-kubelet/providers"
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


nickschuch:
This is such a cool project!!!!

providers/mock/mock.go
```diff
@@ -0,0 +1,206 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// MockProvider implements the virtual-kubelet provider interface and stores pods in memory.
+type MockProvider struct {
+	nodeName           string
+	operatingSystem    string
+	daemonEndpointPort int32
+	pods               []*v1.Pod
```
robbiezhang: shall we change the type to map[string]*v1.Pod, so that we can support DeletePod easily? The key can be "{namespace}-{podName}".

vkubelet/vkubelet.go
```diff
@@ -91,6 +92,11 @@ func New(nodeName, operatingSystem, namespace, kubeConfig, taint, provider, prov
 		if err != nil {
 			return nil, err
 		}
+	case "mock":
+		p, err = mock.NewMockProvider(nodeName, operatingSystem, daemonEndpointPort)
```
robbiezhang: Add the `internalIP`, and return it in the NodeAddresses()

providers/mock/mock.go
nickschuch: Great idea! Ill make it happen.

vkubelet/vkubelet.go
nickschuch: :+1:

nickschuch:
Thanks for the review @robbiezhang !

robbiezhang:
LGTM

nickschuch:
Thankyou!

robbiezhang:
thank you for the contribution!

PR #74, opened 2018-01-26
Fixes #73 - illegal base64 data

### LINKED ISSUE #73: Illegal base64 data when using imagePullSecrets
Receiving the following error when using `imagePullSecrets` specified in the default service account:

```
2018/01/25 20:48:15 Error creating pod 'demo-fr-ir-aci-547fdc6c66-wjz7z': illegal base64 data at input byte 0
```

Removing the `imagePullSecret` from the service account resolves the error.  Note the same pull secret is working for other non-VK workloads.

erikstmartin:
The data being returned by the k8s client-go library is not in base64 format, it's trying to base64 decode json


erikstmartin:
I'm not as familiar with the internal stored format of these pull secrets, assuming this is a standard pull secret and it's type is not incorrectly set to SecretTypeDockercfg instead of SecretTypeDockerConfigJson, the fix should just be removing the base64 decode logic

robbiezhang:
Ops, I'm also working on the issue. One more issue in the code: the data format of the "dockercfg" is as follow:
```
{"DOCKER_REGISTRY_SERVER":{"username":"DOCKER_USER","password":"DOCKER_PASSWORD","email":"DOCKER_EMAIL","auth":"..."}}
```
#75

robbiezhang:
Close as it is fixed in #75

PR #75, opened 2018-01-26
Fix the ImagePullSecret for ACI Provider

### LINKED ISSUE #73: Illegal base64 data when using imagePullSecrets

robbiezhang:
Fix #73


gabrtv:
Confirmed that this fix resolves the `imagePullSecret` issue by modifying the VK deployment to use `microsoft/virtual-kubelet:0.2-beta-4`.

rbitia:
awesome merging in -Thanks!

Issue #71, opened 2018-01-23
 [Azure] authorization failed issue 


rbitia:
2018/01/23 01:25:25 api call to https://management.azure.com/subscriptions/resourceGroups/akscentralusrg/providers/Microsoft.ContainerInstance/containerGroups?api-version=2017-12-01-preview: got HTTP response status code 403 error code "AuthorizationFailed": The client 'with object id ' does not have authorization to perform action 'Microsoft.ContainerInstance/containerGroups/read' over scope '/subscriptions//resourceGroups/akscentralusrg/providers/Microsoft.ContainerInstance'.

Logging to track -> you will get a PEM data error if you don't feed a cert but that's fine if you don't want to use kubectl logs on ACI containers. 

The next error is a SP issue. Looking for a work around

rbitia:
We believe the SP error is an AKS install connector bug: "basically, we assume the AKS's SPN has Contributor permission on the resource group of the AKS instance, but it's not" @robbiezhang

rbitia:
To add the SP:
cat ~/.azure/aksServicePrincipal.json (file might be acsServicePrincipal.json)
Copy the number after service_principle
In your aksResourceGroup add an IAM contributor role to the SP #

Delete the pod and it will come back up on its own: kubectl delete po myaciconnector-linux-virtual-kubelet-1466967171-hbnvj’

PatrickLang:
Hmm, I can't seem to create the assignment:
`az role assignment create --assignee `cat ~/.azure/aksServicePrincipal.json | jq '.[]
 .service_principal'` --resource-group $RGNAME --role Contributor`


Operation failed with status: 'Bad Request'. Details: 400 Client Error: Bad Request for url: https://graph.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/getObjectsB
yObjectIds?api-version=1.6

spboyer:
Created the IAM contributor on RG and getting 
`tls: failed to find any PEM data in certificate input`
but not the fail to `tls`

rbitia:
@spboyer you should be good to go - that error is because a cert isn't being passed through for kubectl logs but o.w. the connector will work fine!

rbitia:
The PEM error is different from the Authorization error - I'll split out the issues to make more sense.

spboyer:
If you are still getting the error after assigning the role and deleting the connector pod, delete the pod again. I had to 2x before success.

rbitia:
[...]

PR #85, opened 2018-02-05
[Azure] Optimize VK Setup in ACS/AKS

robbiezhang:
Fix: #71 
This change load the ACS credential to be used to initialize the ACI provider if the ACS_CREDENTIAL_LOCATION is environment variable is set.


providers/azure/acsCredentials_test.go
```diff
@@ -0,0 +1,164 @@
+package azure
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+const cred = `
```
rbitia: is this for testing??

robbiezhang: yes, all the data in the cred are fake

Issue #91, opened 2018-02-09
[Azure] arm throttling limits due to incorrect region 


rbitia:
We need to add a check in our provider to see if the user is trying to spin out container instances in an unsupported region, if so DO NOT call arm and throw a helpful error instead 

---

### Environment summary

Provider (e.g. Azure, Hyper)
Azure
Version (e.g. 0.1, 0.2-beta)
latest
K8s Master Info (e.g. AKS, ACS, Bare Metal)
AKS
Install Method (e.g. Helm Chart, )
helm chart 
### Issue Details
ARM throttling limits 
### Repo Steps
Create container instances in a rg with canada, or any region aci doesn't support and you will reach an arm throttling limit

bnookala:
@rbitia I'd like to do this!

rbitia:
@robbiezhang comfortable closing this one?

yolo3301:
I think in general we should have proper backoff when we hit throttling.

robbiezhang:
I just created #105 in general to track the throttling issue for ACI provider. We can close this issue.

PR #92, opened 2018-02-10
Adding supported regions for Azure Provider

bnookala:
Based on the region availability for ACI (https://docs.microsoft.com/en-us/azure/container-instances/container-instances-quotas#region-availability), I made the following changes to make sure we only deploy ACI instances where it's supported. Fixes (avoids?) #91. 

Error message when using `westus2`:

```
$ helm install charts/virtual-kubelet -f values-linux-19.yaml

NAME:   fun-cheetah
LAST DEPLOYED: Sat Feb 10 12:04:10 2018
NAMESPACE: default
STATUS: DEPLOYED

RESOURCES:
==> v1/Secret
NAME                         TYPE    DATA  AGE
fun-cheetah-virtual-kubelet  Opaque  3     3s

==> v1beta1/Deployment
NAME                         DESIRED  CURRENT  UP-TO-DATE  AVAILABLE  AGE
fun-cheetah-virtual-kubelet  1        1        1           0          3s

==> v1/Pod(related)
NAME                                          READY  STATUS             RESTARTS  AGE
fun-cheetah-virtual-kubelet-5dd4cf6d8c-xrjdq  0/1    ContainerCreating  0         3s


NOTES:
The virtual kubelet is getting deployed on your cluster.

To verify that virtual kubelet has started, run:

  kubectl --namespace=default get pods -l "app=fun-cheetah-virtual-kubelet"

$ k get po
NAME                                                READY     STATUS    RESTARTS   AGE
fun-cheetah-virtual-kubelet-5dd4cf6d8c-xrjdq        0/1       Error     2          34s

$ k logs fun-cheetah-virtual-kubelet-5dd4cf6d8c-xrjdq
2018/02/10 20:04:45 Region westus2 is invalid. Current supported regions are: westeurope, westus, eastus, and southeastasia
```

### DIFF
--- a/providers/azure/aci.go
+++ b/providers/azure/aci.go
@@ -51,6 +51,28 @@ type AuthConfig struct {
 	RegistryToken string `json:"registrytoken,omitempty"`
 }
 
+// See https://docs.microsoft.com/en-us/azure/container-instances/container-instances-quotas for valid regions.
+var validAciRegions = []string{
+	"westeurope",
+	"westus",
+	"eastus",
+	"southeastasia",
+}
+
+// isValidACIRegion checks to make sure we're using a valid ACI region
+func isValidACIRegion(region string) bool {
+	regionLower := strings.ToLower(region)
+	regionTrimmed := strings.Replace(regionLower, " ", "", -1)
+
+	for _, validRegion := range validAciRegions {
+		if regionTrimmed == validRegion {
+			return true
+		}
+	}
+
+	return false
+}
+
 // NewACIProvider creates a new ACIProvider.
 func NewACIProvider(config string, rm *manager.ResourceManager, nodeName, operatingSystem string, internalIP string, daemonEndpointPort int32) (*ACIProvider, error) {
 	var p ACIProvider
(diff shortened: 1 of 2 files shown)


rbitia:
this is so awesome!! thanks for taking this on :)

providers/azure/aci.go
```diff
@@ -83,7 +102,15 @@ func NewACIProvider(config string, rm *manager.ResourceManager, nodeName, operat
 	}
 
 	if r := os.Getenv("ACI_REGION"); r != "" {
-		p.region = r
+		if isValidACIRegion(r) {
+			p.region = r
```
robbiezhang: please validate upon p.region in L115

providers/azure/aci.go
```diff
@@ -51,6 +51,25 @@ type AuthConfig struct {
...
 }
 
+const westEuropeRegion = "westeurope"
+const westUsRegion = "westus"
+const eastUsRegion = "eastus"
+const southEastAsiaRegion = "southeastasia"
+
+// IsValidACIRegion checks to make sure we're using a valid ACI region
+func isValidACIRegion(region string) bool {
+	switch region {
```
robbiezhang: you need to normalize the region by removing all white spaces, truncated and lowercase before the switch, since "West US" and "WestUS" are also valid Azure region names.

providers/azure/aci.go
```diff
@@ -51,6 +51,28 @@ type AuthConfig struct {
 	RegistryToken string `json:"registrytoken,omitempty"`
 }
 
+const westEuropeRegion = "westeurope"
```
robbiezhang: can you make it an array legalAciRegions? we can add region easily later. right now, we need to update 3 places to add a new region.

otherwise LGTM

bnookala: Latest change should only require making any updates to the 'validAciRegions' array.

PR #96, opened 2018-02-15
Improve ACI provider documentation

arapulido:
It took me a while to figure everything out in order to successfully registered an ACI virtual-kubelet node in my cluster.

Most of the problems I found were due to confusing and/or outdated documentation. 

This PR tries to improve the ACI provider documentation, mostly based on my experience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,22 +92,6 @@ Flags:
 Use "virtual-kubelet [command] --help" for more information about a command.
 ```
 
-## Deploy as a Pod by Helm Chart
-
-Run these commands to deploy the virtual kubelet which connects your Kubernetes cluster to Azure Container Instances.
-If you want to run the connector from the Azure command-line check out this.
-
-```bash
-RELEASE_NAME=virtual-kubelet
-CHART_URL=https://github.com/virtual-kubelet/virtual-kubelet/raw/master/charts/virtual-kubelet-0.1.0.tgz
-
-curl https://raw.githubusercontent.com/virtual-kubelet/virtual-kubelet/master/scripts/createCertAndKey.sh > createCertAndKey.sh
-. createCertAndKey.sh
-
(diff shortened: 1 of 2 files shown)


arapulido:
@ritazh Maybe you can review this one? Thanks!

rbitia:
This is awesome - thanks so much for the contribution. I'll review more in depth in the next day :)

rbitia:
lgtm! merging in. Thanks so much for this PR. Much appreciated.

ritazh:
Thanks @arapulido! 👏

rbitia:
@arapulido didn't realize we actually took out all the az aks install-connector command information. Is there a specific reason you didn't include the command in your updated documentation? 

Do you think we could add it back in? This is really the only place we have to people to see how to use the ACI Connector and using az aks install-connector is wayy easier for users, rather than developers of VK.

arapulido:
The reason was mainly because @ritazh told me that method was no longer supported, and that the helm chart was the current way of doing it. Is that not the case?

ritazh:
Hi folks, sorry for the confusion. I didnt mean az aks install-connector is no longer supported. I just meant if you have a cluster and you wanted to get vk installed, the helm chart was the easiest way regardless if you are using AKS or not. I agree if you are using AKS, that command is the quickest way. But I think we need to make it more clear in the doc that when we talk about ACI connector, it's the equivalent of the Virtual Kubelet ACI provider. The old doc was not clear and a bit misleading. Happy to make a separate PR to make it better.

rbitia:
Ah yes totally makes sense - yes let's go ahead and add the az aks install-connector to the azure provider readme but make sure to differentiate between the end-user connector experience and then the VK developer ACI provider experience. I do also like how you differentiated between an AKS cluster installation and any other K8 cluster installation for the provider.

PR #108, opened 2018-03-07
updated createCertAndKey.sh added -w0 in base64

fareszekri:
I had to add no wrapping in base64 when using bash on windows  (-w0 in base64) other wise commands fails and cert isn't properly formatted


robbiezhang:
LGTM

rbitia:
looks good!

sabbour:
-w does not exist on the MacOS version of base64 (the flag is called -b) and it defaults to 0. This will have to be done through OS detection. @robbiezhang @rbitia @fareszekri

PR #113, opened 2018-03-12
Fix possible conflict

vtrduque:
I think something undesirable entered in master 🤔


robbiezhang:
Close as a dup of #117

PR #121, opened 2018-03-22
Change azreadme

rbitia:


rbitia:
need to change naming but merging for now

PR #131, opened 2018-04-13
same point Pod

xiaoanyunfei:
rm.pods point the same pod address


robbiezhang:
@xiaoanyunfei can you elaborate why do you make this change?

xiaoanyunfei:
`type Pod struct {
	a int
}

func main() {
	Item := []Pod{{1}, {2}, {3}}
	store := make(map[int]*Pod, 3)
	for k, item := range Item {
		store[k] = &item
	}
	fmt.Println(store)
	for k, item := range Item {
		tmp := item
		store[k] = &tmp
	}
	fmt.Println(store)
}`

`
map[0:0xc4200140d0 1:0xc4200140d0 2:0xc4200140d0]
map[2:0xc420014108 0:0xc4200140f8 1:0xc420014100]
`
I think this is a serious bug, vkubelet can not work normally when creating more than 1 pod

robbiezhang:
Thanks for sharing, and it blows my mind!
Luckily, this method is only called when starting the VK, and won't impact the new pods.
Anyway, good catch. Appreciate your contribution!

erikstmartin:
Great catch @xiaoanyunfei another solution would just be to use the key to get it out of the map.

```	
for k, p := range pods.Items {
	rm.pods[p.Name] = &pods.Items[k]
}
```

@robbiezhang This is actually a common thing people run up against in Go

The variables created by the for loop are actually only every allocated once and are reused each iteration, copying in the new value. So if you take the address of it, you're getting the address of this temporary variable created by the for loop, not the address of the value itself in the original collection. 

https://play.golang.org/p/cdEit8Z7rax

ofiliz:
Right. This fix actually creates a deep-copy of the pods slice, twice (once from Items to p, then from p to tmp), which looking at the code seems to be redundant.

PR #135, opened 2018-04-14
fix missing aci_region environment variable in Helm command line

erjadi:
The manual instructions for installation are missing the env.aciRegion variable and will default to "westus", which means it fails if your service principal cannot create container instances there.


robbiezhang:
@erjadi thank you for your contribution. Can you click the "Details" link in the cla/linuxfoundation in this PR, and register with your GitHub account?

PR #138, opened 2018-04-17
Package helm chart for RBAC clusters issue #137

shanepeckham:


robbiezhang:
Please don't remove charts/virtual-kubelet-0.1.0.tgz. It's still referenced by the azure cli as a default package.

robbiezhang:
LGTP

Issue #136, opened 2018-04-16
[Azure] ACI backed pods never leave "pending" state - cannot LoadBalance them + proposed solution


erjadi:
### Environment summary

Provider : Azure
Environment: AKS
Version: Tested on v1.7.7 and v1.8.10 masters

### Issue Details

Any pods that are scheduled to run their containers on ACI never fully leave their pending state. All containers might be in "Running" state and the deployment can be successful, even "kubectl get pods" will show that the pod is ready. However, the status icon on the dashboard for the pod is stuck on pending, and "kubectl get rs" shows 0 pods in READY state.

Although the workloads will run fine, this status issue has a few unfortunate side effects, most notably the fact that these pods will never be added to service endpoints for loadbalancing. My initial plan was to load balance over ACI-backed pods, and this prevented me from doing that.

### Repo Steps

- Create a service (e.g. ClusterIP or LoadBalancer).
- Create a deployment containing a pod to be deployed on aci (set the right nodename and tolerations).
- Make sure the deployment and service labels and selectors match.
- Note that the containers and pods come online, but because of the 'pending' state it is never added to the endpoints of the service.

### Proposed solution

Currently PodConditions  are not yet implemented in the ACI provider 

https://github.com/virtual-kubelet/virtual-kubelet/blob/88bafc701b31ecf25461cd384ac7245498bc24b7/providers/azure/aci.go#L781

I've confirmed that this issue resolved when all three PodConditions are true (PodReady, PodInitialized, PodScheduled) by hard-coding those in my test fork of the kubelet. 

I would propose to either:
- Hard-code them to true (as opposed to omitting them) and keep the TODO
- Implement simple logic that only provides the three statuses once all containers are Running

robbiezhang:
I think we can hard-code the 3 conditions to true, when the provisioningState(aciState) is "Running" or "Succeeded". @erikstmartin

erjadi:
Hi guys, I took the liberty of creating a pull request (https://github.com/virtual-kubelet/virtual-kubelet/pull/147) , hope that's ok.

cpuguy83:
Is this resolved by #147?

robbiezhang:
Yes.

robbiezhang:
Yes, or just delete the existing VK pod, and let the deployment to re-create it.

PR #147, opened 2018-04-19
Azure provider now returns PodConditions based on aciState

erjadi:
Provides fix for https://github.com/virtual-kubelet/virtual-kubelet/issues/136

Tried to add aciStateToPodConditions analogous to the other aciStateToX functions. PodConditions require a transition timestamp which is why I'm passing podCreationTimestamp, however any timestamp < now would do as well.


robbiezhang:
LGTM

PR #149, opened 2018-04-21
Add resource limit check in mock provider

lioncruise:
Add resource limit check in CreatePod operation of mock provider


lioncruise:
I forgot resource limit check is done in scheduling predicate phase, this check is redundant, I'm closing this issue.

PR #175, opened 2018-04-26
[Azure] Make Resoures.Limits Optional

robbiezhang:
Make the resources.limits optional instead of filling default value.


rbitia:
yesss so what's the default values?

robbiezhang:
@rbitia The resources.limits is an optional field in ACI. The PR is to fix the issue that we set a default value when it is not specified, which violate the validation in ACI.

PR #178, opened 2018-04-27
Vendor aws-sdk-go (dep ensure)

ofiliz:
Vendor-in aws-sdk-go. This is the output of "dep ensure".

It looks like my version of dep is newer than yours, so it introduced some syntax changes in gopkg.lock


robbiezhang:
LGTM

Issue #180, opened 2018-04-30
Support additional Fargate AWS regions


ewbankkit:
---

### Environment summary

Provider **Fargate**

Version **v0.2-rc-1**

K8s Master Info **EKS**

Install Method **N/A**

### Issue Details

Fargate [now supports](https://aws.amazon.com/about-aws/whats-new/2018/04/aws-fargate-now-available-in-ohio--oregon--and-ireland-regions/) three additional AWS regions:

* us-east-2
* us-west-2
* eu-west-1

Ensure that VK's Fargate provider runs in these regions.

PR #181, opened 2018-04-30
Make Fargate provider available in 3 additional AWS regions

ewbankkit:
Fixes https://github.com/virtual-kubelet/virtual-kubelet/issues/180.

Added unit tests:

```
Running tool: /usr/local/go/bin/go test -timeout 30s github.com/virtual-kubelet/virtual-kubelet/providers/aws/fargate -run ^TestAvailableRegions$

ok  	github.com/virtual-kubelet/virtual-kubelet/providers/aws/fargate	0.037s
Success: Tests passed.
```

### DIFF
--- a/providers/aws/fargate/region.go
+++ b/providers/aws/fargate/region.go
@@ -12,6 +12,9 @@ var (
 	// FargateRegions are AWS regions where Fargate is available.
 	FargateRegions = Regions{
 		"us-east-1",
+		"us-east-2",
+		"us-west-2",
+		"eu-west-1",
 	}
 )
 


providers/aws/fargate/region_test.go
```diff
@@ -0,0 +1,30 @@
+package fargate
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestAvailableRegions verifies whether Fargate availability in an AWS region is correct.
+func TestAvailableRegions(t *testing.T) {
```
ofiliz: I'm not sure what value this test provides. This is equivalent to having two FargateRegions objects with same region strings and then comparing that they are equal.

It also means every time we add a region we have to update this second lists as well. I'd rather not have this test at all.

providers/aws/fargate/region.go
```diff
@@ -12,6 +12,9 @@ var (
 	// FargateRegions are AWS regions where Fargate is available.
 	FargateRegions = Regions{
 		"us-east-1",
+		"us-east-2",
+		"us-west-2",
+		"eu-west-1",
```
ofiliz: Did you actually do end-to-end deployment tests to these regions using a VK? I was extending the existing deployment tests before adding the new regions.

ofiliz:
Hi @ewbankkit
End-to-end VK deployment tests in the new Fargate regions succeeded. If you want to submit just the region changes in region.go, please do so now. I'll merge it and my deployment test changes.

ofiliz:
Thanks. Just FYI that E2E tests can now be run in any region by:
cd providers/aws
VK_TEST_FARGATE_REGION=us-west-2 go test

Issue #186, opened 2018-05-03
Wrong node name -lack of error


rbitia:
---

### Environment summary

Provider (e.g. ACI, AWS Fargate, Hyper)
ACI

Version (e.g. 0.1, 0.2-beta)

K8s Master Info (e.g. AKS, ACS, Bare Metal, EKS)

Install Method (e.g. Helm Chart, )
az aks install-connector 
### Issue Details
If you use the wrong node name it's really hard to figure out what went wrong.. Need to improve the exp. 
### Repo Steps
Install the connector through the az cli
Create a pod and use the wrong VK node label.

robbiezhang:
It's k8s behavior. If you use a wrong name in the podName property, the pod will disappear soon. Recommend you use nodeSelector or affinity.nodeAffinity in the sample.

robbiezhang:
looks like we don't have the common label `kubernetes.io/hostname` on the VK node.

robbiezhang:
Fix in #188

PR #188, opened 2018-05-07
Add the kubernetes.io/hostname label to the VK node

robbiezhang:
"kubernetes.io/hostname" is a [well-known label for node](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetesiohostname)
It is used in nodeSelector and nodeAffinity for pod scheduling.
In our current examples, we use nodeName to assign the pod to the VK node without the scheduler. If we use a wrong node in the pod, like #186, it's hard to troubleshoot since the pod will disappear in a few seconds.
With the hostname label, we can use the nodeSelector to ask the scheduler to assign the pod to VK node.
```
nodeSelector:
  kubernetes.io/hostname: <vk-node>
```

If the node name is wrong in the pod spec, the pod will stay in the "Pending" state, and the scheduler will create the "FailedScheduling" events in `kubectl describe po <pod-name>`:
```
Events:
  Type     Reason            Age                From               Message
  ----     ------            ----               ----               -------
  Warning  FailedScheduling  11m (x5 over 13m)  default-scheduler  No nodes are available that match all of the predicates: MatchNodeSelector (3), NodeNetworkUnavailable (1).
  Warning  FailedScheduling  11m (x7 over 13m)  default-scheduler  No nodes are available that match all of the predicates: MatchNodeSelector (3), NodeNetworkUnavailable (2).
  Warning  FailedScheduling  3m (x28 over 13m)  default-scheduler  No nodes are available that match all of the predicates: MatchNodeSelector (3).
```

### DIFF
--- a/vkubelet/vkubelet.go
+++ b/vkubelet/vkubelet.go
@@ -147,7 +147,7 @@ func (s *Server) registerNode() error {
 				"type":                  "virtual-kubelet",
 				"kubernetes.io/role":    "agent",
 				"beta.kubernetes.io/os": strings.ToLower(s.provider.OperatingSystem()),
-
+				"kubernetes.io/hostname": s.nodeName,
 				"alpha.service-controller.kubernetes.io/exclude-balancer": "true",
 			},
 		},


robbiezhang:
@ofiliz Sounds fair. Currently, we set the default node name to "virtual-kubelet" if it's not specified in the cmd, and pass in the node name when creating the provider. If we want to support the "default node name for provider", we need:
1. Make the default node name to empty in [root.go](https://github.com/virtual-kubelet/virtual-kubelet/blob/master/cmd/root.go#L76).
2. All providers should handle the empty node name case, and give a default node name.
3. Add a new function GetNodeName in the provider interface.
4. [vkubelet.go](https://github.com/virtual-kubelet/virtual-kubelet/blob/master/vkubelet/vkubelet.go#L108) calls p.GetNodeName() to set the s.nodeName
@ofiliz can you create an issue about this?

PR #195, opened 2018-05-10
CRI Provider implementation

corrieb:
This is the first pass of a CRI provider. See the README for more details. As discussed, the goal is to use this for testing and prototyping. 

It allows us to develop and test against a local Linux node with real pods and containers, which is a significant improvement over the MockProvider. 

I had to update the protobuf version in the vendor directory because of the recent changes made in version 1.1 which breaks the GRPC dependencies I added. See https://github.com/google/protobuf/issues/4582. The CRI API dependencies are configured to be added with a prune, which massively reduces the size of the vendor increases.


corrieb:
Worth noting that there are a bunch of TODOs in the code describing areas that are not fully complete. There are a *lot* of fields in the Pod spec that would need mapping to CRI for this to work for every pod. Given that the use-case is testing and development I tried to find a balance between having something that works and can be iterated on, vs all the work needed to cover every corner case.

rbitia:
Alright we will merge for now since it's a separate test provider :)

PR #201, opened 2018-05-15
Small refactor and Add NewVKubeletRunOptions struct

fisherxu:
Small refactor of the command, and add NewVKubeletRunOptions  struct contains run options of Vkubelet.


cpuguy83:
I rather like this change. The one comment I would make is that options don't really need to be in their own package and instead should convert CLI options into an options struct defined in the vkubelet package.

cpuguy83:
I'm going to close this for now since it is a bit out of date and has gone stale. If someone would like to pick this up, I've made an issue (#453) to discuss such changes.

Thanks!

PR #211, opened 2018-05-24
Add namespace for storeKey in ResourceManager

### LINKED ISSUE #196: ResourceManager needs to index by namespace for objects
Currently the ResourceManager has a single map for each object and indexes by object name. The object name is only unique within a given namespace, so you could have multiple objects of the same name within different namespaces.

As currently written, the maps can only capture one object per name, assuming it's unique. The simplest solution would be to create a composite of "namespace.name" as the key in the map.

fisherxu:
Fix #196 .
We also have this problem when debug our huawei provider, fix it.


manager/resource.go
```diff
@@ -112,35 +113,37 @@ func (rm *ResourceManager) UpdatePod(p *v1.Pod) {
...
+	if old, ok := rm.pods[podKey]; ok {
 		rm.decrementRefCounters(old)
-		delete(rm.pods, p.Name)
+		delete(rm.pods, podKey)
 	}
 }
 
 // GetPod retrieves the specified pod from the cache. It returns nil if a pod is not found.
-func (rm *ResourceManager) GetPod(name string) *v1.Pod {
+func (rm *ResourceManager) GetPod(namespaces, name string) *v1.Pod {
```
robbiezhang: namespace or ns

manager/resource.go
```diff
@@ -278,18 +288,26 @@ func (rm *ResourceManager) decrementRefCounters(p *v1.Pod) {
...
-			rm.secretRef[v.VolumeSource.Secret.SecretName]--
+			secretKey := rm.getStoreKey(p.Namespace, v.VolumeSource.Secret.SecretName)
+			rm.secretRef[secretKey]--
 		}
 	}
 }
+
+// getStoreKey return the key with namespace for store objects from different namespaces
+func (rm *ResourceManager) getStoreKey(namespace, name string) string {
+	return namespace + "-" + name
```
robbiezhang: '-' is a valid character in the name or namespace. suggest to change it to '_' to avoid conflict

fisherxu:
@robbiezhang Have addressed, PTAL. :)

PR #224, opened 2018-06-06
Fix Azure File secret names to match k8s azureFile volume secrets

anthonychu:
The Azure provider is currently using different keys in the Azure File share secrets compared to the Azure File volumes in k8s. Making them consistent.

https://docs.microsoft.com/en-us/azure/aks/azure-files-volume#create-kubernetes-secret


robbiezhang:
LGTM. @rbitia I believe it's a code bug from the very beginning, no backward compatible issue here, unless someone create the secret in the wrong way to satisfy VK in purpose.

rbitia:
Okay cool :) @anthonychu can you sign the CNCF CLA?

anthonychu:
Done thanks.

PR #236, opened 2018-06-21
Add support for displaying node and pod IP address + fix make  clean

lcastellano:
This PR contains two parts:
- A simple fix in the Makefile to make "make clean" work
- A set of changes in the VIC provider to report the IP addresses of nodes and pods and to report the correct pod status.


sflxn:
lgtm

PR #241, opened 2018-06-26
Huawei Cloud Provider implementation 

fisherxu:
Implement the official  Huawei Cloud provider for VK. We have tried to use VK in our products, and also hope to making contibutions in commnity.


fisherxu:
Yeah sure, we can add mock tests for it soon.

fisherxu:
@robbiezhang We have add the mock test for provider like azure, PTAL, Thanks! And the next step we can discuss to how to implement the network and the conformance test.

PR #255, opened 2018-07-11
mock provider: read a section of config

mqliang:
* read nodeName from environment variable
* using nodeName to locate config section

In this way: we can create a bunch of VK pods via StatefulSet, and expose Pod_Name to VK contains environment variable via DownwardAPI. All VK pods share a single configmap, and locate its own config parameter via it's name.

With this PR, we can use a single VK StatefulSet and a ConfigMap files to simulate a bunch of heterogenous nodes.  

example yaml files and config files are like:

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: virtual-kubelet-set
  # TODO: determine the namespace
  namespace: default
spec:
  serviceName: "virtual-kubelet-set"
  replicas: 2
  selector:
      matchLabels:
        app: virtual-kubelet-set
  template:
    metadata:
      labels:
        app: virtual-kubelet-set
    spec:
      containers:
        - name: virtual-kubelet
          image: virtual-kubelet:v0.3
          command: [ "/usr/bin/virtual-kubelet"]
          args:
            - --provider
            - mock
            # Add a "mock" taint to the VK node.
            # Only our regular jobs pods can tolerate this taint
            # This can prevent VK pod been scheduled to a VK node
            - --taint
            - mock
            - --provider-config
            - /etc/config/configs.json
          env:
            - name: NODE_NAME
              valueFrom:
                fieldRef:
                  fieldPath: metadata.name
          volumeMounts:
            - name: config-volume
              mountPath: /etc/config
      volumes:
        - name: config-volume
          configMap:
            name: vk-config

```

configs.json:
```
{
  "virtual-kubelet-set-0":
  {
    "cpu":"5",
    "memory":"5Gi",
    "pods":"5",
  },
  "virtual-kubelet-set-1":
  {
    "cpu":"10",
    "memory":"10Gi",
    "pods":"10"
  }
}
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -67,6 +67,11 @@ func Execute() {
 func init() {
 	cobra.OnInitialize(initConfig)
 
+	// read default node name from environment variable.
+	// it can be overwritten by cli flags if specified.
+	if os.Getenv("DEFAULT_NODE_NAME") != "" {
+		nodeName = os.Getenv("DEFAULT_NODE_NAME")
+	}
 	// Here you will define your flags and configuration settings.
 	// Cobra supports persistent flags, which, if defined here,
 	// will be global for your application.
--- a/providers/mock/mock.go
+++ b/providers/mock/mock.go
@@ -42,7 +42,7 @@ type MockConfig struct {
 
 // NewMockProvider creates a new MockProvider
 func NewMockProvider(providerConfig, nodeName, operatingSystem string, internalIP string, daemonEndpointPort int32) (*MockProvider, error) {
-	config, err := loadConfig(providerConfig)
+	config, err := loadConfig(providerConfig, nodeName)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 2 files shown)


mqliang:
@robbiezhang
cc @glikson

providers/mock/mock.go
```diff
@@ -59,17 +59,18 @@ func NewMockProvider(providerConfig, nodeName, operatingSystem string, internalI
...
-		if err != nil {
-			return config, err
-		}
+
+func loadConfig(providerConfig, nodeName string) (config MockConfig, err error) {
+	data, err := ioutil.ReadFile(providerConfig)
+	if err != nil {
+		return config, err
+	}
+	configMap := map[string]MockConfig{}
```
robbiezhang: it's not backward compatible

providers/mock/mock.go
```diff
@@ -59,17 +59,18 @@ func NewMockProvider(providerConfig, nodeName, operatingSystem string, internalI
...
+	err = json.Unmarshal(data, &configMap)
+	if err != nil {
+		return config, err
 	}
+	config = configMap[nodeName]
```
robbiezhang: check the config exists

cmd/root.go
```diff
@@ -82,6 +82,9 @@ func init() {
 	// Cobra also supports local flags, which will only run
 	// when this action is called directly.
 	// RootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+	if os.Getenv("NODE_NAME") != "" {
```
robbiezhang: to be clear, if --nodename is specified, the value will override the NODE_NAME environment variables, right?

providers/mock/mock.go
mqliang: @robbiezhang Hi, the config feature for mock provider is added by me just few days ago. I don't think others is already using it. So, should we still want to make it backward compatible?

cmd/root.go
robbiezhang: In this case, shall we make it the "defaultNodeName"?
move it before the PersistentFlags sections, and change the envVar name to DEFAULT_NODE_NAME

```
defaultNodeName := os.Getenv("DEFAULT_NODE_NAME")
if  defaultNodeName == "" {
    defaultNodeName = "virtual-kubelet"
}
...
RootCmd.PersistentFlags().StringVar(&nodeName, "nodename", defaultNodeName, "kubernetes node name")

```

providers/mock/mock.go
robbiezhang: Sounds fair.

mqliang:
@robbiezhang Updated. PTAL.

PR #258, opened 2018-07-16
Consolidate helm charts

### LINKED ISSUE #215: Reason for Having Two Helm Charts
Hi there,

I'm considering submitting a pull request to generalize the existing virtual-kubelet Helm chart(s) and I was wondering why there are currently two separate chart directories (`virtual-kubelet` and `virtual-kubelet-for-aks`). From what I can tell, these are very similar, especially considering the `virtual-kubelet` chart is really only useful on AKS at the moment anyway.

As the number "virtual infrastructure" services grow, I think having a pluggable chart that's as generic as possible makes sense, but if there are differences between the environments that warrant their own charts, then the existing layout might make sense. I just wanted to ensure that that is in fact the case before proceeding with any future work.

Thanks!

jlegrone:
Closes #215.

Changes
- Added an appVersion and icon to helm Chart.yaml
- Refactored to use `required` in chart manifests rather than
  outputting an error message in notes
- Namespaced `name` and `fullname` template partials to `vk`
- Enabled rbac apiVersion configuration
- Removed role-binding and service-account suffixes from resource
  names
- Fixed bug where virtual-kubelet service account would not be
  bound to cluster role if chart was installed outside the default
  namespace
- Removed hardcoded `azure` provider
- Added a helm test hook (usage: `helm test <release-name>`)
- Rollout new pods when secret updated

Breaking Changes
- Virtual-Kubelet specific values previously nested under `env` are
  now declared at the top level of the values file.
- Azure provider configuration values previously nested under `env`
  are now at `providers.azure`

Please note that this needs some QA! (I've only tested on AKS with k8s 1.10.5)

AKS installation instructions:

```bash
#!/bin/bash

. ./scripts/createCertAndKey.sh &&

helm upgrade --install $RELEASE_NAME ./charts/virtual-kubelet \
  --set provider=azure \
  --set rbac.install=true \
  --set providers.azure.tenantId=$AZURE_TENANT_ID \
  --set providers.azure.subscriptionId=$AZURE_SUBSCRIPTION_ID \
  --set providers.azure.clientId=$AZURE_CLIENT_ID \
  --set providers.azure.clientKey=$AZURE_CLIENT_SECRET \
  --set providers.azure.aciResourceGroup=$AZURE_RG \
  --set providers.azure.aciRegion=$ACI_REGION \
  --set apiserverCert=$cert \
  --set apiserverKey=$key
```

You can also test your installation of virtual-kubelet by running

```cli
helm test $RELEASE_NAME
```


jlegrone:
Thoughts on also defaulting `rbac.install` to `true` as part of this breaking change? I have a feeling most folks are deploying to clusters with rbac on these days.

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -30,15 +30,61 @@ spec:
...
+        - name: AZURE_TENANT_ID
+          value: {{ required "tenantId is required" .tenantId }}
+        - name: AZURE_SUBSCRIPTION_ID
+          value: {{ required "subscriptionId is required" .subscriptionId }}
+        - name: AZURE_CLIENT_ID
+          value: {{ required "clientId is required" .clientId }}
+        - name: AZURE_CLIENT_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "vk.fullname" $ }}
```
robbiezhang: naive question, what does $ mean?

jlegrone: `$` is always a reference to the root scope (we're currently in a `with` block to make referencing keys in `.Values.providers.azure` a bit less repetitive).

Reference at the bottom of this page: https://github.com/helm/helm/blob/01da56f956329bbfcef14ac164f9ab49081ad108/docs/chart_template_guide/variables.md

stevenpall:
Along the same lines as #215, I was wondering if the scope of this PR is limited to Azure providers or if this chart could eventually also support others? I'm particularly interested in Fargate. I see that there are provider-specific variables in this PR's updated `values.yaml`, so perhaps this is already supported here? Thanks!

charts/virtual-kubelet/templates/deployment.yaml
robbiezhang: Make sense. Thanks for the explanation.

robbiezhang:
This helm chart is originally created for the Azure provider. I'm fine with sharing it with other providers. However, it will introduce a lot of if-else conditions and increase the maintenance cost. @rbitia

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -30,15 +28,65 @@ spec:
...
+        - name: acs-credential
+          mountPath: "/etc/acs/azure.json"
+{{- end }}
+{{- end }}
         command: ["virtual-kubelet"]
-        args: ["--provider", "azure", "--namespace", {{ default "" .Values.env.monitoredNamespace | quote }}, "--nodename", {{ default "virtual-kubelet" .Values.env.nodeName | quote }} , "--os", {{ default "Linux" .Values.env.nodeOsType | quote }}, "--taint", {{ default "azure.com/aci" .Values.env.nodeTaint | quote }}]
+        args: [
+          "--provider", "{{ required "provider is required" .Values.provider }}",
+          "--namespace", "{{ .Values.monitoredNamespace }}",
+          "--nodename", "{{ required "nodeName is required" .Values.nodeName }}",
```
robbiezhang: no need to be required, since we've given the default value "virtual-kubelet" in the values.yaml

robbiezhang:
@jlegrone the helm chart is not backward compatible. I would suggest the virtual-kubelet-for-aks folder until the `az aks install-connector` switch to the new chart.

stevenpall:
I have some changes ready to go based on this PR's chart so that it also supports Fargate as a provider. I can create a separate PR after this gets merged in.

jeremyrickard:
Hey @jlegrone are you able to rebase this with the changes that have occurred since you opened the PR? If not, I can take a look.

jlegrone:
@jeremyrickard I just got back from vacation. Yes, I'm working on getting this mergeable. :)

charts/virtual-kubelet/templates/deployment.yaml
jlegrone: Marking values as required is still desirable in the case that the user overrides with a falsy value or someone removes the default value in the future. This catches that case at install time.

jlegrone:
> the helm chart is not backward compatible. I would suggest to keep the virtual-kubelet-for-aks folder until the az aks install-connector switch to the new chart.

@robbiezhang this PR does not touch `virtual-kubelet-latest.tgz` or `virtual-kubelet-for-aks-latest.tgz`, so the current installation method should continue working, correct?

There are some breaking changes, but they can be handled by updating the aks-connector script to use `virtual-kubelet-0.2.0.tgz` temporarily.

robbiezhang:
@jlegrone, the `az aks install-connector` update takes weeks to release. I'm recommending that we keep the virtual-kubelet-for-aks folder in your PR until we update the `az aks`, just in case of any urgent bug fix. Keeping the folder can help us to check the code easily.

robbiezhang:
@jeremyrickard, can you take another look? Especially, the api server cert generation part.

PR #265, opened 2018-07-24
Fix bug in exec command retrieval

d-e-s-o:
The exec command as extracted from the query comprises only the first
part of the command and does not include potentially supplied
parameters. E.g.,
  $ kubectl exec pod -- ls -t /usr
  > command: ls

This change fixes the problem by moving away from the Query.Get API.
  $ kubectl exec pod -- ls -t /usr
  > command: [ls -t /usr]


robbiezhang:
thanks for the fix. LGTM

PR #269, opened 2018-07-24
Adding network policy reference

jeremyrickard:
Updating aci sdk to add network profile bits


robbiezhang:
one minor issue in the comment; otherwise, looks good to me.

PR #274, opened 2018-07-25
Plumb through log analytics values

rchakra3:
This PR adds the ability to specify a log analytics workspace key and ID

### DIFF
--- a/charts/virtual-kubelet/templates/deployment.yaml
+++ b/charts/virtual-kubelet/templates/deployment.yaml
@@ -26,10 +26,10 @@ spec:
           value: /etc/virtual-kubelet/cert.pem
         - name: APISERVER_KEY_LOCATION
           value: /etc/virtual-kubelet/key.pem
-        - name: LOG_ANALYTICS_ID
-          value: {{ .Values.env.logAnalyticsID }}
-        - name: LOG_ANALYTICS_KEY
-          value: {{ .Values.env.logAnalyticsKey }}
+        {{ if .Values.loganalytics.enabled }}
+        - name: LOG_ANALYTICS_AUTH_LOCATION
+          value: /etc/virtual-kubelet/loganalytics.json
+        {{ end }}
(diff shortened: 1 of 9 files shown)


robbiezhang:
LGTM

providers/azure/client/aci/analytics.go
```diff
@@ -0,0 +1,17 @@
+package aci
+
+import "errors"
+
+func NewContainerGroupDiagnostics(logAnalyticsID, logAnalyticsKey string) (*ContainerGroupDiagnostics, error) {
```
robbiezhang: nit: please add a comment to the public functions

providers/azure/aci.go
```diff
@@ -155,6 +156,25 @@ func NewACIProvider(config string, rm *manager.ResourceManager, nodeName, operat
 		return nil, err
 	}
 
+	// If the log analytics file has been specified, load workspace credentials from the file
+	if logAnalyticsAuthFile := os.Getenv("LOG_ANALYTICS_AUTH_LOCATION"); logAnalyticsAuthFile != "" {
```
robbiezhang: add LOG_ANALYTICS_AUTH_LOCATION into the helm chart?

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -26,6 +26,10 @@ spec:
           value: /etc/virtual-kubelet/cert.pem
         - name: APISERVER_KEY_LOCATION
           value: /etc/virtual-kubelet/key.pem
+        - name: LOG_ANALYTICS_ID
+          value: {{ .Values.env.logAnalyticsID }}
+        - name: LOG_ANALYTICS_KEY
+          value: {{ .Values.env.logAnalyticsKey }}
```
robbiezhang: let's use the LOG_ANALYTICS_AUTH_LOCATION approach, write the ID and key into the vk secret. It will be mounted as a secret volume.

PR #275, opened 2018-07-25
Introduce certificate generation

jeremyrickard:
This PR adds the ability for vk to generate self signed certificates if none are provided (or if invalid certs are provided). We currently instruct users to run a script to generate self signed certs when they
install. This removes that step and makes initial experience a little nicer.

The certificate that is generated is pretty similar to what you'd get from the existing script, however when a certificate is generated, it will be valid for 365 days (1 year) vs 3650 days (10 years).


vkubelet/certs/certificates.go
```diff
@@ -0,0 +1,93 @@
...
+		log.Printf("Failed to create certificate: %s", err)
+		return "", "", err
+	}
+	certOut, err := os.Create("cert.pem")
+	if err != nil {
+		log.Printf("failed to open cert.pem for writing: %s", err)
+		return "", "", err
+	}
+	pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
+	certOut.Close()
```
robbiezhang: defer, and move before previous line.

vkubelet/certs/certificates.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		log.Print("failed to open key.pem for writing:", err)
+		return "", "", err
+	}
+	p := &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(priv),
+	}
+	pem.Encode(keyOut, p)
+	keyOut.Close()
```
robbiezhang: defer and call it after the L85

jeremyrickard:
Will clean up the comments.

vkubelet/apiserver.go
```diff
@@ -26,6 +27,16 @@ func ApiserverStart(provider Provider) {
 	p = provider
 	certFilePath := os.Getenv("APISERVER_CERT_LOCATION")
 	keyFilePath := os.Getenv("APISERVER_KEY_LOCATION")
+
+	var err error
+	if valid := certs.IsCertKeyPairValid(certFilePath, keyFilePath); !valid {
+		log.Printf("provided cert/key pair is invalid or empty, generating.")
```
cpuguy83: Should this just error out in this case and the user would have to restart VK without the certs?

This also would need a `\n` or a `log.Println`.

Some minor nits:
If we do keep this it may be good to separate the log entries.
The error message my not be exactly accurate either in case these env vars were not set at all.

vkubelet/apiserver.go
```diff
@@ -26,6 +27,16 @@ func ApiserverStart(provider Provider) {
...
+		certFilePath, keyFilePath, err = certs.GenerateCertKeyPair()
+		if err != nil {
+			log.Printf("error generating certificate pair: %s\n", err)
```
cpuguy83: "... pair for HTTP listener"

vkubelet/apiserver.go
cpuguy83: Also should warn that these certs are not suitable for production and the user should provide their own.

jeremyrickard: 👍 both of these are great suggestions.

jeremyrickard: > Should this just error out in this case and the user would have to restart VK without the certs?

@robbiezhang said we should generate them if they were invalid. I'd tend to say it should _fail_ if invalid and not generate over top of them. I'll make that change if we agree that's the right behavior.

vkubelet/apiserver.go
```diff
@@ -26,6 +27,21 @@ func ApiserverStart(provider Provider) {
 	p = provider
 	certFilePath := os.Getenv("APISERVER_CERT_LOCATION")
 	keyFilePath := os.Getenv("APISERVER_KEY_LOCATION")
+
+	var err error
+	
+	log.Println(certFilePath)
+	log.Println(keyFilePath)
+	
+	if certFilePath == "" || keyFilePath == "" {
```
cpuguy83: Should this be "&&"?

jeremyrickard: Ahhhh yeah. Forgot to swap this when we switched to not overwriting bad combos.

cpuguy83:
Can you clean up commits?

PR #276, opened 2018-07-26
use secure value in ACI for secrets

rchakra3:


robbiezhang:
can you add a mock test?

rchakra3:
@robbiezhang sorry for the delay - I've added tests. It looks like tests for another provider are failing.. the ACI tests look good

PR #281, opened 2018-07-30
Update VK AKS chart to have RBAC resources

jeremyrickard:
This PR updates the VK AKS chart to have RBAC resources and conditionally create them. rbac.install defaults to true. 

I'm going to make a corresponding PR to azure cli to add a call to get the rbac status of the cluster and use that flag to drive the chart install.


jeremyrickard:
I shouldn't have updated the virtual-kubelet-latest.tgz, let me fix that.

charts/virtual-kubelet-for-aks/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,14 @@
+{{ if .Values.rbac.install }}
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "fullname" . }}-role-binding
```
robbiezhang: shall we remove all suffixes like, "-role-binding", "-service-account"?

charts/virtual-kubelet-for-aks/templates/deployment.yaml
```diff
@@ -56,5 +56,6 @@ spec:
         hostPath:
           path: /etc/kubernetes/azure.json
           type: File
+      serviceAccountName: {{ if .Values.rbac.install }} "{{ template "fullname" . }}-service-account" {{ end }}
```
robbiezhang: How about:
```
{{ if .Values.rbac.install }}
      serviceAccountName: {{ template "fullname" . }}
{{ end }}
```

charts/virtual-kubelet-for-aks/values.yaml
```diff
@@ -15,3 +15,12 @@ env:
   apiserverCert: 
   apiserverKey: 
   monitoredNamespace:
+
+# Install Default RBAC roles and bindings
+rbac:
+  install: true
+  serviceAccountName: virtual-kubelet
```
robbiezhang: serviceAccountName is not used in the deployment. I'm fine with using the fullname, but please remove this value to avoid confusion.

charts/virtual-kubelet-for-aks/values.yaml
```diff
@@ -15,3 +15,12 @@ env:
...
+  # RBAC api version (currently v1beta1)
+  apiVersion: v1beta1
```
robbiezhang: not used either, please use it or remove.

charts/virtual-kubelet-for-aks/templates/deployment.yaml
jeremyrickard: I just copied these over as is from the main virtual-kubelet chart. Do you want the following cleaned up for that as well?

jeremyrickard:
@robbiezhang i copied the rbac related files over as is from the main chart directory. Would you like those cleaned up as well as part of this PR? All the comments you have above will apply there as well.

charts/virtual-kubelet-for-aks/templates/deployment.yaml
robbiezhang: Yup, appreciate it if you can make them consistent.

robbiezhang: Another thought: since we have the open PR #258, let's not touch the `virtual-kubelet` chart.

jeremyrickard: 👍

robbiezhang:
@jeremyrickard my comments are mostly to make it consistent to the PR #258. So, let's keep virtual-kubelet chart unchanged, and wait for #258 to be merged

jeremyrickard:
I've updated everything, but let me validate w/ AKS connector before we merge.

PR #286, opened 2018-08-01
Add certificate generation to Helm charts

jeremyrickard:
This PR adds the ability for vk to generate self signed certificates if none are provided (or if invalid certs are provided). We currently instruct users to run a script to generate self signed certs when they
install. This removes that step and makes initial experience a little nicer.

This PR is an alternative approach to #275 and handles everything within Helm, rather than adding that logic to the VK api server. IMO, this is a better scenario. My preference would be to accept this and close #275


charts/virtual-kubelet/templates/secrets.yaml
```diff
@@ -1,12 +1,22 @@
+{{- $ca := genCA "virtual-kubelet-ca" 3650 }}
+{{- $cn := printf "%s-virtual-kubelet-apiserver" .Release.Name }}
+{{- $altName1 := printf "%s-virtual-kubelet-apiserver.%s" .Release.Name .Release.Namespace }}
+{{- $altName2 := printf "%s-cvirtual-kubelet-apiserver.%s.svc" .Release.Name .Release.Namespace }}
```
cpuguy83: How com these are all generated here instead of only when they are needed below?

jeremyrickard: updated :)

charts/virtual-kubelet-for-aks/templates/secrets.yaml
```diff
@@ -4,6 +4,16 @@ metadata:
...
 type: Opaque
 data:
-  cert.pem: {{ (default "TUlTU0lORw==" .Values.env.apiserverCert) | quote }}
-  key.pem: {{ (default "TUlTU0lORw==" .Values.env.apiserverKey) | quote }}
-  clientSecret: {{ default "" .Values.env.azureClientKey | b64enc | quote }}
+  {{- if (not .Values.env.apiserverCert) and (not .Values.env.apiserverKey) }}
+  {{- $ca := genCA "virtual-kubelet-ca" 3650 }}
+  {{- $cn := printf "%s-virtual-kubelet-apiserver" .Release.Name }}
+  {{- $altName1 := printf "%s-virtual-kubelet-apiserver.%s" .Release.Name .Release.Namespace }}
+  {{- $altName2 := printf "%s-cvirtual-kubelet-apiserver.%s.svc" .Release.Name .Release.Namespace }}
```
robbiezhang: remove the extra "c" in -cvirtual...

charts/virtual-kubelet-for-aks/templates/secrets.yaml
```diff
@@ -4,6 +4,16 @@ metadata:
...
+  {{- $cert := genSignedCert $cn nil (list $altName1 $altName2) 3650 $ca }}
+  cert.pem: {{ b64enc $cert.Cert }}
+  key.pem: {{ b64enc $cert.Key }}
+  {{ else }}
+  cert.pem: {{ quote .Values.env.apiserverCert }}
```
robbiezhang: shall we error out if any of them is empty here? since in L7, you only generate the cert when both of them are empty.

jeremyrickard: The direction in the other PR was to allow things to fall through and let the http listener report the error. I'm good either way, I did this to follow the behavior from that discussion, but I don't have a strong opinion either way. Your call.

robbiezhang: Sounds fair. we can leave it as it is.

Issue #291, opened 2018-08-03
go deps not correct after servicefabricmesh provider was added


jeremyrickard:
Looks like we need to bump the azure-go-sdk dependency for the service fabric mesh provider:

```
+ make build
Building...
providers/sfmesh/sfmesh.go:14:2: cannot find package "github.com/Azure/azure-sdk-for-go/profiles/preview/preview/servicefabricmesh/mgmt/servicefabricmesh" in any of:
	/go/src/github.com/virtual-kubelet/virtual-kubelet/vendor/github.com/Azure/azure-sdk-for-go/profiles/preview/preview/servicefabricmesh/mgmt/servicefabricmesh (vendor tree)
	/usr/local/go/src/github.com/Azure/azure-sdk-for-go/profiles/preview/preview/servicefabricmesh/mgmt/servicefabricmesh (from $GOROOT)
	/go/src/github.com/Azure/azure-sdk-for-go/profiles/preview/preview/servicefabricmesh/mgmt/servicefabricmesh (from $GOPATH)
providers/sfmesh/sfmesh.go:15:2: cannot find package "github.com/Azure/go-autorest/autorest/azure/auth" in any of:
	/go/src/github.com/virtual-kubelet/virtual-kubelet/vendor/github.com/Azure/go-autorest/autorest/azure/auth (vendor tree)
	/usr/local/go/src/github.com/Azure/go-autorest/autorest/azure/auth (from $GOROOT)
	/go/src/github.com/Azure/go-autorest/autorest/azure/auth (from $GOPATH)
make: *** [Makefile:25: build] Error 1
```

cpuguy83:
How did CI pass?

robbiezhang:
CI runs `go get`
https://github.com/virtual-kubelet/virtual-kubelet/blob/bcea8b955bfcbf8aba4863146313918a05c11814/.circleci/config.yml#L20

Shall we remove it?

jeremyrickard:
Yes, CI should verify that any new PR doesn't introduce a new dependency. The build should fail if that's the case.

jeremyrickard:
We also _REALLY_ need to update our dep config to pin versions. To fix this, I now have updates to all kinds of stuff, including the vic provider.

PR #293, opened 2018-08-03
Validate deps in CI

cpuguy83:
This makes sure a PR has properly updated deps.

Note that some reps are not pinned right now so any changes will be picked up each time `dep ensure` is run. This needs to be fixed separately.

Relates to #291 
Ping @jeremyrickard @robbiezhang


cpuguy83:
Ok, this should be ready.

cpuguy83:
Sorry, this was a terrible idea... fixing the script

cpuguy83:
Ok, this is much better. I think it should work for all cases of mismatched deps.

scripts/validate/dep.sh
```diff
@@ -0,0 +1,16 @@
+
```
cpuguy83: Looks like my `#!` got deleted somehow...

PR #294, opened 2018-08-03
Update dependencies to add service fabric mesh via new azure-sdk-go

jeremyrickard:
When we merged the PR for `added Service Fabric Mesh provider`, a dependency change was introduced that required an update of our azure-sdk-go dependency.  Unfortunately, we don't properly constrain our versions so this PR has a number of other updates.

Related to #291

PR #297, opened 2018-08-03
Fix how ACI provider handles empty environment variables

### LINKED ISSUE #296: ACI provider doesn't handle empty environment variables well
### Environment summary

Provider (e.g. ACI, AWS Fargate, Hyper)
*ACI*

Version (e.g. 0.1, 0.2-beta)
*Master*

*AKS* 

*Helm Chart*

### Issue Details
When you create a pod with an environment variable without a value, ACI returns an error:

```
2018/08/03 18:41:55 Error creating pod 'nginx': api call to https://management.azure.com/subscriptions/76fb3144-bc19-4baf-ab79-432d526559b7/resourceGroups/MC_aks-group_vk-cluster_eastus/providers/Microsoft.ContainerInstance/containerGroups/default-nginx?api-version=2018-06-01: got HTTP response status code 400 error code "InvalidEnvironmentVariable": The environment variable 'SOME_EMPTY_ENV' in container 'nginx' of container group 'default-nginx' is invalid. One and only one property of 'value' and 'secureValue' can be specified in an environment variable.
```

The error indicates that _both_ are being sent. In reality, there is _no_ value. `kubectl` / kuberentes allows this, so we need to be able to handle it. 

Additionally, there was an issue introduced during the `SecureValue` PR that resulted in a panic when value was empty. 

### Repo Steps

This YAML:

```
apiVersion: v1
kind: Pod
metadata:
  name: nginx-bad-env2
spec:
  containers:
  - image: nginx
    imagePullPolicy: Always
    name: nginx
    env:
    - name: SOME_EMPTY_ENV
    ports:
    - containerPort: 80
      name: http
      protocol: TCP
    - containerPort: 443
      name: https
  dnsPolicy: ClusterFirst
  nodeSelector:
    kubernetes.io/role: agent
    beta.kubernetes.io/os: linux
    type: virtual-kubelet
  tolerations:
  - key: azure.com/aci
    effect: NoSchedule
```
will result in the error.

jeremyrickard:
There are a couple of issues with how the ACI provider handles empty environment variables...i.e. those defined like:

```
apiVersion: v1
kind: Pod
metadata:
  name: nginx-bad-env2
spec:
  containers:
  - image: nginx
    imagePullPolicy: Always
    name: nginx
    env:
    - name: SOME_EMPTY_ENV
    ports:
    - containerPort: 80
      name: http
      protocol: TCP
    - containerPort: 443
      name: https
  dnsPolicy: ClusterFirst
  nodeSelector:
    kubernetes.io/role: agent
    beta.kubernetes.io/os: linux
    type: virtual-kubelet
  tolerations:
  - key: azure.com/aci
    effect: NoSchedule
```

First, we ended up having a panic if the value was empty now. This was due to the change for SecureValue. Second, ACI actually errors if you don't provide a value. So this PR excludes any env without a value (which is a totally valid thing to send via k8s apis)

Fixes #296


jeremyrickard:
This will fail CI now until #294 is merged. deps are out of synch,

cpuguy83:
Required PR's are merged.

jeremyrickard:
> It seems like ACI should accept an empty env as an empty env is totally valid.

I agree. @robbiezhang is that something that might get updated in ACI later?

robbiezhang:
@jeremyrickard looks like a nice-to-have feature. I'll create a work item in ACI to track it.

PR #298, opened 2018-08-04
Update dep validation to only check the vendor dir

cpuguy83:
This helps make sure that different versions of dep don't give different
results.


cpuguy83:
All green after this.

Issue #259, opened 2018-07-17
Support configuring node taint value and effect


jlegrone:
Providing a node taint with a key would enable supporting multiple providers with the same key:

```yaml
tolerations:
- key: "virtual-kubelet.github.io/provider"
  value: "azure"
  operator: "Equal"
  effect: "NoSchedule"
```

```yaml
tolerations:
- key: "virtual-kubelet.github.io/provider"
  value: "azurebatch"
  operator: "Equal"
  effect: "NoSchedule"
```

While also being able to schedule workloads on any virtual-kubelet node:

```yaml
tolerations:
- key: "virtual-kubelet.github.io/provider"
  operator: "Exists"
  effect: "NoSchedule"
```

Supporting effect configuration meanwhile would be advantageous for use cases where virtual-kubelet nodes are meant to handle excess workloads, using `PreferNoSchedule`.

Reference: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#concepts

jlegrone:
A few related questions:

1. Should providers be able to override or add their own taint(s)?
1. Would it be worthwhile to provide a flag to create a default taint (like the provider example above)?
1. What flags should be used to specify taint value and effect?

rbitia:
So we do want to move to this model. Issue is backwards compatibility. We also don't want to host all providers in one chart but having a generic taint is helpful for users. 
Should providers be able to override or add their own taint(s)?
Probably not.
Would it be worthwhile to provide a flag to create a default taint (like the provider example above)?
yes
What flags should be used to specify taint value and effect? Use env variables instead?
hmm I believe so
What prefix should be used for a VK taint (must be a valid subdomain), if any?
do you mean key for prefix?

jlegrone:
> What prefix should be used for a VK taint (must be a valid subdomain), if any?
do you mean key for prefix?

@rbitia in the above examples I used `virtual-kubelet.github.io` as the prefix.  I can't find up to date documentation at the moment but this seems to be a good explanation (I believe the same rules probably apply for taints):
https://github.com/brendandburns/kubernetes/blob/25b37b6978709aa83d808d368af846e7536f7383/docs/labels.md#syntax-and-character-set

To avoid breaking changes we might want to choose a prefix that is associated with a domain that won't change if this repo becomes owned by a different github organization in the future.

jeremyrickard:
@rbitia @jlegrone I have a follow up question

[...]

PR #305, opened 2018-08-08
Provide default "provider" taint and support configuration of taint effect

jlegrone:
Closes https://github.com/virtual-kubelet/virtual-kubelet/issues/259

I'm opening this PR as a straw-man proposal to get feedback on the design and requirements.  There are several things I think should be validated:

1. Is `virtual-kubelet.github.io` the right prefix for us to use for things like taints, labels, and annotations long term? Updating this prefix in the future would be a breaking change.
1. Is this the right level of configuration?
    1. The provider taint can be disabled
    1. The user can specify a custom taint to be used in addition to the default provider taint
    1. The user cannot specify a value for their custom taint
    1. The provider taint and a user specified taint inherit the same taint effect, which is configurable via an environment variable
1. Is the breaking change of adding a provider taint by default acceptable?
1. Is the breaking change of removing the `azure.com/aci` taint in the helm chart acceptable?

These changes have not been tested (and probably break something) so I'll only remove the WIP tag from this PR after I've done more validation.


jeremyrickard:
@jlegrone thanks for opening this! I was doing the same thing in parallel and I arrived at a pretty similar end state. I have some opinions, I'll comment with them inline :)

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -77,7 +79,8 @@ spec:
           "--namespace", "{{ .Values.monitoredNamespace }}",
           "--nodename", "{{ required "nodeName is required" .Values.nodeName }}",
           "--os", "{{ .Values.nodeOsType }}",
-          "--taint", "{{ .Values.nodeTaint }}"
+          "--taint", "{{ .Values.nodeTaint }}",
+          "--disable-default-taint", "{{ .Values.disableDefaultTaint }}"
```
jeremyrickard: I'd be in favor a simpler approach where we only go with something like `virtual-kubelet.github.io:<provider>=<Effect>` as the built in Virtual Kubelet functionality and not include the `disable-default-taint` functionality.  I thought one of the goals here was to kind of harmonize on behavior that would allow multiple providers, but also the "don't care" behavior. I think adding the `disable-default-taint` behavior kind of distracts from that. Same for adding custom taints beyond the "default". It's fairly easy to add/remove whatever taints you want via kubectl, so if you want to deviate from the built in functionality, that's possible. (counter arguement: you could do that with the taints in general, but I feel like having the common format one makes VK a more usable thing across providers when i'm the end user).

charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -77,7 +79,8 @@ spec:
           "--namespace", "{{ .Values.monitoredNamespace }}",
           "--nodename", "{{ required "nodeName is required" .Values.nodeName }}",
           "--os", "{{ .Values.nodeOsType }}",
-          "--taint", "{{ .Values.nodeTaint }}"
+          "--taint", "{{ .Values.nodeTaint }}",
```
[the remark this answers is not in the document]
jlegrone: I'd also be happy with that approach.  Switching to an environment variable would be a breaking change for `virtual-kubelet`, but not for the helm chart.

Flags can be a bit more self-documenting, but I have a feeling that most users will be interacting with vk through helm so practically speaking that benefit is limited.  Environment variables are also compatible with things like pod presets.

charts/virtual-kubelet/templates/deployment.yaml
jlegrone: I was trying to think of a use case for needing two node taints, and have been coming up empty.

How about three environment variables:
- `VKUBELET_TAINT_KEY` (default `virtual-kubelet.github.io/provider`)
- `VKUBELET_TAINT_VALUE` (default `{{ .Values.provider }}`)
- `VKUBELET_TAINT_EFFECT` (default `NoSchedule`)

I'd be in favor of making this a breaking change in the helm values file so that those who have been overriding `azure.com/aci` get the new default taint key instead.

jlegrone: One other thought is that we should probably provide a way to disable taints altogether on virtual-kubelet nodes.  That could be a future change too, but something to keep in mind for a time when vk is stable enough for users to not care about explicitly "opting in".  Perhaps that makes sense as a flag?

jeremyrickard: Yeah, this is a great point ☝️ would, 100% be in favor of that. 

I'm good with either flags or environment variables for either of these, I just want to be consistent within the taint definition part..so either all envs (which works well with pod presets like you mentioned) or all flags.

jeremyrickard: I like those three environment variables BTW, I vote this way. @rbitia @robbiezhang, any input?

charts/virtual-kubelet/values.yaml
```diff
@@ -6,12 +6,18 @@ image:
...
 monitoredNamespace: ""
 apiserverCert: 
 apiserverKey: 
 
+taint:
+  enabled: true
+  key: virtual-kubelet.github.io/provider
+  value: "{{ .Values.provider }}"
+  ## `effect` must be `NoSchedule`, `PreferNoSchedule` or `NoExecute`.
+  effect: NoSchedule
```
jlegrone: Enabling the taint and configuring its effect should definitely be options, but I'm ambivalent about allowing configuration of key and value from the helm chart.  Thoughts?

jeremyrickard: I think my preference would be to default them like this, but allow people to override, if needed. I really think that the override is probably a SUPER special case and most people will chose to roll with this, but it doesn't hurt to allow it to be changed.

jeremyrickard:
@jlegrone do you think this is good to go now? You still have WIP on the title? We should update the examples to have the updated taints, but I think other than that this looks good to me.

charts/virtual-kubelet/values.yaml
jeremyrickard: Also, the disable is cool for down the road!

jlegrone:
@jeremyrickard unfortunately I still haven't gotten around to actually validating these changes work beyond compiling the binary.  Would you have time to validate that this works as expected in a real environment?

I can also squash my commits since this went a couple different ways... was just showing my work. 😄

jeremyrickard:
@jlegrone I'll pull your branch and validate it up. 

A squash would be good, needs to be updated as well.

jlegrone:
@jeremyrickard I did make an attempt to fix any existing references to the `azure.com/aci` node taint; are there other places in the docs that should be updated?

jeremyrickard:
@jlegrone i totally spaced on those when I was reviewing. My bad! Thanks.

jeremyrickard:
This works as expected locally for me, LGTM pending @robbiezhang @rbitia

charts/virtual-kubelet/values.yaml
```diff
@@ -6,12 +6,18 @@ image:
...
 nodeName: "virtual-kubelet"
-nodeTaint: "azure.com/aci"
 nodeOsType: "Linux"
 monitoredNamespace: ""
 apiserverCert: 
 apiserverKey: 
 
+taint:
+  enabled: true
+  key: virtual-kubelet.github.io/provider
```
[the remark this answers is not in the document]
jeremyrickard: We could also do something like node.kubernetes.io/virtual-kubelet ? This would be in line with what k8s auto appleis on things.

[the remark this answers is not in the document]
cpuguy83: Not for long 😈 

Honestly it's probably fine for now, but probably shouldn't make a new release without changing it.

cmd/root.go
```diff
@@ -47,7 +47,7 @@ backend implementation allowing users to create kubernetes nodes without running
 This allows users to schedule kubernetes workloads on nodes that aren't running Kubernetes.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		fmt.Println(kubeConfig)
-		f, err := vkubelet.New(nodeName, operatingSystem, kubeNamespace, kubeConfig, taint, provider, providerConfig)
```
robbiezhang: I would recommend to keep the taint in the first version to make the image backward compatible with the old helm chart.
That is - 
if disableTaint is specified in the command, do not add any taint.
if taint is specified in the command, use the old taint
if neither disableTaint nor taint is specified, use the taint defined in new environment variables.

jeremyrickard: Related: we should move away from using `:latest` in the chart and version things for real.

jeremyrickard:
@jlegrone good point from @robbiezhang can you add back the flag and amend the logic per Robbie's comment so we don't break the chart? We should _also_ update the -for-aks chart in the interim until we can update the azure cli.

charts/virtual-kubelet/values.yaml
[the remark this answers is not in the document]
jlegrone: > We could also do something like node.kubernetes.io/virtual-kubelet ? This would be in line with what k8s auto applies on things.

It might be better to go with a virtual-kubelet "owned" prefix instead for things like taints, annotations, etc.  That allows us to use keys that are nice and descriptive, like `provider`, without worrying about naming conflicts.

`virtual-kubelet.io` is 🔥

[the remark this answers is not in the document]
jeremyrickard: I agree with your statement about descriptive name and not worrying about conflict. Let’s go with virtual-kubelet.io

I’ve gone ahead and purchased that domain. We plan to put a static site up with netlify or something, so let’s just use that.

robbiezhang:
LGTM

PR #306, opened 2018-08-09
Add support fort kubelet stats summary

cpuguy83:
This allows hipster or metrics-server to scrape stats from VK for providers that support it.
Of course added an implementation for the Azure provider.

By default this enables a new HTTP listener on `:10255` with a `/stats/summary` endpoint.
hitting this endpoint will give a result like this:

```
{
  "node": {
    "nodeName": "vk-metrics",
    "startTime": "2018-08-08T23:52:26Z"
  },
  "pods": [
    {
      "podRef": {
        "name": "nginx",
        "namespace": "default",
        "uid": "4305386a-9b66-11e8-b3ec-000d3a5dae4d"
      },
      "startTime": "2018-08-01T23:53:34Z",
      "containers": [
        {
          "name": "nginx",
          "startTime": "2018-08-01T23:53:34Z",
          "cpu": {
            "time": "2018-08-09T00:14:00Z",
            "usageNanoCores": 0
          },
          "memory": {
            "time": "2018-08-09T00:14:00Z",
            "usageBytes": 23470080
          },
          "userDefinedMetrics": null
        }
      ],
      "cpu": {
        "time": "2018-08-09T00:14:00Z",
        "usageNanoCores": 0
      },
      "memory": {
        "time": "2018-08-09T00:14:00Z",
        "usageBytes": 23470080
      },
      "network": {
        "time": "2018-08-09T00:14:00Z",
        "name": "",
        "rxBytes": 254,
        "txBytes": 172
      }
    }
  ]
}
```

Everything is great... except it's not actually working. metrics-server is scraping that stats but not storing the results.
This leads to, in the above case, a message in the metrics server logs when you try to fetch metrics from it like `reststorage.go:93] No metrics for pod default/nginx`, even though I know it (or something... something else for me to check) grabbing metrics from VK.

Posting this here for review/discussion while I try to figuring out why the stats are not actually being picked up. Also please let me know if you have some ideas as to what might be causing this.

### DIFF
--- a/providers/azure/aci.go
+++ b/providers/azure/aci.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -26,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/remotecommand"
+	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
 )
 
 // The service account secret mount path.
@@ -47,6 +49,10 @@ type ACIProvider struct {
 	internalIP         string
 	daemonEndpointPort int32
 	diagnostics        *aci.ContainerGroupDiagnostics
+
+	metricsSync     sync.Mutex
+	metricsSyncTime time.Time
+	lastMetric      *stats.Summary
 }
 
 // AuthConfig is the secret returned from an ImageRegistryCredential
@@ -296,11 +302,15 @@ func (p *ACIProvider) CreatePod(pod *v1.Pod) error {
 	// TODO(BJK) containergrouprestartpolicy??
 	_, err = p.aciClient.CreateContainerGroup(
 		p.resourceGroup,
-		fmt.Sprintf("%s-%s", pod.Namespace, pod.Name),
+		containerGroupName(pod),
 		containerGroup,
 	)
 
 	return err
+}
+
+func containerGroupName(pod *v1.Pod) string {
+	return fmt.Sprintf("%s-%s", pod.Namespace, pod.Name)
 }
 
 // UpdatePod is a noop, ACI currently does not support live updates of a pod.
--- a/providers/azure/metrics.go
+++ b/providers/azure/metrics.go
@@ -1,0 +1,219 @@
+package azure
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/virtual-kubelet/virtual-kubelet/providers/azure/client/aci"
+	"golang.org/x/sync/errgroup"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
+)
+
+// GetStatsSummary returns the stats summary for pods running on ACI
+func (p *ACIProvider) GetStatsSummary(ctx context.Context) (summary *stats.Summary, err error) {
+	p.metricsSync.Lock()
+	defer p.metricsSync.Unlock()
+
+	if time.Now().Sub(p.metricsSyncTime) < time.Minute {
+		return p.lastMetric, nil
+	}
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
+
+	defer func() {
+		if err != nil {
+			return
+		}
+		p.lastMetric = summary
+		p.metricsSyncTime = time.Now()
+	}()
+
+	pods := p.resourceManager.GetPods()
+	var errGroup errgroup.Group
+	chResult := make(chan stats.PodStats, len(pods))
+
+	end := time.Now()
+	start := end.Add(-1 * time.Minute)
+
+	sema := make(chan struct{}, 10)
+	for _, pod := range pods {
+		if pod.Status.Phase != v1.PodRunning {
+			continue
+		}
+		errGroup.Go(func() error {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case sema <- struct{}{}:
+			}
+			defer func() {
+				<-sema
+			}()
+
+			cgName := containerGroupName(pod)
+			// cpu/mem and net stats are split because net stats do not support container level detail
+			systemStats, err := p.aciClient.GetContainerGroupMetrics(ctx, p.resourceGroup, cgName, aci.MetricsRequest{
+				Dimension:    "containerName eq '*'",
+				Start:        start,
+				End:          end,
+				Aggregations: []aci.AggregationType{aci.AggregationTypeAverage},
+				Types:        []aci.MetricType{aci.MetricTypeCPUUsage, aci.MetricTypeMemoryUsage},
+			})
+			if err != nil {
+				return errors.Wrapf(err, "error fetching cpu/mem stats for container group %s", cgName)
+			}
+
+			netStats, err := p.aciClient.GetContainerGroupMetrics(ctx, p.resourceGroup, cgName, aci.MetricsRequest{
+				Start:        start,
+				End:          end,
+				Aggregations: []aci.AggregationType{aci.AggregationTypeAverage},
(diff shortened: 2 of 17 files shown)


providers/azure/aci.go
```diff
@@ -216,6 +217,7 @@ func NewACIProvider(config string, rm *manager.ResourceManager, nodeName, operat
 	p.nodeName = nodeName
 	p.internalIP = internalIP
 	p.daemonEndpointPort = daemonEndpointPort
+	p.startTime = time.Now()
```
cpuguy83: Cheating here a bit... really was just checking if this was an issue with metrics-server... doesn't appear to be.

cpuguy83:
Ok, so after adding in some extra logging, it seems metrics-server is not populating pod CPU stats (have to investigate why still as vk is providing it), when the metrics-server API is hit it skips the pod because the pod CPU stats are missing.

All in all, the CPU stats are a bit rough for me to wrap my head around how to report. Right now taking the provided average millicores, converting to nano, and multiplying by 60 (stat interval from ACI) to produce the number of nanoseconds of core time for the container... and also having to reset the core time for each stat request.

The k8s stats summary does support nanocore usage, but metrics-server (and hipster) don't use it from what I can tell.

Anyway, please let me know if you have some suggestions here.

providers/azure/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/virtual-kubelet/virtual-kubelet/providers/azure/client/aci"
+	"golang.org/x/sync/errgroup"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
+)
+
+const startTimeFormat = "2006-01-01 15:04:05 -0700 MST"
```
robbiezhang: -01-02

vkubelet/vkubelet.go
```diff
@@ -151,6 +151,11 @@ func New(nodeName, operatingSystem, namespace, kubeConfig, taint, provider, prov
 	}
 
 	go ApiserverStart(p)
+	if metricsAddr != "" {
```
robbiezhang: since it has a default value :10255, we don't need expect it's empty

vkubelet/apiserver.go
```diff
@@ -32,13 +33,50 @@ func ApiserverStart(provider Provider) {
 	r := mux.NewRouter()
 	r.HandleFunc("/containerLogs/{namespace}/{pod}/{container}", ApiServerHandler).Methods("GET")
 	r.HandleFunc("/exec/{namespace}/{pod}/{container}", ApiServerHandlerExec).Methods("POST")
+	r.HandleFunc("/stats/summary", MetricsSummaryHandler).Methods("GET")
```
robbiezhang: remove as they are added in MetricsServerStart function

cpuguy83:
Ok, so the real main issue is that it wants node level stats... 

https://github.com/kubernetes-incubator/metrics-server/blob/f90c6705d2381ea2db1a6343da6c400bd2ef4cb2/metrics/storage/util/util.go#L27-L34

I'm not sure if any value makes sense for these stats considering that metrics-server assumes all the pods are on the node.

cpuguy83:
I should note metrics-server and heapster were nearly identical codebases until a couple of days ago. So maybe metrics-server HEAD behaves differently, I will have to look into it.

providers/azure/metrics.go
cpuguy83: Ah nice catch, that would be why I'm getting negative uptime :)

cpuguy83:
Ok, this is working.

Please pay extra attention to stat calculations. @robbiezhang I'm sure you have more insight into what that cpu stat is actually saying about the container group.

cpuguy83:
Ok, I've removed WIP from this.

providers/azure/metrics.go
```diff
@@ -0,0 +1,230 @@
...
+
+	defer func() {
+		if err != nil {
+			return
+		}
+		p.lastMetric = summary
+		p.metricsSyncTime = time.Now()
+	}()
+
+	cgs, err := p.aciClient.ListContainerGroups(p.resourceGroup)
```
robbiezhang: I think we should get the pods from resource manager. Ideally, it should be in-sync with the container groups in the resource group. Since it's in-memory, it's more lightweight, and won't be throttled by ARM.

robbiezhang: You need to filter out the terminated ones.

providers/azure/metrics.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var errGroup errgroup.Group
+	chResult := make(chan stats.PodStats, len(cgs.Value))
```
robbiezhang: by value? prefer by pointer to avoid object copy.

providers/azure/metrics.go
```diff
@@ -0,0 +1,230 @@
...
+			data := entry.Data[len(entry.Data)-1] // get only the last entry
+			switch m.Desc.Value {
+			case aci.MetricTypeCPUUsage:
+				if cs.CPU == nil {
+					cs.CPU = &stats.CPUStats{}
+				}
+
+				// average is the average number of millicores over a 1 minute interval (which is the interval we are pulling the stats for)
+				nanoCores := uint64(data.Average * 1000000)
+				usuageNanoSeconds := nanoCores * 60
```
robbiezhang: typo 'usage'

vkubelet/apiserver.go
```diff
@@ -34,11 +37,58 @@ func ApiserverStart(provider Provider) {
 	r.HandleFunc("/exec/{namespace}/{pod}/{container}", ApiServerHandlerExec).Methods("POST")
 	r.NotFoundHandler = http.HandlerFunc(NotFound)
 
+	if metricsAddr != "" {
+		go MetricsServerStart(metricsAddr)
+	} else {
+		log.Println("skipping metrics server startup sicne no address was provided")
```
robbiezhang: typo: ... startup 'since' no address ...

vkubelet/apiserver.go
```diff
@@ -34,11 +37,58 @@ func ApiserverStart(provider Provider) {
...
 }
 
+// MetricsServerStart starts an HTTP server on the provided addr for serving the kubelset summary stats API.
+// TLS is never enabled on this endpoint.
+func MetricsServerStart(addr string) {
+	r := mux.NewRouter()
+	r.HandleFunc("/stats/summary", MetricsSummaryHandler).Methods("GET")
+	r.HandleFunc("/stats/summary/", MetricsSummaryHandler).Methods("GET")
+	r.NotFoundHandler = http.HandlerFunc(NotFound)
+	http.ListenAndServe(addr, r)
```
robbiezhang: log the error

vkubelet/apiserver.go
```diff
@@ -34,11 +37,58 @@ func ApiserverStart(provider Provider) {
...
+
+	b, err := json.Marshal(stats)
+	if err != nil {
+		log.Println("could not marshal stats:", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if _, err := w.Write(b); err != nil {
+		io.WriteString(w, err.Error())
```
robbiezhang: what if it still fail?

vkubelet/apiserver.go
```diff
@@ -34,11 +37,58 @@ func ApiserverStart(provider Provider) {
 	r.HandleFunc("/exec/{namespace}/{pod}/{container}", ApiServerHandlerExec).Methods("POST")
 	r.NotFoundHandler = http.HandlerFunc(NotFound)
 
+	if metricsAddr != "" {
+		go MetricsServerStart(metricsAddr)
```
robbiezhang: can we move the Metrics Server into a new file, and start it from the vkubelet.go?

cpuguy83: The reason I didn't do this for now is right now `ApiServerStart` is setting a global (for provider) and these will race with each other.
I'd rather refactor this in a separate PR.

cpuguy83:
This is all fixed up.

cpuguy83:
Yeah, that *might* be easier 😃

cpuguy83:
This is all integrated, merging.

mcraken:
Is the stats endpoint running now for the virtual kubelet? I see from the metrics server that it is still not able to scrap metrics because it can not find the stats endpoint. The virtual kubelet version is 1.11.2.

Issue #256, opened 2018-07-11
proposal: provider specific label


mqliang:
Does it sounds good to add a `Labels() (map[string]string)` method to provider interface, and merge the provider specific labels and VK generic labels (https://github.com/virtual-kubelet/virtual-kubelet/blob/master/vkubelet/vkubelet.go#L174-L179)

provider specific labels can been read from provider config files. This feature is very useful for using the mock provider to simulate a bunch of heterogeneous nodes. (add "az", "rack numbers" information as labels.)

mqliang:
@robbiezhang
cc @glikson

robbiezhang:
Agree. #221 was created to discuss this issue. We also have such requirement, and we believe other providers might also need this feature. 
Some ideas:
1. Do we allow the provider to overwrite the common VK labels?
2. Shall we validate the label upon k8s label name convention in the VK? https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
3. How many labels can a node have? Shall we set a limit to the provider?

rbitia:
We use taints to solve the same purpose. Is there a reason you would rather use labels?

pires:
The `Provider` interface defines a func `ConfiguredNode` that the provider can use to handle the `Node` object before its registration. An example on how to use it is included in the `MockProvider`

https://github.com/virtual-kubelet/virtual-kubelet/blob/0d1f6f162509a9b4b73f1d23ac133541337545b0/cmd/virtual-kubelet/internal/provider/mock/mock.go#L346-L347

Another example is how `systemk` does it

https://github.com/virtual-kubelet/systemk/blob/b1e585c6fce0b9cd86d0f109997482e8f515e118/internal/provider/node.go#L33-L70

PR #311, opened 2018-08-12
support provider specific labels

mqliang:
ref https://github.com/virtual-kubelet/virtual-kubelet/issues/256

@robbiezhang
cc @glikson


vkubelet/vkubelet.go
```diff
@@ -147,16 +147,24 @@ func (s *Server) registerNode() error {
...
 
+	labels := map[string]string{
+		"type":                                                    "virtual-kubelet",
+		"kubernetes.io/role":                                      "agent",
+		"beta.kubernetes.io/os":                                   strings.ToLower(s.provider.OperatingSystem()),
+		"kubernetes.io/hostname":                                  s.nodeName,
+		"alpha.service-controller.kubernetes.io/exclude-balancer": "true",
+	}
+	// add provider specific labels
+	if s.provider.Labels() != nil {
```
cpuguy83: No need to check nil here here.

vkubelet/vkubelet.go
```diff
@@ -147,16 +147,24 @@ func (s *Server) registerNode() error {
...
+		for k, v := range s.provider.Labels() {
+			labels[k] = v
```
cpuguy83: Should we make sure the provider is not overwriting labels?

cpuguy83:
This has many merge conflicts, and also we are pushing this functionality up the stack a bit into the CLI implementation rather than in the core framework.
Basically in the core we just expect the caller to give us as node object, we don't care about all the details of it.

See https://github.com/virtual-kubelet/virtual-kubelet/pull/526 for more details.

Let me know if you want to tackle this still.
Thanks! 🙇  👼

Issue #277, opened 2018-07-26
proposal: flag to disable pod statuses updating


mqliang:
Hey, does it sounds good to add a flag to disable pod statuses updating?

In our use case, we use VK (with the mock provider) to simulate nodes, and we have a controller to simulate the completion of pods, in other words, we have a controller to update pod status, which may conflict with VK because VK will always update pod status as "Running", even if our controller update status as "Succeeded", it will been overwrite by VK.

I am glad to send a PR if this sounds good.

@robbiezhang
cc @glikson

robbiezhang:
@mqliang I think it's not a generic scenario for VK. Your issue is mainly because the mock provider doesn't collaborate with your controller well - both of them updates the pod status, and cause conflict. I don't know whether other provider have such requirement, so I would recommend you make the change
in the mock provider only. An easy trick is to add a flag to always return null in GetPodStatus. More fancy one is to have a new provider/allow the mock provider to collaborate with your controller by other means, i.e. pod annotation. Just my $.02 for your reference.

rbitia:
This isn't in the scope for VK, so we are closing the issue for the time being! Thanks for the issue!

PR #313, opened 2018-08-13
add a flag to allow disable pod status updating

mqliang:
ref https://github.com/virtual-kubelet/virtual-kubelet/issues/277

@robbiezhang
cc @glikson


cpuguy83:
I don't feel like this is a good fit since the main purpose of VK is to do the thing that this flag is disabling.

If you are having issues with testing, maybe we can address those in a different way?
Let's discuss on the open issue.

Closing for now. 🙇 👼

Issue #234, opened 2018-06-21
Build is broken on MacOS


sabbour:
---

### Environment summary

```
System Version: macOS 10.13.5 (17F77)
Kernel Version: Darwin 17.6.0
```

### Issue Details

Build is broken on MacOS

```
Building...
# github.com/virtual-kubelet/virtual-kubelet/providers/cri
providers/cri/cri.go:809:9: undefined: syscall.Sysinfo_t
providers/cri/cri.go:810:9: undefined: syscall.Sysinfo
# github.com/virtual-kubelet/virtual-kubelet/vendor/github.com/vmware/vic/lib/apiservers/engine/network
vendor/github.com/vmware/vic/lib/apiservers/engine/network/utils.go:62:12: undefined: netlink.LinkByName
vendor/github.com/vmware/vic/lib/apiservers/engine/network/utils.go:64:12: undefined: netlink.LinkByAlias
vendor/github.com/vmware/vic/lib/apiservers/engine/network/utils.go:281:12: undefined: netlink.LinkByName
vendor/github.com/vmware/vic/lib/apiservers/engine/network/utils.go:286:36: undefined: netlink.FAMILY_V4
make: *** [build] Error 2
```

### Repo Steps

```sh
mkdir -p cd $GOPATH/src/github.com/virtual-kubelet
cd $GOPATH/src/github.com/virtual-kubelet
git clone https://github.com/virtual-kubelet/virtual-kubelet.git
```

then apply the fix in https://github.com/virtual-kubelet/virtual-kubelet/pull/231
then try to build

```sh
cd $GOPATH/src/github.com/virtual-kubelet/virtual-kubelet
make deps
make build
```

sabbour:
Pinging @sflxn and @corrieb since it appears both are from their commits.
Is this supposed to build on MacOS?

ofiliz:
I don't think we care about macos. That is neither a container-capable operating system nor a Kubernetes platform. On the other hand, these changes probably broke Windows, too, which is both a container-capable operating system and a Kubernetes platform.

You can cross-compile for Linux by setting your env GOOS=linux.

sflxn:
@sabbour Yeah, my code will not build on Mac or Windows unless you do as @ofiliz suggested.  Our code wasn't designed to run on Mac or Windows.  Another suggestion is to use a linux VM to do all your work.

sabbour:
The `export GOOS=linux` workaround did the trick. Thanks.

[...]

PR #320, opened 2018-08-15
Fixing the broken build in OSX

ageekymonk:
Fixing the broken build as raised in #234 .  Let me know if there is anything i need to do.


robbiezhang:
LGTM

PR #323, opened 2018-08-16
Use standard logging package

cpuguy83:


vkubelet/lookup_windows.go
```diff
@@ -43,7 +44,8 @@ func lookupProvider(provider, providerConfig string, rm *manager.ResourceManager
 	case "sfmesh":
 		return sfmesh.NewSFMeshProvider(rm, nodeName, operatingSystem, internalIP, daemonEndpointPort)
 	default:
-		fmt.Printf("Provider '%s' is not supported\n", provider)
+		// TODO(@cpuguy83): perahps this should return an error
```
robbiezhang: typo: perhaps
let's return an error.

vkubelet/vkubelet.go
```diff
@@ -98,7 +103,8 @@ func New(nodeName, operatingSystem, namespace, kubeConfig, provider, providerCon
 	case "PreferNoSchedule":
 		vkTaintEffect = corev1.TaintEffectPreferNoSchedule
 	default:
-		fmt.Printf("Taint effect '%s' is not supported\n", vkTaintEffectEnv)
+		// TODO(@cpuguy83): should this return early as an error?
+		log.G(context.TODO()).Errorf("Taint effect '%s' is not supported\n", vkTaintEffectEnv)
```
robbiezhang: agree. we should error out early.

PR #326, opened 2018-08-20
Plumb context to providers

cpuguy83:
This allows us to pass tracing, logging, and potentially other request scoped context to a provider.


cpuguy83:
Ok, and now it's actually green.

PR #327, opened 2018-08-20
Remove unused method from providers.

cpuguy83:
Looks like this was possibly part of the provider interface at some
point, but is no longer needed.

### DIFF
--- a/providers/aws/provider.go
+++ b/providers/aws/provider.go
@@ -178,11 +178,6 @@ func (p *FargateProvider) GetContainerLogs(namespace, podName, containerName str
 	return p.cluster.GetContainerLogs(namespace, podName, containerName, tail)
 }
 
-// GetPodFullName retrieves the full pod name as defined in the provider context.
-func (p *FargateProvider) GetPodFullName(namespace string, pod string) string {
-	return ""
-}
-
 // ExecInContainer executes a command in a container in the pod, copying data
 // between in/out/err and the container's stdin/stdout/stderr.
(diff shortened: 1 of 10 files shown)


robbiezhang:
The GetPodFullName should be used here ( [fmt.Sprintf("%s-%s", namespace, pod)] ):
https://github.com/virtual-kubelet/virtual-kubelet/blob/bb8ba567e3a33c827e0ffa56755e3fdeabb9af53/vkubelet/apiserver.go#L179

We introduce the GetPodFullName funtion in the provider interface, because the ServeExec function only pass through the "FullName" to the provider instead of Namespace and PodName. The provider can implement it based on their own convention.

For example, ACI uses [fmt.Sprintf("%s-%s", namespace, pod)], and it is passed in as the 1st parameter (name):
https://github.com/virtual-kubelet/virtual-kubelet/blob/bb8ba567e3a33c827e0ffa56755e3fdeabb9af53/providers/azure/aci.go#L378-L390
Thus, it can get container group from RP by the name directly.

So, the correct use of the GetPodFullName is to replace [fmt.Sprintf("%s-%s", namespace, pod)] with s.p.GetPodFullName(namespace, pod).

However, I'm open to any idea since ACI is the only provider, which implemented the `exec` function.

cpuguy83:
Perhaps we should move that call down into the provider, otherwise the API is making assumptions about the provider.

cpuguy83:
Going to close this for now,

PR #330, opened 2018-08-22
Implement port forward (new interface, added to ACI provider)

### LINKED ISSUE #170: Feature: Implement Port forwarding
**Alpha Release**

There's a need to be able to open a port on the host running kubectl and tunnel through to a port on a Pod. We need to implement this capability in VK, which will likely require an update to the Provider interface.

jeremyrickard:
This PR adds a new interface to support port forwarding. It takes the approach of adding a smaller interface, with the API server checking to see if a provider implements that smaller interface. 

The plumbing from the API server to the provider endpoint is vendored from k8s. 

The implementation of port forward in the ACI module is pretty close to what the Docker Shim does with socat in the normal kubelet, with identical user experience. 

Closes #170

### DIFF
--- a/providers/azure/aci.go
+++ b/providers/azure/aci.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"reflect"
@@ -1150,4 +1151,50 @@ func getACIEnvVar(e v1.EnvVar) aci.EnvironmentVariable {
 		}
 	}
 	return envVar
+}
+
+func containsPort(exposedPorts []aci.Port, port int32) bool {
+	for _, p := range exposedPorts {
+		if p.Port == port {
+			return true
+		}
+	}
+	return false
+}
+
+func (p *ACIProvider) PortForward(name string, uid types.UID, port int32, stream io.ReadWriteCloser) error {
+	cg, err, _ := p.aciClient.GetContainerGroup(p.resourceGroup, name)
+	if err != nil {
+		return fmt.Errorf("unable to obtain pod info: %s", err)
+	}
+	if cg.IPAddress == nil {
+		return fmt.Errorf("no IP Address exposed for pod")
+	}
+	if !containsPort(cg.IPAddress.Ports, port) {
+		return fmt.Errorf("port not exposed: %d", port)
+	}
+	fmt.Printf("conencting to %s", fmt.Sprintf("%s:%d", cg.IPAddress.IP, port))
+	// open a connection to IP Address on port
+	conn, err := net.Dial("tcp", fmt.Sprintf("%s:%d", cg.IPAddress.IP, port))
+	if err != nil {
+		return fmt.Errorf("unable to open connection: %s", err)
+	}
+	defer conn.Close()
+	errs := make(chan error)
+	go func() {
+		defer stream.Close()
+		fmt.Printf("Writing to stream")
+		_, err := io.Copy(stream, conn)
+		if err != nil {
+			errs <- err
+		} else {
+			errs <- nil
+		}
+	}()
+	buf := make([]byte, 1024)
+	_, err = io.CopyBuffer(conn, stream, buf)
+	if err != nil {
+		return err
+	}
+	return <-errs
 }
--- a/providers/azure/aci.go
+++ b/providers/azure/aci.go
@@ -36,6 +36,8 @@ const serviceAccountSecretMountPath = "/var/run/secrets/kubernetes.io/serviceacc
 
 const virtualKubeletDNSNameLabel = "virtualkubelet.io/dnsnamelabel"
 
+var pfPool *sync.Pool
+
 // ACIProvider implements the virtual-kubelet provider interface and communicates with Azure's ACI APIs.
 type ACIProvider struct {
 	aciClient          *aci.Client
@@ -99,6 +101,11 @@ func NewACIProvider(config string, rm *manager.ResourceManager, nodeName, operat
 	var p ACIProvider
 	var err error
 
+	pfPool = &sync.Pool{
+		New: func() interface{} {
+			return make([]byte, 32*1024)
+		},
+	}
 	p.resourceManager = rm
 
 	if config != "" {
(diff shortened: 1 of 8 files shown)


cpuguy83:
Should we be able to generalize this for all providers by reading the IP from the pod?

cpuguy83:
Hmm yeah you're right it's probably not that easy to to generalize.

providers/azure/aci.go
```diff
@@ -1151,3 +1152,54 @@ func getACIEnvVar(e v1.EnvVar) aci.EnvironmentVariable {
...
+		return fmt.Errorf("unable to open connection: %s", err)
+	}
+	defer conn.Close()
+	done := make(chan bool)
+	errs := make(chan error)
+	go func() {
+		defer stream.Close()
+		fmt.Printf("Writing to stream")
+		_, err := io.Copy(stream, conn)
+		if err != nil {
```
cpuguy83: Can send a nil error through the channel instead of dealing with a "done" channel.

providers/azure/aci.go
```diff
@@ -1151,3 +1152,54 @@ func getACIEnvVar(e v1.EnvVar) aci.EnvironmentVariable {
...
+			errs <- err
+		} else {
+			done <- true
+		}
+	}()
+	_, err = io.Copy(conn, stream)
+	if err != nil {
+		return err
+	}
+	select {
```
cpuguy83: This probably needs to implement `CloseWrite` and `CloseRead` so that you can signal the grouting to exit when this side is done copying (and vice-versa).

providers/azure/aci.go
```diff
@@ -1151,3 +1152,54 @@ func getACIEnvVar(e v1.EnvVar) aci.EnvironmentVariable {
...
+		defer stream.Close()
+		fmt.Printf("Writing to stream")
+		_, err := io.Copy(stream, conn)
+		if err != nil {
+			errs <- err
+		} else {
+			done <- true
+		}
+	}()
+	_, err = io.Copy(conn, stream)
```
cpuguy83: May be premature here, but since this is a long running daemon it may be prudent to use a buffer pool here and `io.CopyBuf`.

providers/azure/aci.go
jeremyrickard: 👍

providers/azure/aci.go
jeremyrickard: Cool, will dive into that.

providers/azure/aci.go
jeremyrickard: I debated that and decided that the allocation in Copy probably wasn't that big of a deal, but swapping over to CopyBuf won't hurt so I'll update.

providers/azure/aci.go
```diff
@@ -1151,3 +1152,56 @@ func getACIEnvVar(e v1.EnvVar) aci.EnvironmentVariable {
...
+	if err != nil {
+		return fmt.Errorf("unable to open connection: %s", err)
+	}
+	defer conn.Close()
+	errs := make(chan error)
+	go func() {
+		fmt.Printf("Writing to stream")
+		_, err := io.Copy(stream, conn)
+		conn.CloseRead()
+		if err != nil {
```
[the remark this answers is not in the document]
jeremyrickard: ahhh yeah, forgot to remove this check when i updated to just one channel. thanks!

providers/azure/aci.go
```diff
@@ -1151,3 +1159,55 @@ func getACIEnvVar(e v1.EnvVar) aci.EnvironmentVariable {
...
+		return fmt.Errorf("unable to resolve remote address: %s", err)
+	}
+	conn, err := net.DialTCP("tcp", nil, remoteAddr)
+	if err != nil {
+		return fmt.Errorf("unable to open connection: %s", err)
+	}
+	defer conn.Close()
+	errs := make(chan error, 1)
+	go func() {
+		fmt.Printf("Writing to stream")
```
robbiezhang: use log

vkubelet/apiserver.go
```diff
@@ -178,3 +180,16 @@ func (s *KubeletServer) ApiServerHandlerExec(w http.ResponseWriter, req *http.Re
...
 	remotecommand.ServeExec(w, req, s.p, fmt.Sprintf("%s-%s", namespace, pod), "", container, command, streamOpts, idleTimeout, streamCreationTimeout, supportedStreamProtocols)
 }
+
+func (s *KubeletServer) ApiServerHandlerPortForward(w http.ResponseWriter, req *http.Request) {
+	pf, ok := s.p.(PortForwarder)
+	if ok {
+		vars := mux.Vars(req)
+		namespace := vars["namespace"]
+		pod := vars["pod"]
+		portforward.ServePortForward(w, req, pf, fmt.Sprintf("%s-%s", namespace, pod), "", nil, time.Second*60, time.Second*60, []string{})
```
robbiezhang: suggest to call p.GetPodFullName instead of fmt.Sprintf("%s-%s", namespace, pod)

cpuguy83:
>  I made the assumption that not every provider creates pods that might be fronted by a public IP

I think this is even true for ACI.
Should we implement this in the ACI provider if there isn't a guarantee that VK can actually reach the pod IP?

jeremyrickard:
> > I made the assumption that not every provider creates pods that might be fronted by a public IP
> 
> I think this is even true for ACI.
> Should we implement this in the ACI provider if there isn't a guarantee that VK can actually reach the pod IP?

Yeah, that's a good point.

rbitia:
@jeremyrickard fix the conflicts!

jeremyrickard:
@rbitia I think this needs some rework behind fixing the conflicts. Probably implementing the logic in the framework instead of the provider @cpuguy83?

jeremyrickard:
I’ll work on this next week though!

pires:
@jeremyrickard please ping me over slack as I'd like to help here, even if it's just testing.
And speaking of which, adding one or two e2e tests would be really cool.

srrengar:
hi @jeremyrickard were you able to make the changes you mentioned on this?

cpuguy83:
This has gone stale.
For now I'm going to close this and I've assigned someone from the ACI team to look into this further.
I think having VK trying to proxy these streams for ACI is not a great situation to be in and likely needs some service from the Azure API to enable this in a way that works most of the time verses only some of the time.

Thanks Jeremy!

PR #355, opened 2018-09-20
Add support for tracing via OpenCencus

cpuguy83:
This adds a few flags for configuring the tracer.
Includes support for jaeger tracing (built into OC).

Example tracing output (though span names are slightly different now):
image


jeremyrickard:
@cpuguy83 The current AppInsights functiality with the go library will probably work, however there is like a 3-5 minute latency (we observed this during the ignite demo work). To get “live” metrics we’d need to use that opencensus forwarder, but we’d need to run it per VK pod and I’m not sure that’s something we’d want to consider ATM.

cpuguy83:
This is now rebased.

PR #360, opened 2018-09-25
Refactor provider init

cpuguy83:
- Refactor provider init
i. Moves provider init out of vkubelet setup, instead preferring to initialize vkubelet with a provider.
- Split API server configuration from setup.
i. This makes sure that configuration (which is done primarily through env vars) is separate from actually standing up the servers
ii. This also makes sure to abort daemon initialization if the API servers are not able to start


cmd/root.go
```diff
@@ -50,11 +66,21 @@ var RootCmd = &cobra.Command{
...
+			MetricsAddr:     metricsAddr,
+			Provider:        p,
+			ResourceManager: rm,
+			APIConfig:       apiConfig,
+		})
 		if err != nil {
 			log.L.WithError(err).Fatal("Error initializing virtual kubelet")
 		}
-		if err := f.Run(context.Background()); err != nil {
+		if err := f.Run(ctx); err != nil {
```
robbiezhang: will it fix #363

cmd/root.go
```diff
@@ -155,4 +181,61 @@ func initConfig() {
 	})
 	logger.Level = level
 	log.L = logger
+
+	if !disableTaint {
+		taint, err = getTaint(taintKey, provider)
+		if err != nil {
+			logger.WithError(err).Fatal("Error setting up desired kubernetes node tain")
```
robbiezhang: /tain/taint

vkubelet/vkubelet.go
```diff
@@ -189,7 +138,6 @@ func (s *Server) registerNode(ctx context.Context) error {
 				Architecture:    "amd64",
 				KubeletVersion:  "v1.11.2",
 			},
-			Capacity:        s.provider.Capacity(ctx),
```
robbiezhang: why Capacity is not set?

cpuguy83: Good question. Don't remember even touching this. Nice catch.

cmd/client.go
```diff
@@ -0,0 +1,35 @@
...
+		// Get the kubeconfig from the filepath.
+		config, err = clientcmd.BuildConfigFromFlags("", configPath)
+		if err != nil {
+			return nil, errors.Wrap(err, "error building client config")
+		}
+	} else {
+		// Set to in-cluster config.
+		config, err = rest.InClusterConfig()
+		if err != nil {
+			return nil, errors.Wrap(err, "error building in cliuster config")
```
robbiezhang: /cliuster/cluster/

providers/register/register.go
```diff
@@ -0,0 +1,36 @@
...
+	ResourceManager *manager.ResourceManager
+}
+
+type initFunc func(InitConfig) (providers.Provider, error)
+
+// Get gets the provider specified by the given name
+func Get(name string, cfg InitConfig) (providers.Provider, error) {
+	f, ok := providerInits[name]
+	if !ok {
+		return nil, strongerrors.NotFound(errors.New("provider not found"))
```
robbiezhang: include the provider name in the message

vkubelet/vkubelet.go
```diff
@@ -32,123 +31,73 @@ type Server struct {
...
-		}
-	} else {
-		// Set to in-cluster config.
-		config, err = rest.InClusterConfig()
-		if err != nil {
-			return nil, err
-		}
-	}
+type APIConfig struct {
+	Cert string
```
robbiezhang: CertPath?

vkubelet/vkubelet.go
```diff
@@ -32,123 +31,73 @@ type Server struct {
...
+	Key  string
```
robbiezhang: KeyPath?

vkubelet/vkubelet.go
```diff
@@ -32,123 +31,73 @@ type Server struct {
...
-	}
+	ctx = log.WithLogger(ctx, log.G(ctx))
 
-	rm, err := manager.NewResourceManager(clientset)
+	apiL, err := net.Listen("tcp", cfg.APIConfig.Addr)
 	if err != nil {
-		return nil, pkgerrors.Wrap(err, "error creating resource manager")
+		return nil, pkgerrors.Wrap(err, "error setting up API listener")
 	}
+	defer func() {
```
robbiezhang: this seems wrong. the apiL should not be closed in this method.

vkubelet/vkubelet.go
```diff
@@ -32,123 +31,73 @@ type Server struct {
...
-	var defaultTaintValue string
-	if taintKey != "" {
-		defaultTaintKey = taintKey
-		defaultTaintValue = ""
+	if cfg.MetricsAddr != "" {
+		metricsL, err := net.Listen("tcp", cfg.MetricsAddr)
+		if err != nil {
+			return nil, pkgerrors.Wrap(err, "error setting up metrics listener")
+		}
+		defer func() {
```
robbiezhang: same here. metricsL should not be closed in the function

vkubelet/vkubelet.go
cpuguy83: This is only closing if the main function returns an error.

providers/register/register.go
```diff
@@ -0,0 +1,36 @@
...
+	OperatingSystem string
+	InternalIP      string
+	DaemonPort      int32
+	ResourceManager *manager.ResourceManager
+}
+
+type initFunc func(InitConfig) (providers.Provider, error)
+
+// Get gets the provider specified by the given name
+func Get(name string, cfg InitConfig) (providers.Provider, error) {
```
robbiezhang: change the function name to GetProvider?

cmd/root.go
cpuguy83: I don't think so. Wasn't aware of the problem. Will take a look after this.

cpuguy83:
This is all updated.
@jeremyrickard Could you take a look?

cmd/root.go
```diff
@@ -16,18 +16,29 @@ package cmd
...
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/cpuguy83/strongerrors"
```
jeremyrickard: this library is cool

PR #366, opened 2018-10-04
Pass go build tags from environment.

cpuguy83:
Can now do something like `make VK_BUILD_TAGS=no_jaeger_exporter build`
and jaeger support will not be compiled in.

Currently there are build tags to omit each provider and tracing exporter.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,28 +4,23 @@ ENV PATH /go/bin:/usr/local/go/bin:$PATH
 ENV GOPATH /go
 
 RUN	apk add --no-cache \
-	ca-certificates
-
-COPY . /go/src/github.com/virtual-kubelet/virtual-kubelet
-
-RUN set -x \
-	&& apk add --no-cache --virtual .build-deps \
+	ca-certificates \
+	--virtual .build-deps \
 		git \
 		gcc \
 		libc-dev \
 		libgcc \
-        make \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -17,12 +18,12 @@ all: test build
 # safebuild builds inside a docker container with no clingons from your $GOPATH
 safebuild:
 	@echo "Building..."
-	$Q docker build -t $(DOCKER_IMAGE):$(VERSION) .
+	$Q docker build --build-arg BUILD_TAGS=$(tags) -t $(DOCKER_IMAGE):$(VERSION) .
```
robbiezhang: where is $(tags) defined?

cpuguy83: Ah dang I changed it without updating the references.

cpuguy83:
Fixed.

Dockerfile
```diff
@@ -4,28 +4,23 @@ ENV PATH /go/bin:/usr/local/go/bin:$PATH
...
 RUN	apk add --no-cache \
-	ca-certificates
-
-COPY . /go/src/github.com/virtual-kubelet/virtual-kubelet
-
-RUN set -x \
-	&& apk add --no-cache --virtual .build-deps \
+	ca-certificates \
+	--virtual .build-deps \
 		git \
```
robbiezhang: fix the indent

cpuguy83: Done.

PR #368, opened 2018-10-06
Minor refactorings

cpuguy83:
Mostly just re-arranging things, fixing some doc strings.
The biggest thing is making that reconciliation loop easier to read.

PR #373, opened 2018-10-10
[VK] Use Cache controller and Make create/delete pod Concurrently

robbiezhang:
1. Update the virtual-kubelet to use cache controller.
2. Instead of calling reconcile for each pod event, calls createPod/deletePod accordingly in parallel.
3. Add a pod synchronizer pool (size of 10) to handle the create/delete pod operation.


vkubelet/vkubelet.go
```diff
@@ -165,77 +169,77 @@ func (s *Server) registerNode(ctx context.Context) error {
 
 // Run starts the server, registers it with Kubernetes and begins watching/reconciling the cluster.
 // Run will block until Stop is called or a SIGINT or SIGTERM signal is received.
-func (s *Server) Run(ctx context.Context) error {
-	shouldStop := false
+func (s *Server) Run(ctx context.Context) {
```
cpuguy83: We should still let this return an error so the CLI can exit non-zero.

vkubelet/vkubelet.go
```diff
@@ -165,77 +169,77 @@ func (s *Server) registerNode(ctx context.Context) error {
...
+	)
 
-// Stop shutsdown the server.
-// It does not shutdown pods assigned to the virtual node.
-func (s *Server) Stop() {
-	if s.podWatcher != nil {
-		s.podWatcher.Stop()
+	stopCh := make(chan struct{})
+
+	sig := make(chan os.Signal, 1)
```
cpuguy83: See https://github.com/virtual-kubelet/virtual-kubelet/pull/368/files#diff-ff7686b39bf90dc2520886fb874371a4R92 where this is moved up into the CLI (and uses context cancelation instead).

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
 	)
 }
 
+func (s *Server) onAddPod(ctx context.Context, obj interface{}) {
+	pod := obj.(*corev1.Pod)
```
cpuguy83: Getting a tracing span here would be great so we can measure latency on adding sending to the queue.

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+
+	logger.Debugf("Receive added pod '%s/%s' ", pod.GetNamespace(), pod.GetName())
+
+	if s.resourceManager.UpdatePod(pod) {
+		s.podCh <- pod
+	}
+}
+
+func (s *Server) onUpdatePod(ctx context.Context, obj interface{}) {
+	pod := obj.(*corev1.Pod)
```
cpuguy83: tracing span

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+
+	logger.Debugf("Receive updated pod '%s/%s'", pod.GetNamespace(), pod.GetName())
+
+	if s.resourceManager.UpdatePod(pod) {
+		s.podCh <- pod
+	}
+}
+
+func (s *Server) onDeletePod(ctx context.Context, obj interface{}) {
+	pod := obj.(*corev1.Pod)
```
cpuguy83: tracing span

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+		s.podCh <- pod
+	}
+}
+
+func (s *Server) startPodSynchronizer(ctx context.Context, id int) {
+	logger := log.G(ctx).WithField("podSynchronizer", id)
+	logger.Debug("Start pod synchronizer")
+
+	for pod := range s.podCh {
+		if pod.DeletionTimestamp != nil {
```
cpuguy83: tracing span per loop iteration

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+		s.podCh <- pod
+	}
+}
+
+func (s *Server) onUpdatePod(ctx context.Context, obj interface{}) {
+	pod := obj.(*corev1.Pod)
+
+	logger := log.G(ctx).WithField("method", "onUpdatePod")
+	if pod == nil {
+		logger.Error("obj is not a valid pod:", obj)
```
cpuguy83: Errorf here.

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+	}
+}
+
+func (s *Server) startPodSynchronizer(ctx context.Context, id int) {
+	logger := log.G(ctx).WithField("podSynchronizer", id)
+	logger.Debug("Start pod synchronizer")
+
+	for pod := range s.podCh {
+		if pod.DeletionTimestamp != nil {
+			logger.Infof("Deleting pod '%s/%s'", pod.GetNamespace(), pod.GetName())
```
cpuguy83: This seems like a debug message.

cpuguy83: But really good to have as a span annotation.

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+			if err := s.deletePod(ctx, pod); err != nil {
+				logger.WithError(err).Errorf("Failed to delete pod '%s/%s'", pod.GetNamespace(), pod.GetName())
+			}
+		} else {
+			logger.Infof("Creating pod '%s/%s'", pod.GetNamespace(), pod.GetName())
+			if err := s.createPod(ctx, pod); err != nil {
+				logger.WithError(err).Errorf("Failed to create pod '%s/%s'", pod.GetNamespace(), pod.GetName())
```
cpuguy83: `WithField("pod", pod.GetName()).WithField("namespace", pod.GetNamespace())`

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+	logger.Debug("Start pod synchronizer")
+
+	for pod := range s.podCh {
+		if pod.DeletionTimestamp != nil {
+			logger.Infof("Deleting pod '%s/%s'", pod.GetNamespace(), pod.GetName())
+			if err := s.deletePod(ctx, pod); err != nil {
+				logger.WithError(err).Errorf("Failed to delete pod '%s/%s'", pod.GetNamespace(), pod.GetName())
+			}
+		} else {
+			logger.Infof("Creating pod '%s/%s'", pod.GetNamespace(), pod.GetName())
```
cpuguy83: Seems like a debug message.

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+	if s.resourceManager.DeletePod(pod) {
+		s.podCh <- pod
+	}
+}
+
+func (s *Server) startPodSynchronizer(ctx context.Context, id int) {
+	logger := log.G(ctx).WithField("podSynchronizer", id)
+	logger.Debug("Start pod synchronizer")
+
+	for pod := range s.podCh {
```
cpuguy83: Doesn't looks like anything is closing this channel.
We should either make sure it gets closed and/or use `for { select {}}` to select on `<-ctx.Done()` and `<-podCh`

vkubelet/vkubelet.go
robbiezhang: OK. Seems like lot of conflicts need to handle. I'll review #368 first.

vkubelet/vkubelet.go
```diff
@@ -425,6 +429,73 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+	if s.resourceManager.UpdatePod(pod) {
+		s.podCh <- pod
+	}
+}
+
+func (s *Server) onUpdatePod(ctx context.Context, obj interface{}) {
+	pod := obj.(*corev1.Pod)
+
+	logger := log.G(ctx).WithField("method", "onUpdatePod")
+	if pod == nil {
```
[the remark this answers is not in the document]
robbiezhang: it should not happen, but just in case

cpuguy83:
```
vkubelet/pod.go:137:13: s.populateSecretsAndConfigMapsInEnv undefined (type *Server has no field or method populateSecretsAndConfigMapsInEnv)
vkubelet/pod.go:240:18: undefined: fields
vkubelet/pod.go:253:17: undefined: cache
vkubelet/pod.go:254:18: undefined: cache
vkubelet/vkubelet.go:14:2: imported and not used: "github.com/virtual-kubelet/virtual-kubelet/vendor/k8s.io/apimachinery/pkg/apis/meta/v1" as metav1
vkubelet/vkubelet.go:15:2: imported and not used: "github.com/virtual-kubelet/virtual-kubelet/vendor/k8s.io/apimachinery/pkg/fields"
vkubelet/vkubelet.go:16:2: imported and not used: "github.com/virtual-kubelet/virtual-kubelet/vendor/k8s.io/apimachinery/pkg/runtime"
vkubelet/vkubelet.go:17:2: imported and not used: "github.com/virtual-kubelet/virtual-kubelet/vendor/k8s.io/apimachinery/pkg/watch"
vkubelet/vkubelet.go:19:2: imported and not used: "github.com/virtual-kubelet/virtual-kubelet/vendor/k8s.io/client-go/tools/cache"
vkubelet/vkubelet.go:125:4: break is not in a loop, switch, or select
vkubelet/pod.go:254:18: too many errors
```

vkubelet/vkubelet.go
```diff
@@ -28,8 +25,8 @@ type Server struct {
 	k8sClient       *kubernetes.Clientset
 	taint           *corev1.Taint
 	provider        providers.Provider
-	podWatcher      watch.Interface
 	resourceManager *manager.ResourceManager
+	podCh           chan *corev1.Pod
```
cpuguy83: Just a thought, wdyt about passing a wrapped object that includes the a context that we can trace through from the original request?

vkubelet/vkubelet.go
```diff
@@ -59,6 +57,7 @@ func New(ctx context.Context, cfg Config) (s *Server, retErr error) {
 		k8sClient:       cfg.Client,
 		resourceManager: cfg.ResourceManager,
 		provider:        cfg.Provider,
+		podCh:           make(chan *corev1.Pod, 1024),
```
cpuguy83: This seems like a pretty high number.
I'm wondering if the channel is really neccessary?

vkubelet/pod.go
```diff
@@ -21,6 +30,114 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
+		s.podCh <- pod
+	}
+}
+
+func (s *Server) startPodSynchronizer(ctx context.Context, id int) {
+	logger := log.G(ctx).WithField("method", "startPodSynchronizer").WithField("podSynchronizer", id)
+	logger.Debug("Start pod synchronizer")
+
+	for {
+		select {
```
cpuguy83: This can just be `for pod := range s.podCh`, which will break when the channel is closed.

vkubelet/vkubelet.go
cpuguy83: Is the worry here that VK can block up k8s processing?

cpuguy83: I wouldn't worry about this blocking VK because it will block eventually anyway, be it slow to queue or slow to process.

cmd/root.go
```diff
@@ -325,6 +328,28 @@ func initConfig() {
 			)
 		}
 	}
+
+	podSyncPoolSize = getPodSyncPoolSize(logger)
+}
+
+func getPodSyncPoolSize(logger *logrus.Entry) int {
+	poolSizeEnv := os.Getenv("POD_SYNC_POOL_SIZE")
```
cpuguy83: Why not a flag instead of an environment var?

cpuguy83: "--pod-sync-workers"

robbiezhang: I'm also struggling about using a flag vs. env var. Sounds flag makes more sense as it's a configuration to the VK itself. Will switch to it.

PR #375, opened 2018-10-13
[ACI] Update the network SDK and add more validations

robbiezhang:
1. Update the Azure network SDK to 2018-08-01.
2. Add more validations on the subnet to be delegated to ACI.
3. Update the network profile if it's not associated to the subnet.


cpuguy83:
:( Rate limited again...

PR #376, opened 2018-10-15
[VK] Change the default taint effect to NoSchedule

robbiezhang:
Regression bug which causes daemonset pods assign to virtual kubelet node. 
Change the default taint effect to NoSchedule.


cpuguy83:
This regression was caused by https://github.com/virtual-kubelet/virtual-kubelet/pull/360 where the default taint effect was inadvertently changed.

PR #378, opened 2018-10-18
[ACI] Support to add extra user agent to the request

robbiezhang:
User can set the extra user agent in the ACI_EXTRA_USER_AGENT environment.
If more than one extra user agent, use whitespace to separate them.

### DIFF
--- a/providers/azure/client/aci/client.go
+++ b/providers/azure/client/aci/client.go
@@ -13,9 +13,9 @@ import (
 
 const (
 	// BaseURI is the default URI used for compute services.
-	baseURI    = "https://management.azure.com"
-	userAgent  = "virtual-kubelet/azure-arm-aci/2018-09-01"
-	apiVersion = "2018-09-01"
+	baseURI          = "https://management.azure.com"
+	defaultUserAgent = "virtual-kubelet/azure-arm-aci/2018-09-01"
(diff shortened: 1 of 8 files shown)


cpuguy83:
Curious what is this for?

robbiezhang:
In general, ACI doesn't care how the VK is created, so, adding a special header to indicate the VK installation method in the ACI PUT API doesn't make much sense. So, I decide to append the installation method to the user agent.

providers/azure/client/aci/client.go
```diff
@@ -34,13 +34,13 @@ type Client struct {
...
-// NewClient creates a new Azure Container Instances client.
-func NewClient(auth *azure.Authentication) (*Client, error) {
+// NewClient creates a new Azure Container Instances client with extra user agent.
+func NewClient(auth *azure.Authentication, extraUserAgent string) (*Client, error) {
 	if auth == nil {
 		return nil, fmt.Errorf("Authentication is not supplied for the Azure client")
 	}
 
-	client, err := azure.NewClient(auth, baseURI, userAgent)
+	client, err := azure.NewClient(auth, baseURI, []string{defaultUserAgent, extraUserAgent})
```
cpuguy83: Actually we should only append `extrauserAgent` if it's not an empty string.

PR #379, opened 2018-10-18
[VK] Fix the potential runtime type casting error

robbiezhang:
We need to expect cache.DeletedFinalStateUnknown type in the OnDelete function. [reference](https://github.com/kubernetes/client-go/blob/80f980e507e36cabebdbedc628496c4d19c3af2c/tools/cache/controller.go#L173-L176)

### DIFF
--- a/vkubelet/pod.go
+++ b/vkubelet/pod.go
@@ -3,7 +3,6 @@ package vkubelet
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"time"
 
 	pkgerrors "github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


vkubelet/pod.go
```diff
@@ -30,11 +31,10 @@ func (s *Server) onAddPod(ctx context.Context, obj interface{}) {
...
 	logger := log.G(ctx).WithField("method", "onAddPod")
 
-	pod := obj.(*corev1.Pod)
-
-	if pod == nil {
-		span.SetStatus(trace.Status{Code: trace.StatusCodeInvalidArgument, Message: fmt.Sprintf("Unexpected object from event: %v", obj)})
-		logger.Errorf("obj is not a valid pod: %v", obj)
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		span.SetStatus(trace.Status{Code: trace.StatusCodeInvalidArgument, Message: fmt.Sprintf("Unexpected object from event: %s", reflect.TypeOf(obj))})
```
cpuguy83: `fmt.Sprintf("unexpected object from event: %T)`

PR #380, opened 2018-10-18
[ACI] Set the ACI_EXTRA_USER_AGENT in the helm charts

robbiezhang:


charts/virtual-kubelet/templates/deployment.yaml
```diff
@@ -60,13 +60,17 @@ spec:
           value: {{ .aciResourceGroup }}
         - name: ACI_REGION
           value: {{ .aciRegion }}
+        - name: ACI_EXTRA_USER_AGENT
+          value: {{ printf "helm-chart/aks/%s/%s" $.Chart.Name $.Chart.Version }}
```
cpuguy83: This doesn't seem right?

robbiezhang: anything wrong? I've verified the package locally.

cpuguy83: Why are we adding aks to the generic template?

robbiezhang: L45, it has the "targetAKS" flag == true

robbiezhang: we will deprecate the virtual-kubelet-for-aks helm chart soon.

PR #388, opened 2018-10-31
Fix loganalytics settings in chart's secret.yaml file

andrey-moor:
This fixes chart's `secret.yaml` template in case user configures Log Analytics workspace. 

The data in the secret should always be encoded.


andrey-moor:
I signed it

robbiezhang:
Good catch. Would you mind updating the package too?

andrey-moor:
@robbiezhang I updated the packages to `0.1.10` version? Is there anything else needed here?

PR #389, opened 2018-11-01
Add AWS/Fargate configuration options to chart.

headcr4sh:
While playing around with virtual-kubelet in my AWS environment, I noticed, that the helm chart lacks the necessary configuration options to enable AWS/Fargate support. This PR should add everything that is needed to create the requried .toml configuration file by using a ConfigMap.

### DIFF
--- a/charts/virtual-kubelet/values.yaml
+++ b/charts/virtual-kubelet/values.yaml
@@ -4,7 +4,7 @@ image:
   pullPolicy: Always
 
 ## `provider` should be one of aws, azure, azurebatch, etc...
-provider: 
+provider: ""
 nodeName: "virtual-kubelet"
 nodeOsType: "Linux"
 monitoredNamespace: ""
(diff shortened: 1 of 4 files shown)


headcr4sh:
I have just updated my PR and fixed an issue where the list of security groups was not correctly mapped into .toml-format.

(I also rebased my changes onto the current upstream master branch to make the review/merge easier)

charts/virtual-kubelet/values.yaml
```diff
@@ -42,7 +42,24 @@ providers:
       subnetCidr: 
       clusterCidr: 
       kubeDnsIp: 
-
+  aws:
+    targetFargate: true
```
robbiezhang: is there a scenario that aws is enabled but not target fargate?

headcr4sh: Mhh.. good point. I was trying to make the AWS/Fargate configuration similar to the Azure/AKS one, see `providers.azure.targetAKS` -- but I think I might have misunderstood the meaning of this flag.

cpuguy83:
Let's move charts to a different repo. I don't like that they are here (binary blobs bloat the repo a lot)... additionally we are working on splitting providers out of this repo and having each provider include it's own binary.

Thanks!

PR #390, opened 2018-11-01
Fix a case where provider pop status is not found

cpuguy83:
Updates the pod status in Kubernetes to "Failed" when the pod status is
not found from the provider.

Note that currently thet most providers return `nil, nil` when a pod is
not found. This works but should possibly return a typed error so we can
determine if the error means not found or something else... but this
works as is so I haven't changed it.


vkubelet/pod.go
```diff
@@ -217,24 +221,65 @@ func (s *Server) updatePodStatuses(ctx context.Context) {
...
+		return pkgerrors.Wrap(err, "error retreiving pod status")
+	}
+
+	// Update the pod's status
+	if status != nil {
+		pod.Status = *status
+	} else {
+		// Only change the status when the pod was already up
+		// Only doing so when the pod was successfully running makes sure we don't run into race conditions during pod creation.
+		if pod.Status.Phase == corev1.PodRunning {
```
robbiezhang: pod in Pending state might also hit this issue right? since the Succeeded and Failed phases are handled in L236, shall we remove this check?

vkubelet/vkubelet.go
```diff
@@ -161,8 +161,8 @@ func (s *Server) reconcile(ctx context.Context) {
 
 	var failedDeleteCount int64
 	for _, pod := range deletePods {
-		logger := logger.WithField("pod", pod.Name)
-		logger.Debug("Deleting pod '%s'\n", pod.Name)
```
robbiezhang: just about to fix the issue. can you also add the namespace?

vkubelet/pod.go
cpuguy83: Not putting this check was *really* problematic.

cpuguy83: Maybe I can check "Reason" and make sure it's not in "Creating".

vkubelet/vkubelet.go
cpuguy83: Fixed.

vkubelet/pod.go
robbiezhang: The reason "Creating" is only used in ACI. Add this check might break the other providers.

robbiezhang: Can we check the pod's creationTimestamp and set a timeout threshold?

cpuguy83: Can you explain what you mean here?

robbiezhang: if pod.Status.Phase == corev1.PodRunning || pod.ObjectMeta.CreationTimestamp.Add(time.Minute) < time.Now() {
...
}

cpuguy83: 👍 

Updated

PR #391, opened 2018-11-01
[VK] Add a reconcile thread to clear the stale pods in provider

robbiezhang:
After we changed to the concurrent mode, there is a chance of race condition that a pod is created and deleted in parallel when the pod is deleted immediately after it is created.
This change adds a reconcile thread to clear the stale pods in provider.


vkubelet/vkubelet.go
```diff
@@ -136,11 +144,11 @@ func (s *Server) Stop() {
 
 // reconcile is the main reconciliation loop that compares differences between Kubernetes and
 // the active provider and reconciles the differences.
-func (s *Server) reconcile(ctx context.Context) {
+func (s *Server) reconcile(ctx context.Context, clearStaleProviderPodsOnly bool) {
```
cpuguy83: Would it make sense to break out stale pod reconciliation into a separate function?

vkubelet/pod.go
```diff
@@ -280,21 +284,28 @@ func (s *Server) updatePodStatus(ctx context.Context, pod *corev1.Pod) error {
 	} else {
 		// Only change the status when the pod was already up
 		// Only doing so when the pod was successfully running makes sure we don't run into race conditions during pod creation.
-		if pod.Status.Phase == corev1.PodRunning || pod.ObjectMeta.CreationTimestamp.Add(time.Minute).Before(time.Now()) {
+		if pod.Status.Phase == corev1.PodRunning || pod.CreationTimestamp.Add(time.Minute).Before(time.Now()) {
+			logger.Error("Pod doesn't exist in the provider, and set pod phase to failed.")
```
cpuguy83: Is this an error?

cpuguy83: Maybe info, or warn even?

cpuguy83:
Testing this out, for some reason I still have a stale pod even though stale pod reconciliation has run multiple times.

4 containers in ACI, 3 pods in k8s.

cpuguy83:
Even a full reconcile run is missing it.

robbiezhang:
The fix breaks multiple VK (Linux, Windows) running at the same time in the AKS.

PR #392, opened 2018-11-02
Fixed typo on helm install sample

cmendible:
Fixed typo on helm install sample. Replaced providers.azure.clientID with provider.azure.clientId so the values are correctly parsed.


rbitia:
Thanks for the edit!

PR #393, opened 2018-11-02
Serialize event handling per pod

cpuguy83:
Before this change all pod events are handling concurrently. This means
that multiple events for the same pod can be handled at the same time
and end up in a race where they may be processed in the incorrect order.

An easy way to reproduce this issue is to fire up VK, add some pods, then suspend it
(ctrl+z) for a minute or two, then resume. A blast of events will come
from k8s and cause out of order handling. This will cause duplicate pods
to be created in the provider that can never be cleaned up (by VK).

This change uses a mutex keyed on pod namespace+name to ensure that only
one event is handled at a time and as an added check to ensure that the
pod does not already exist in the resource manager before triggering a
pod create.

### DIFF
--- a/vkubelet/pod.go
+++ b/vkubelet/pod.go
@@ -105,7 +105,21 @@ func (s *Server) startPodSynchronizer(ctx context.Context, id int) {
 	logger.Debug("Start pod synchronizer")
 
 	for event := range s.podCh {
-		s.syncPod(event.ctx, event.pod)
+		logger := log.G(event.ctx).WithField("podSyncronizer", id)
+		ctx := log.WithLogger(event.ctx, logger)
+
+		lockCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
+		if err := s.lock(lockCtx, event.pod); err != nil {
+			cancel()
+			if pkgerrors.Cause(err) != context.Canceled {
+				logger.WithError(err).Error("error acquiring lock for pod")
+			}
+			continue
+		}
+		cancel()
+
+		s.syncPod(ctx, event.pod)
+		s.unlock(event.pod)
 	}
 
 	logger.Info("pod channel is closed.")
@@ -114,6 +128,7 @@ func (s *Server) startPodSynchronizer(ctx context.Context, id int) {
 func (s *Server) syncPod(ctx context.Context, pod *corev1.Pod) {
 	ctx, span := trace.StartSpan(ctx, "syncPod")
 	defer span.End()
+
 	logger := log.G(ctx).WithField("pod", pod.GetName()).WithField("namespace", pod.GetNamespace())
 
 	addPodAttributes(span, pod)
@@ -281,7 +296,7 @@ func (s *Server) watchForPodEvent(ctx context.Context) error {
 		time.Minute,
 
 		cache.ResourceEventHandlerFuncs{
-			AddFunc:    func(obj interface{}) {
+			AddFunc: func(obj interface{}) {
 				s.onAddPod(ctx, obj)
 			},
 			UpdateFunc: func(oldObj, newObj interface{}) {
--- a/vkubelet/sync.go
+++ b/vkubelet/sync.go
@@ -1,0 +1,144 @@
+package vkubelet
+
+import (
+	"context"
+	"errors"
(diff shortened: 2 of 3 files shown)


vkubelet/pod.go
```diff
@@ -105,7 +105,21 @@ func (s *Server) startPodSynchronizer(ctx context.Context, id int) {
 	logger.Debug("Start pod synchronizer")
 
 	for event := range s.podCh {
-		s.syncPod(event.ctx, event.pod)
+		log.G(ctx).WithField("podSyncronizer", id)
```
robbiezhang: Should it be used in the next line?

vkubelet/sync.go
```diff
@@ -0,0 +1,144 @@
...
+		panic("unlock of unlocked mutex")
+	}
+}
+
+// ErrNoSuchLock is returned when the requested lock does not exist
+var ErrNoSuchLock = errors.New("no such lock")
+
+// NamedLock provides a locking mechanism based on the passed in reference name
+type NamedLock struct {
+	mu    sync.Mutex
```
robbiezhang: use *sync.Mutex or derive from sync.Mutex

cpuguy83: This is fine here because we are operating on a a pointer for `NamedLock`, and generally how snyc.Mutex would be initialized in situations like this.

vkubelet/pod.go
cpuguy83: ah darn... also should be event.ctx here.
Fixed.

vkubelet/sync.go
```diff
@@ -0,0 +1,144 @@
...
+
+func NewCancellableMutex() *CancellableMutex {
+	return &CancellableMutex{
+		ch: make(chan struct{}, 1),
+	}
+}
+
+// CancellableMutex is a mutex in which the Lock() action can be canncelled by
+// cancelling a context.
+// This must be initialized iby calling `NwwCancellableMutex()`
```
robbiezhang: typos: iby, Nww

vkubelet/pod.go
```diff
@@ -125,6 +140,10 @@ func (s *Server) syncPod(ctx context.Context, pod *corev1.Pod) {
 			logger.WithError(err).Error("Failed to delete pod")
 		}
 	} else {
+		if stored := s.resourceManager.GetPod(pod.GetName(), pod.GetNamespace()); stored != nil {
```
robbiezhang: this check is not correct, as we add the pod into the resource manager before adding it into the podSync channel.

vkubelet/sync.go
```diff
@@ -0,0 +1,144 @@
...
+//
+// You should expect this to be quite a bit slower than a normal sync.Mutex from
+//n the stdlib.
```
robbiezhang: typo: n

vkubelet/pod.go
cpuguy83: oh my this is so wrong. It's not preventing things from working because the call is incorrect!

cpuguy83:
This seems to resolve the issues I've been seeing, however...

It's still racey in that a `Delete` could be processed before a `Create` just based on thread scheduling... just that on my system it happens to be performing these in the correct order in all of my testing.

vkubelet/pod.go
```diff
@@ -105,7 +105,21 @@ func (s *Server) startPodSynchronizer(ctx context.Context, id int) {
 	logger.Debug("Start pod synchronizer")
 
 	for event := range s.podCh {
-		s.syncPod(event.ctx, event.pod)
+		logger := log.G(event.ctx).WithField("podSyncronizer", id)
+		ctx := log.WithLogger(event.ctx, logger)
+
+		lockCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
+		if err := s.lock(lockCtx, event.pod); err != nil {
```
robbiezhang: if we have multiple thread, lock cannot guarantee the event order. With the timeout, we might lost event if the operation accumulates.

cpuguy83: Yep, agreed. This does seem to make everything safer, but ultimately it's not a perfect solution.

PR #394, opened 2018-11-02
Rollback to use only 1 pod-sync worker in charts.

cpuguy83:
This overrides the default value from VK. The reason for this change is
to work-around race conditions caused by concurrent pod syncronization.
This is a temporary change until the underlying concurrency issues are
resolved.


cpuguy83:
Closed in favor of #395

PR #398, opened 2018-11-02
[VK] BugFix for Deleting Pod

robbiezhang:
1. Fix the wrong error handling in deletePod. It should return error when delErr is **NOT** NotFound.
2. If delete failed, we don't retry. It's a potential cause of the pod leaking.


robbiezhang:
Fix #400

cpuguy83:
This looks pretty good, I'll run this through some tests.

vkubelet/pod.go
```diff
@@ -179,30 +183,28 @@ func (s *Server) deletePod(ctx context.Context, pod *corev1.Pod) error {
 	addPodAttributes(span, pod)
 
 	var delErr error
-	if delErr = s.provider.DeletePod(ctx, pod); delErr != nil && errors.IsNotFound(delErr) {
+	if delErr = s.provider.DeletePod(ctx, pod); delErr != nil && !errors.IsNotFound(delErr) {
```
cpuguy83: I don't think providers are returning a K8s error type? I don't think this is working as intended with or without the change.

We need providers to convert to errors that we can understand here.

cpuguy83: We could use something like `strongerrors.NotFound(err)` in the provider and `strongerrors.IsNotFound(err)` here.

Would be nice to make sure all providers do this, but can fallback to `GetPod` as well.

robbiezhang: Sounds good

cpuguy83: https://github.com/virtual-kubelet/virtual-kubelet/pull/403 adds this to providers.

cpuguy83: btw I think we want this to return when it is a NotFound error because there is nothing to do.

cpuguy83: Or more specifically, we want to requeue with it's not a NotFound error.

robbiezhang: we need to hard delete the pod from k8s

robbiezhang:
thanks to #425 , this is no longer needed.

PR #403, opened 2018-11-06
Use strongerrors in provider implementations.

cpuguy83:
This ensures that we can catch certain types of errors from providers
and handle accordingly in the core. There is still more to do here to
improve that but this resolves an immediate need to know why a Delete
failed.

vic provider was not updated since I could not figure out where to get
this information.


robbiezhang:
@cpuguy83 Nice job. I'll take a look.

providers/azure/client/api/api.go
```diff
@@ -56,12 +61,20 @@ func CheckResponse(res *http.Response) error {
 			}
 			jerr.Error.Body = string(slurp)
 			jerr.Error.URL = res.Request.URL.String()
-			return jerr.Error
+			e = jerr.Error
+		}
+	} else {
+		e = &Error{
+			StatusCode: res.StatusCode,
+			Body:       string(slurp),
```
robbiezhang: old issue: slurp is nill here. Shall we use res.Status instead?

providers/azure/client/api/api.go
```diff
@@ -56,12 +61,20 @@ func CheckResponse(res *http.Response) error {
...
+			Header:     res.Header,
 		}
 	}
-	return &Error{
-		StatusCode: res.StatusCode,
-		Body:       string(slurp),
-		Header:     res.Header,
+
+	switch e.StatusCode {
```
robbiezhang: wondering whether we should implement it in the aci provider like what you did for the other providers.

cpuguy83: Yeah, I definitely cheated here.

providers/azure/client/api/api.go
cpuguy83: Fixed

providers/azure/client/api/api.go
cpuguy83: Moved this handling into the provider.

PR #405, opened 2018-11-09
[WIP] Use shared informer and informer cache

cpuguy83:
This refactors everything to rely on the inforer cache and a work queue
provided with the k8s client.

The work queue ensures that only one operation happens at a time on a
given key, supports requeing with tracking retries, and backoff on
retries.

This mostly removes the need for the resource manager, however it's
still used in some places (e.g. to get secrets) that should be
refactored separately.

This change fixes concurrency issues when operating on a pod by ensuring
that only one operation occurs at a time on a pod and ensures the worker
must get the most up to date version of a pod.


jeremyrickard:
Nice!

cmd/root.go
```diff
@@ -61,6 +67,8 @@ var p providers.Provider
 var rm *manager.ResourceManager
 var apiConfig vkubelet.APIConfig
 var podSyncWorkers int
+var listWatcher *cache.ListWatch
+var watchSyncInterval = time.Minute
```
pires: Is there the need for such an aggressive interval? I would even argue it should be entirely disabled. See https://groups.google.com/forum/#!topic/kubernetes-sig-api-machinery/PbSCXdLDno0.

cmd/root.go
```diff
@@ -76,6 +84,16 @@ This allows users to schedule kubernetes workloads on nodes that aren't running
 	Run: func(cmd *cobra.Command, args []string) {
 		ctx, cancel := context.WithCancel(context.Background())
 
+		sig := make(chan os.Signal, 1)
+		signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM)
```
pires: IIRC this doesn't work on Windows. Should use `os.Interrupt` instead. See https://stackoverflow.com/a/35683558

vkubelet/vkubelet.go
```diff
@@ -117,12 +139,18 @@ func New(ctx context.Context, cfg Config) (s *Server, retErr error) {
...
 	}()
 
-	return s, nil
-}
+	go func() {
+		tick := time.Tick(time.Minute)
+		for {
+			select {
+			case <-ctx.Done():
+				return
```
pires: Can you please confirm the program exists when this goroutine returns?

Sorry if I'm wasting your time, but I've been bitten before by `time.Tick` leaking the underlying ticker even if a goroutine ends. If the program exits when this goroutine exits, it's fine.

vkubelet/vkubelet.go
```diff
@@ -133,15 +161,16 @@ func (s *Server) Run(ctx context.Context) error {
 	return nil
 }
 
-// reconcile is the main reconciliation loop that compares differences between Kubernetes and
-// the active provider and reconciles the differences.
+// reconcile ensures that pods that exist in the provdier which don't exist in
```
pires: typo: _provdier_

cmd/root.go
pires: I just realized I may have assumed wrongly that it's intended for the VK to also run on Windows.

cmd/root.go
cpuguy83: Mostly just copying what was already there. Thinking about it, I agree we probably don't need to  sync so aggressively.

cmd/root.go
cpuguy83: Yep, will fix.

vkubelet/vkubelet.go
cpuguy83: You are right, we shouldn't assume the application is exiting in this case. There are some other tickers to update as well.

PR #410, opened 2018-11-14
Replace dep validation with dep check and resyncing Gopkg.lock

tariq1890:
Reasons for adding this

i) `dep check` is a lightweight alternative to the current dep validation in virtual-kubelet
ii) `dep check` provides deeper validation than the current method. For eg:- It also catches instances where new imports from an existing lib is added to the code base and the Gopkg.lock has not been updated


tariq1890:
@cpuguy83 FYI

PR #415, opened 2018-11-21
hypersh: fix pod leaking

pires:
While doing some testing, I realized the Hyper.sh provider mishandles its container names, causing pods to leak. This is a fix.

@resouer I think you're no longer working for Hyper.sh but you're still documented as owner for this provider. Do let me know if this has changed and who we should reach.

### DIFF
--- a/providers/hypersh/hypersh.go
+++ b/providers/hypersh/hypersh.go
@@ -219,7 +219,7 @@ func (p *HyperProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 	// Iterate over the containers to create and start them.
 	for k, ctr := range containers {
 		//one container in a Pod in hyper.sh currently
-		containerName := fmt.Sprintf("pod-%s-%s", pod.Name, pod.Spec.Containers[k].Name)
+		containerName := containerName(pod.Namespace, pod.Name)
 
 		if err = p.ensureImage(ctr.Image); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


providers/hypersh/hypersh.go
```diff
@@ -304,7 +304,7 @@ func (p *HyperProvider) DeletePod(ctx context.Context, pod *v1.Pod) (err error)
 // returns nil if a pod by that name is not found.
 func (p *HyperProvider) GetPod(ctx context.Context, namespace, name string) (pod *v1.Pod, err error) {
 	var (
-		containerName = fmt.Sprintf("pod-%s-%s", name, name)
+		containerName = containerName(pod.Namespace, pod.Name)
```
robbiezhang: pod is the output parameter. do you mean namespace, name?

providers/hypersh/hypersh.go
```diff
@@ -219,7 +219,7 @@ func (p *HyperProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 	// Iterate over the containers to create and start them.
 	for k, ctr := range containers {
 		//one container in a Pod in hyper.sh currently
-		containerName := fmt.Sprintf("pod-%s-%s", pod.Name, pod.Spec.Containers[k].Name)
+		containerName := containerName(pod.Namespace, pod.Name)
```
robbiezhang: don't quite understand the logic here. do all the containers in a pod use the same container name in hyper?

pires:
Thank you @robbiezhang for taking the time to look into this.

I definitely underestimated the complexity of the issue here:

- This provider handles multiple containers during `CreatePod`, but *not* during `DeletePod` or `GetPod`, where only one container is to be handled (???).
- Comment in L221 even mentions `one container in a Pod in hyper.sh currently`.
- In `CreatePod`, each container computed name is `<podspec.name>-<containerspec.name>`, while in `DeletePod` and `GetPod`, the computed container name is `<podspec.name>-<podspec.name>`.
- While normalizing the computation of one container's name here, indeed I am introducing another bug in `CreatePod` where all containers will have the same name.

It seems to me now that this provider is in dire need for love. Love that I'm not willing to give it now. Therefore, I will close this PR.

PR #424, opened 2018-11-28
Fix pod deletion

pires:
The introduction of `skaffold` - and ultimately the unfortunate fact that a permission was missing from the virtual-kubelet cluster role binding - allowed to uncover a bug in `pod.go`: when trying to delete a Pod resource from the Kubernetes API, any errors except for 404 would be swallowed (i.e. not propagated, and no log messages would be printed).

### DIFF
--- a/hack/skaffold/virtual-kubelet/base.yml
+++ b/hack/skaffold/virtual-kubelet/base.yml
@@ -12,9 +12,18 @@ rules:
   - ""
   resources:
   - configmaps
-  - pods
   - secrets
   verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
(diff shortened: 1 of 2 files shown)


robbiezhang:
The helm chart uses "cluster-admin" role
https://github.com/virtual-kubelet/virtual-kubelet/blob/338a7c2213f897a452eadbcaee840776d5b64779/charts/virtual-kubelet/values.yaml#L53
Shall we make it consistent?

pires:
@robbiezhang I'm not familiar with Helm so I leave up to you to decide and execute, if you decide to do so.

pires:
`test` check is [failing due to issues authenticating with Azure](https://circleci.com/gh/virtual-kubelet/virtual-kubelet/1301)

Issue #404, opened 2018-11-09
consolidate logging and tracing


cpuguy83:
Logging and tracing log a lot of the same information, it would be nice if the handling of these are consolidated as much as possible.

Issue #348, opened 2018-09-14
Refactor Go API interfaces


cpuguy83:
In light of https://github.com/virtual-kubelet/virtual-kubelet/issues/160 (making VK a library)

I wanted to propose some changes to the (go) API surface:

1. Make a clean split between services provided by virtual-kubelet
i. Examples of the mentioned services: a daemon, a control loop, HTTP API routers for exec, logs, and metrics
ii. e.g., the vkubelet package should not be automatically starting up an HTTP server but instead rely on the implementation (e.g cmd/) to handle this.
iii. Do not depend on environment variables except in the actual CLI, and pass down the values of these as arguments to functions that need them 
2. Define stronger error types
i. Components should be able to determine what kinds of errors have been returned
ii. Think along the lines of GRPC error codes
iii. e.g., a Provider should be able to return an error that is a "not found' type of error and the HTTP router should be able to understand it.
iv. I have an example of this in https://github.com/cpuguy83/strongerrors, which is essentially what is used in moby (see https://godoc.org/github.com/moby/moby/errdefs)
3. Setup tracing/metrics which are propagated to all components and any potential consumers of the Go API.
i. I've been playing with OpenCencus for this as it is trivial to write exporters for whatever system the consumer wants to use (rather than enforcing a common library that must be used for any consumer).... but I'm not necessarily tied to it

Most of this is just improving the code hygiene and having a cleaner separation of responsibilities.
Right now most of this code base is assuming it's just a binary that's being run in a particular way.
The idea here is to have a framework for building a kubelet implementation, and `cmd/virtual-kubelet` would continue to provide that implementation.

All in all, this should not introduce any breaking changes to the daemon, installation methods, or anything of that nature.
This repo can still provide a daemon with all the available providers packaged up as it does today, but in addition it will make it easier to develop a custom VK daemon for providers (whether or not they exist in the repo today or not) to do their own thing.
Actually making any sort of breaking change with regard to the distribution of a daemon is a wholly separate topic.

cpuguy83:
ping @jeremyrickard @rbitia @robbiezhang

pires:
@cpuguy83 what is holding this back?

cpuguy83:
Just a status update on this.
Still working on re-doing some of the provider interfaces

- #578 (exec)
- #574 (log streaming)

After these, the other big thing is defining a strong error interface between providers and the core.
The split between the core framework and the built-in CLI/daemon implementation is complete.
[...]

PR #425, opened 2018-11-28
use shared informers and workqueue

pires:
This PR reworks how VK implements its Kubernetes controller:

- Replaces the current manual mechanism of event watching and processing with using:
  - two shared informer factories (one for `Pod`, and one for `ConfigMap` and `Secret`), and
  - a work queue.
- The pod shared informer factory is constrained via `WithTweakListOpts` so that it only produces events for pods that have been assigned to the current node.
- The work queue bits were mostly drawn from from https://github.com/kubernetes/sample-controller, with a minor adjustment so we can stop processing a pod after its reconciliation has failed _N_ times.
- `ResourceManager` is no longer needed as informers' listers can be used to read Kubernetes resources, but since the `ResourceManager` is used in the ACI provider test, I decided not to remove it at this point. We can think about it as we progress.

Closes #405


vkubelet/podcontroller.go
```diff
@@ -0,0 +1,339 @@
...
+	})
+
+	// Return the instance of PodController back to the caller.
+	return pc
+}
+
+// Run will set up the event handlers for types we are interested in, as well as syncing informer caches and starting workers.
+// It will block until stopCh is closed, at which point it will shutdown the work queue and wait for workers to finish processing their current work items.
+func (pc *PodController) Run(ctx context.Context, threadiness int) error {
+	defer runtime.HandleCrash()
```
cpuguy83: Not sure I like this. We should just let panics bubble up, also seems like a todo item to remove this upstream.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,339 @@
...
+
+const (
+	// maxRetries is the number of times we try to process a given key before permanently forgetting it.
+	maxRetries = 20
+)
+
+// PodController is the controller implementation for Pod resources.
+type PodController struct {
+	// server is the instance to which this controller belongs.
+	server *Server
```
cpuguy83: This seems a little off (i.e. something is not abstracted well).
We should revisit this.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,339 @@
...
+
+// runWorker is a long-running function that will continually call the processNextWorkItem function in order to read and process an item on the work queue.
+func (pc *PodController) runWorker() {
+	for pc.processNextWorkItem() {
+	}
+}
+
+// processNextWorkItem will read a single work item off the work queue and attempt to process it,by calling the syncHandler.
+func (pc *PodController) processNextWorkItem() bool {
+	obj, shutdown := pc.workqueue.Get()
```
cpuguy83: Some tracing in this function (and recursively) would be nice.

pires: I have very little experience with tracing. Adding #404, I thought maybe we could postpone this to a later phase (pre-1.0).

Also, I am trying to keep the size of this PR as little as possible so it's easier to review.

Does this convince you? If not, I'll try and see what I can do as part of this PR.

pires: I can do this as part of my next task, work on #348. WDYT?

vkubelet/podcontroller.go
pires: I was thinking about doing this as part of my next task, work on #348, and keep this PR focused on the controller side of things. WDYT?

vkubelet/podcontroller.go
pires: I'll look deeper into this.

cmd/root.go
```diff
@@ -251,11 +266,30 @@ func initConfig() {
...
+	configMapInformer := scmInformerFactory.Core().V1().ConfigMaps()
+
+	// Create a new instance of the resource manager that uses the listers above for pods, secrets and config maps.
+	rm, err = manager.NewResourceManager(podInformer.Lister(), secretInformer.Lister(), configMapInformer.Lister())
 	if err != nil {
 		logger.WithError(err).Fatal("Error initializing resource manager")
 	}
 
+	// Start the shared informer factory for pods.
+	go podInformerFactory.Start(context.Background().Done())
```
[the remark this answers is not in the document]
pires: Done in https://github.com/virtual-kubelet/virtual-kubelet/pull/425/commits/e8218079772f9ec1be1854c7dfb83b1c3ac34de8.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,339 @@
...
+	// Convert the namespace/name string into a distinct namespace and name.
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		// Log the error but do not requeue the key as it is invalid.
+		runtime.HandleError(vkerrors.InvalidArgument(err))
+		return nil
+	}
+
+	// Create a context to pass to the provider.
+	ctx := context.Background()
```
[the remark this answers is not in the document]
pires: Done in https://github.com/virtual-kubelet/virtual-kubelet/pull/425/commits/e8218079772f9ec1be1854c7dfb83b1c3ac34de8.

cmd/root.go
```diff
@@ -251,11 +266,30 @@ func initConfig() {
...
-	rm, err = manager.NewResourceManager(k8sClient)
+	// Create a shared informer factory for Kubernetes pods in the current namespace (if specified) and scheduled to the current node.
+	podInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(k8sClient, kubeSharedInformerFactoryResync, kubeinformers.WithNamespace(kubeNamespace), kubeinformers.WithTweakListOptions(func(options *metav1.ListOptions) {
+		options.FieldSelector = fields.OneTermEqualSelector("spec.nodeName", nodeName).String()
+	}))
+	// Create a pod informer so we can pass its lister to the resource manager.
+	podInformer = podInformerFactory.Core().V1().Pods()
+
+	// Create another shared informer factory for Kubernetes secrets and configmaps (not subject to any selectors).
+	scmInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(k8sClient, kubeSharedInformerFactoryResync)
```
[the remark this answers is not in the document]
pires: That's because listers obtained from the `podInformerFactory` are all subject to what is specified in `WithTweakListOptions`. Have in mind that [the method you are using in your PR is deprecated](https://github.com/kubernetes/client-go/blob/master/informers/factory.go#L99-L100) in favor of what I'm adopting here.

If we create a secret/configmap informer from `podInformerFactory` - so that we could grab its lister - we would see error messages indicating that `spec.nodeName` is not a valid selector for secrets/configmaps. On the other hand, we must constrain the `podInformerFactory` with said selector, since we are only interested in pods (and in particular pod events) that have been scheduled to the current node. Hence, we need two separate factories.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,339 @@
...
+		podsLister:   server.podInformer.Lister(),
+		workqueue:    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "pods"),
+		recorder:     recorder,
+	}
+
+	// Set up event handlers for when Pod resources change.
+	pc.podsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(pod interface{}) {
+			if key, err := cache.MetaNamespaceKeyFunc(pod); err != nil {
+				runtime.HandleError(err)
```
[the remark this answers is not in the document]
pires: It's doing slightly more than that: [it's preventing error logs from flooding the output](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/runtime/runtime.go#L91-L98).

Unfortunately, it is not possible to replace the logging facility that `HandleError` uses.
We could roll our own handler in order to use the VK's logger. as you suggest, but the problem is that we would have to implement/steal backoff ourselves, as [`rudimentaryErrorBackoff` is not exported](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/runtime/runtime.go#L121-L140).

We can always forget backoff and just log away - but is that something we really want to do? I'd say no. WDYT?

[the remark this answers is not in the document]
cpuguy83: I'm not as worried about logging backoff at the moment.
Though we can easily copy/paste that function here, it doesn't seem al that necessary.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,339 @@
...
+		podsInformer: server.podInformer,
+		podsLister:   server.podInformer.Lister(),
+		workqueue:    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "pods"),
+		recorder:     recorder,
+	}
+
+	// Set up event handlers for when Pod resources change.
+	pc.podsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(pod interface{}) {
+			if key, err := cache.MetaNamespaceKeyFunc(pod); err != nil {
```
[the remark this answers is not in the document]
cpuguy83: Can you rename `metaKey` to make it clear the intent is to use it for logging?

errors/errors.go
```diff
@@ -0,0 +1,49 @@
+// Copyright © 2017 The virtual-kubelet authors
```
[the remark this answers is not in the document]
cpuguy83: We can use strongerrors, but I just don't think this package is adding much.

vkubelet/podcontroller.go
cpuguy83: SGTM

vkubelet/podcontroller.go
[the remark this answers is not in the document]
pires: Yes, sir!

cmd/root.go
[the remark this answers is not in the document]
cpuguy83: The deprecation only seems to apply with the method without options, but it doesn't seem like it's deprecated to create an informer from another?
The shared informer factory seems to lose value if you can't have listers with different list options applied coming from the same informer.

vkubelet/podcontroller.go
cpuguy83: It's pretty easy to add. I'd like to get it in here if we can.

You create a span (`trace.StartSpan(ctx, <a span name>)`), the span should end at the end of the function (typically, and should be fine here by calling `defer span.End()`)
The span context is passed through a `context.Context`.
There should already be a helper to add pod attributes to the span.

If the function ends in an error you can set a span status `span.SetStatus(ocstatus.FromError(err))`.

errors/errors.go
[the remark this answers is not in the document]
pires: OK I'm going to revert the commit.

vkubelet/podcontroller.go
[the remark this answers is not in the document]
pires: Will forget backoff.

vkubelet/podcontroller.go
pires: I'll look into it tomorrow morning my time.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,322 @@
...
+
+	// Perform a reconciliation step that deletes any dangling pods from the provider.
+	// This happens only when the virtual-kubelet is starting, and operates on a "best-effort" basis.
+	// If by any reason the provider fails to delete a dangling pod, it will stay in the provider and deletion won't be retried.
+	pc.deleteDanglingPods()
+
+	// Launch two workers to process Pod resources.
+	log.G(pc.context).Info("starting workers")
+	for i := 0; i < threadiness; i++ {
+		go wait.Until(pc.runWorker, time.Second, pc.context.Done())
```
robbiezhang: It will still cause pod leaking if threadiness > 1!

If the user create a pod and delete it immediately, you will receive at least 2 events: ADD pod, DELETE pod. Both of them are en-queued. If there are 2 workers, they will probably be run in different thread. There is a chance that DELETE event is handled before ADD event, and cause the leaking of the pod in the provider.

manager/resource.go
```diff
@@ -2,321 +2,50 @@ package manager
...
-	rm.RLock()
-	defer rm.RUnlock()
-
-	pods := make([]*v1.Pod, 0, len(rm.pods)+len(rm.deletingPods))
-	for _, p := range rm.pods {
-		pods = append(pods, p)
-	}
-	for _, p := range rm.deletingPods {
-		pods = append(pods, p)
+	l, err := rm.podLister.Pods(v1.NamespaceAll).List(labels.Everything())
```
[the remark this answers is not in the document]
pires: Nice catch! Old habits die hard and this one comes with me from working on other operators.

Fixed in eb92f52.

manager/resource.go
```diff
@@ -2,321 +2,50 @@ package manager
...
+	if err == nil {
+		return l
 	}
-
-	return pods
+	log.L.Errorf("failed to fetch pods from lister: %v", err)
```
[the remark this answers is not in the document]
pires: How the VK passes context is confusing to me.

The `ResourceManager` does have no context specified in its methods, when it should, so you're asking me to refactor that and change all its clients (VK, providers, tests) that use it? If that's the case, I'd ask for it to be done as part of #348.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,322 @@
...
+	context context.Context
+}
+
+// NewPodController returns a new instance of PodController.
+func NewPodController(server *Server, context context.Context) *PodController {
+	// Create an event broadcaster.
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(log.L.Infof)
+	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: server.k8sClient.CoreV1().Events("")})
+	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "pod-controller"})
```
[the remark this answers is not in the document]
pires: In order to make this collision-proof, I've settled with `<node name>/podcontroller`. Later on, we may have multiple controllers in the same VK that we want to record events for and even multiple VKs in the same cluster.

Done in 828ad18.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,322 @@
...
+	if ok := cache.WaitForCacheSync(pc.context.Done(), pc.podsInformer.Informer().HasSynced); !ok {
+		return pkgerrors.New("failed to wait for caches to sync")
+	}
+
+	// Perform a reconciliation step that deletes any dangling pods from the provider.
+	// This happens only when the virtual-kubelet is starting, and operates on a "best-effort" basis.
+	// If by any reason the provider fails to delete a dangling pod, it will stay in the provider and deletion won't be retried.
+	pc.deleteDanglingPods()
+
+	// Launch two workers to process Pod resources.
```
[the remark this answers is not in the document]
pires: Right, `threadiness`!

Fixed in e536b87.

vkubelet/podcontroller.go
pires: Actually, it won't.

To start with, we don't enqueue events anymore. We enqueue (pod) keys.

Workers popping from the queue are not aware of the type of event anymore. They'll always look into the state of both worlds (Kubernetes and provider) before making any decisions.

The work queue also offers us the guarantees we need. It guarantees, for example, that items are processed in the order in which they are queued. Furthermore, while a (pod) key is being processed by a worker, no other worker can process that same key - it is remembered by the work queue as being in "processing" state. That being, what would happen in the scenario you describe is roughly the following:

1. `ADDED` event causes `foo` to be added to the work queue.
2. `MODIFIED`/`DELETED` event causes `foo` to be added to the work queue.
3. Worker 1 pops `foo` to process. Processing takes N time. Worker 2 cannot pick `foo` to process in the meantime.
4. Worker 1 finishes processing. Since the key has been re-added in the meantime, it can now be picked by either Worker 1 or Worker 2. The process repeats from 3.

Alternatively, this can happen:

1. `ADDED` event causes `foo` to be added to the work queue.
2. Worker 1 immediately pops `foo` to process. Processing takes N time.
3. Roughly at the same time, `MODIFIED`/`DELETED` event causes `foo` to be added to the work queue. One of two things happen:
   1. `foo` is still being processed by Worker 1, in which case it is not added to the queue.
   2. `foo` has already been processed by Worker 1, in which case it is re-added and the process repeats itself.

For further information about the guarantees provided by the work queue, you can have a look at https://github.com/kubernetes/client-go/blob/master/util/workqueue/doc.go#L17-L25.

vkubelet/podcontroller.go
pires: Done in 1a4b5b0 and 5ea3b72.

vk_create_pod_trace
vk_delete_pod_trace

vkubelet/podcontroller.go
pires: [I've added tracing](https://github.com/virtual-kubelet/virtual-kubelet/pull/425#discussion_r237898598) so you can try and reproduce this behavior on your side, if you want to.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,380 @@
...
+			err := pkgerrors.Wrapf(err, "failed to delete pod %q in the provider", loggablePodName(pod))
+			span.SetStatus(ocstatus.FromError(err))
+			return err
+		}
+		return nil
+	}
+
+	// Ignore the pod if it is in the "Failed" state.
+	if pod.Status.Phase == corev1.PodFailed {
+		log.G(ctx).Warnf("skipping sync of pod %q in %q phase", loggablePodName(pod), pod.Status.Phase)
```
cpuguy83: Should this be debug?

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(len(ptd))
+
+	// Iterate over the slice of pods to be deleted and delete them in the provider.
+	for _, pod := range ptd {
+		go func(ctx context.Context, pod *corev1.Pod) {
+			// Create a span for the current call.
```
cpuguy83: Most of these comments seem superfluous.

cpuguy83: Usually for code comments I keep it to why something is being done when it's not obvious from the code.
But here the code comment is saying exactly what the code is saying.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,380 @@
...
+	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: fmt.Sprintf("%s/pod-controller", server.nodeName)})
+
+	// Create an instance of PodController having a work queue that uses the rate limiter created above.
+	pc := &PodController{
+		server:       server,
+		podsInformer: server.podInformer,
+		podsLister:   server.podInformer.Lister(),
+		workqueue:    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "pods"),
+		recorder:     recorder,
+		context:      context,
```
cpuguy83: Ideally we wouldn't store the context here but rather pass it into `Run`

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,380 @@
...
+			return
+		}
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(len(ptd))
+
+	// Iterate over the slice of pods to be deleted and delete them in the provider.
+	for _, pod := range ptd {
+		go func(ctx context.Context, pod *corev1.Pod) {
```
cpuguy83: Can we use a semaphore to limit the concurrency here?
No need for anything fancy, can use a buffered channel to handle that.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,380 @@
...
+	if err := pc.server.createOrUpdatePod(ctx, pod); err != nil {
+		err := pkgerrors.Wrapf(err, "failed to sync pod %q in the provider", loggablePodName(pod))
+		span.SetStatus(ocstatus.FromError(err))
+		return err
+	}
+	return nil
+}
+
+// deleteDanglingPods checks whether the provider knows about any pods which Kubernetes doesn't know about, and deletes them.
+func (pc *PodController) deleteDanglingPods() {
```
cpuguy83: Curious what others may think... do we need to stop the world for this collection?

vkubelet/podcontroller.go
pires: Indeed! My bad. Context should never be stored and only passed.

vkubelet/podcontroller.go
pires: Wow, I think github hate my comment here. Damn...

So the panic bubbles up anyway, as we don't override [`ReallyCrash`](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/runtime/runtime.go#L56-L59).

cpuguy83: So no need for the call?

test/e2e/basic_test.go
```diff
@@ -110,15 +116,27 @@ func TestPodLifecycle(t *testing.T) {
...
 	if err := f.DeletePod(pod1.Namespace, pod1.Name); err != nil {
 		t.Fatal(err)
 	}
-
-	// Wait for the "nginx-1-Y" pod to be reported as having been marked for deletion.
-	if err := f.WaitUntilPodDeleted(pod1.Namespace, pod1.Name); err != nil {
-		t.Fatal(err)
-	}
+	// Wait for the delete event to be ACKed.
+	<-pod1Ch
```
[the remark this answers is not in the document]
pires: I have amended the test. If `WaitUntilPodDeleted` returns an error - and it will if no event is observed after 2m - the test now fails.

Done in 3a0874c.

test/e2e/basic_test.go
```diff
@@ -130,6 +148,39 @@ func TestPodLifecycle(t *testing.T) {
...
+	}()
+
+	// Forcibly delete the "nginx-0" pod.
+	if err := f.DeletePodImmediately(pod0.Namespace, pod0.Name); err != nil {
+		t.Fatal(err)
+	}
+	// Wait for the delete event to be ACKed.
+	<-pod0Ch
+	// Give the provider some time to react to the MODIFIED/DELETED events before proceeding.
+	time.Sleep(deleteGracePeriodForProvider)
```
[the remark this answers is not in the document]
pires: The mock provider will react immediately. We just need a small delay (and it could probably be smaller than this) so that the event is processed. Whatever interval we'd use in a loop, we'd most probably succeed after the very first attempt.

vkubelet/podcontroller.go
pires: IMHO a pod in `Failed` state is enough reason to issue a warning to the user, especially in a situation where we're ignoring an "event". If we change this to debug, the user will most probably never be aware of this fact.

vkubelet/podcontroller.go
pires: Removed this in 66059f5.

vkubelet/podcontroller.go
pires: I strongly believe we should start the controller only after this initial reconciliation. FWIW, VK, as is in the master branch, stops the world for [something that is almost identical to this](https://github.com/virtual-kubelet/virtual-kubelet/blob/master/vkubelet/pod.go#L326-L327).

cpuguy83: Ah yes. Was thinking this was being run periodically but I guess it's not. This is fine.

vkubelet/pod.go
```diff
@@ -29,138 +25,18 @@ func addPodAttributes(span *trace.Span, pod *corev1.Pod) {
...
-			logger.WithError(err).Errorf("Failed to create pod")
-		}
+func (s *Server) createOrUpdatePod(ctx context.Context, pod *corev1.Pod) error {
+	// Check if the pod is already known by the provider.
+	// NOTE: Some providers return a non-nil error in their GetPod implementation when the pod is not found while some other don't.
+	// Hence, we ignore the error and just act upon the pod if it is non-nil (meaning that the provider still knows about the pod).
+	if pp, _ := s.provider.GetPod(ctx, pod.Namespace, pod.Name); pp != nil {
+		// The pod has already been created in the provider.
+		// Hence, we return since pod updates are not yet supported.
+		log.G(ctx).Warnf("skipping update of pod %s as pod updates are not supported", pp.Name)
```
cpuguy83: Nit, should use `log.G(ctx).WithField("pod", <podName>)` (could probably even set this in the context before passing it through).

vkubelet/pod.go
```diff
@@ -199,7 +75,17 @@ func (s *Server) createPod(ctx context.Context, pod *corev1.Pod) error {
...
 
-func (s *Server) deletePod(ctx context.Context, pod *corev1.Pod) error {
+func (s *Server) deletePod(ctx context.Context, namespace, name string) error {
+	// Grab the pod as known by the provider.
+	// Since this function is only called when the Pod resource has already been deleted from Kubernetes, we must get it from the provider so we can call "deletePod".
+	// NOTE: Some providers return a non-nil error in their GetPod implementation when the pod is not found while some other don't.
+	// Hence, we ignore the error and just act upon the pod if it is non-nil (meaning that the provider still knows about the pod).
+	pod, _ := s.provider.GetPod(ctx, namespace, name)
+	if pod == nil {
+		// The provider is not aware of the pod, so we just exit.
```
cpuguy83: Nit: we should still create the span and make an annotation that the pod didn't exist on it.

cpuguy83: This could probably be applied in other places as well, and also in general anything that's logged also tends to be a good candidate for adding as an annotation to the span.

pires: Taking notes here for working on this during #348.

vkubelet/pod.go
pires: Taking notes here for working on this during #348.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,378 @@
...
+		if err := pc.server.deletePod(ctx, pod.Namespace, pod.Name); err != nil {
+			err := pkgerrors.Wrapf(err, "failed to delete pod %q in the provider", loggablePodName(pod))
+			span.SetStatus(ocstatus.FromError(err))
+			return err
+		}
+		return nil
+	}
+
+	// Ignore the pod if it is in the "Failed" state.
+	if pod.Status.Phase == corev1.PodFailed {
```
robbiezhang: we should also ignore the Succeeded pods

test/e2e/basic_test.go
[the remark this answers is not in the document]
pires: OK, I'll take note and think about this in my (wip) plan for #368.

vkubelet/podcontroller.go
pires: Done in 0b2e1f8.

vkubelet/pod.go
```diff
@@ -199,7 +75,17 @@ func (s *Server) createPod(ctx context.Context, pod *corev1.Pod) error {
...
+		return nil
```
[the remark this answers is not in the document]
pires: Done in 95c94aa.

vkubelet/podcontroller.go
```diff
@@ -0,0 +1,377 @@
...
+			log.G(ctx).Warnf("expected string in work queue but got %#v", obj)
+			return nil
+		}
+		// Add the current key as an attribute to the current span.
+		span.AddAttributes(trace.StringAttribute("key", key))
+		// Run the syncHandler, passing it the namespace/name string of the Pod resource to be synced.
+		if err := pc.syncHandler(ctx, key); err != nil {
+			if pc.workqueue.NumRequeues(key) < maxRetries {
+				// Put the item back on the work queue to handle any transient errors.
+				log.G(ctx).Warnf("requeuing %q due to failed sync", key)
```
cpuguy83: Can you add `WithError(err)` here?

cpuguy83:
🎉

jeremyrickard:
🎊  this is super cool thanks @pires

vkubelet/podcontroller.go
pires: Taking notes for #348 work.

PR #431, opened 2018-12-01
test: deploy vk when running e2e

pires:
This PR makes the `e2e` target:
- cleanup: delete the VK pod and delete the node from the Kubernetes API, and
- automatically builds and deploys the VK (through `skaffold`).

This way we can be sure that we have a clean environment and that we are testing the very latest changes when developing.

The motivation here is the many times I ran (locally) `make e2e` only to learn later that I hadn't build and deployed the latest changes 🤦‍♂️

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,10 +87,6 @@ jobs:
               sleep 1;
             done
       - run:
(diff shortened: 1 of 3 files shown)


jeremyrickard:
Thanks @pires!

PR #433, opened 2018-12-03
Make pod status updates concurrent.

cpuguy83:
This uses the same number of workers as the pod sync workers.

We may want to start a worker queue here instead, but I think for now
this is ok, particularly because we are limiting the number of
goroutines being spun up at once.

### DIFF
--- a/vkubelet/pod.go
+++ b/vkubelet/pod.go
@@ -3,6 +3,7 @@ package vkubelet
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cpuguy83:
Provider blocking could still happen without this, just that there is 1 goroutine instead of N.
But I think N is fine when it is finite, especially being that it is really always going to be a very small number.

cpuguy83:
Going to go ahead and merge since this is a relatively simple change that is very much needed for even medium sized deployments.

PR #446, opened 2018-12-06
vkubelet: set kubelet version to build version

### LINKED ISSUE #304: Kubernetes version of VK
Maybe we could have VK's version align with K8s and test against that K8s version so we know the functionality is stable. This will make the version make more sense to customers.

pires:
```
$ kubectl get nodes
NAME                 STATUS    ROLES     AGE       VERSION
vkubelet-mock-0      Ready     agent     2s        v1.13.1-vk-v0.7.4-28-g56927973
```

Fixes #304


cpuguy83:
I rather like this, but would like to wait for broader agreement here.

Thanks @pires!

cpuguy83:
Can you change the commit message (and PR title) to reflect that we are setting the kubelet version to the version of VK instead of some random K8s version?

pires:
@cpuguy83 I have addressed your feedback.

Now, I don't follow why more agreement is needed for approving such a simple PR that delivers what was already agreed over a scrum call (yesterday). I mean, even if I don't like the version semantics, I'm not changing it for the sake of keeping it simple. I'm sure a discussion may take place in the future whether to change or not the aforementioned semantics.

But just to make sure I set my expectations right, please, do let me know if I should stop contributing until more resources become available to validate decisions made, and code.

cpuguy83:
@pires We had agreement on the call, but I do want to make sure we take into account any insight from @robbiezhang here.

robbiezhang:
My concern for this change is that some k8s component/tool uses the kubeletVersion property to determine the behavior. Since VK's version doesn't align with the k8s version, and is much lower than the latest k8s version, the k8s component might treat the VK node as an old kubelet, and cause weird behavior.

pires:
> My concern for this change is that some k8s component/tool uses the kubeletVersion property to determine the behavior. 

Can you point me in the direction of code that validates the kubelet version?

robbiezhang:
> > My concern for this change is that some k8s component/tool uses the kubeletVersion property to determine the behavior.
> 
> Can you point me in the direction of code that validates the kubelet version when deciding what behavior to choose?

I saw such change in the k8s release note, and assume the kubelet version is not an optional/literal field.
https://github.com/kubernetes/kubernetes/blame/83f115ec799ae5fad11dc5920c36354fb6653e96/CHANGELOG-1.4.md#L1218

pires:
That piece of code you pointed out is no longer part of Kubernetes, as it was part of the changes for 1.4 (released more than two years ago). But I found something [more recent](https://github.com/kubernetes/kubernetes/pull/69566/files). So yeah, we may play safe here and follow recent Kubernetes versions.

> v1.12.4-virtualkubelet-[vk version]+[vk commit id]

I have an alternative suggestion: bump VK versioning to follow Kubernetes version and keep 3-month release lifecycle too.

pires:
```
$ kubectl get nodes
NAME                 STATUS    ROLES     AGE       VERSION
docker-for-desktop   Ready     master    21d       v1.10.3
vkubelet-mock-0      Ready     agent     2s        v1.13.1-vk-v0.7.4-28-g56927973
```

PR #454, opened 2018-12-08
adding virtual node docs

rbitia:
Adding document links to our official Azure docs


rbitia:
Alright so the circleci failed but that probably shouldn't be happening? @pires

cpuguy83:
Looks like Azure issues.

PR #466, opened 2018-12-14
Refactor http server stuff

cpuguy83:
Move http server handling up to daemon.

This removes the burden of dealing with listeners, http servers, etc in
the core framework.

Instead provide helpers to attach the appropriate routes to the
caller's serve mux.

With this change, the vkubelet package only helps callers setup HTTP
rather than forcing a specific HTTP config on them.


cmd/root.go
```diff
@@ -317,6 +321,7 @@ func initConfig() {
 	if err != nil {
 		logger.WithError(err).Fatal("Error reading API config")
 	}
+	apiConfig.MetricsAddr = metricsAddr
```
robbiezhang: how about moving the assignment into the getAPIConfig function?

cmd/http.go
```diff
@@ -0,0 +1,130 @@
...
+			log.G(ctx).WithError(err).Errorf("Error setting up %s http server", name)
+		}
+	}
+	l.Close()
+}
+
+type apiServerConfig struct {
+	CertPath    string
+	KeyPath     string
+	Addr        string
```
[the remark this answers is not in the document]
pires: I'm assuming `Addr` is (as it is today and below in `MetricsAddr`) a host:port pair. If this is the case, `Addr` is OK.

cmd/root.go
```diff
@@ -114,8 +108,18 @@ This allows users to schedule kubernetes workloads on nodes that aren't running
 			rootContextCancel()
 		}()
 
-		if err := f.Run(rootContext); err != nil && errors.Cause(err) != context.Canceled {
-			log.L.Fatal(err)
+		defer rootContextCancel()
```
pires: Why the duplicated defer?

cpuguy83: Ah, probably don't need it after I changed some stuff around.

PR #468, opened 2018-12-15
Revisit containers env/envFrom processing

### LINKED ISSUE #460: Populate configMap and secret in environment variables for Init containers
`populateEnvironmentVariables` does not populate init container's env variables which are sourced from configMaps or secrets.

### LINKED ISSUE #461: VK ignores `EnvFrom` field for container spec.
From k8s.io/api/core/v1/types.go
```	// List of sources to populate environment variables in the container.
	// The keys defined within a source must be a C_IDENTIFIER. All invalid keys
	// will be reported as an event when the container is starting. When a key exists in multiple
	// sources, the value associated with the last source will take precedence.
	// Values defined by an Env with a duplicate key will take precedence.
	// Cannot be updated.
	// +optional
        EnvFrom []EnvFromSource `json:"envFrom,omitempty" protobuf:"bytes,19,rep,name=envFrom"`
	// List of environment variables to set in the container.
	// Cannot be updated.
	// +optional
	// +patchMergeKey=name
	// +patchStrategy=merge
	Env []EnvVar `json:"env,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,7,rep,name=env"` ```

### LINKED ISSUE #465: pod env var processing fails silently when facing errors
### Environment summary

- Any provider
- As of master branch
- Any Kubernetes version

### Issue Details

From a Kubernetes API perspective, the VK fails silently when an error occurs processing a pod's environment variables references (`envFrom` `Secret` and `ConfigMap`). This means a user will never know a pod is failing to run just by querying the pod status from the API.

### Repo Steps

Create a pod spec with `envFrom` that refers a non-existing secret or configmap. Wait a few seconds and then `kubectl describe pod <pod name>`. It should be assigned to the VK and the VK should be printing errors, but the pod status doesn't change.

### Proposed solution

Record events as specified [here](https://kubernetes.io/docs/concepts/configuration/secret/#restrictions) and [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#restrictions).

### PR DESCRIPTION
**Pull request by pires**
This PR enables the following behavior:
- emit pod events when trying to populate environment variables sourced from configmaps/secrets, optional or not, that are non-existing, or that do exist but do not contain the referenced (optional) key. As a way to improve UX, events are also emitted when the keys are mandatory.
- adds support for populating a pod's environment from the `.envFrom` fields for each of its containers.
- adds support for populating the environment of init containers.

Fixes #460 
Fixes #461 
Fixes #465

### CODE REVIEW

File: manager/resource_test.go [hunk 1]
```diff
@@ -1,21 +1,23 @@
-package manager
+package manager_test
```
robbiezhang: why do you change the package?

pires: To avoid cyclic imports, as the new `test/util` subpackage now imports from manager.

PR #493, opened 2019-01-09
Add async provider pod status updates

cpuguy83:
This adds a new interface that a provider can implement which enables
async notifications of pod status changes rather than the existing loop
which goes through every pod in k8s and checks the status in the
provider.
In practice this should be significantly more efficient since we are not
constantly listing all pods and then looking up the status in the
provider.

For providers that do not support this interface, the old method is
still used to sync state from the provider.

This commit does not update any of the providers to support this
interface.

Closes #493


rbitia:
@cpuguy83 are you looking to merge this asap?

pires:
Please, give me a chance to review this next week. @cpuguy83 please hold me accountable to my request and ping me on slack!

pires:
Is a future goal to remove the loop that exists in the VK?

cpuguy83:
@pires I think so, that was my original intention at least.

pires:
I have mixed feelings about that idea, unless the long-term strategy is to make VK as dumb as possible, and just an abstraction of what a provider should do, i.e. all providers must implement `PodNotifier`. In this case, this seems acceptable to me, as the provider *is* indeed the _product_ the user is adopting and not VK, which becomes just an implementation detail.

If we can agree on the strategy, I do have some comments on the code. As one example, something I am concerned about is the fact we're leaking the VK pod controller workqueue to outside parties, and that seems a not-so-good thing to do.
But let's first make sure we're aligned in terms of strategy for the VK.

cpuguy83:
> But let's first make sure we're aligned in terms of strategy for the VK.

Indeed we are aligned here. The idea here is the provider tells VK that the pod status has changed and then VK can react to that... VK shoudn't have to continuously probe the provider for statuses. In fact I think even just for the interim we can move that loop out of the core and wrap providers that don't implement `PodNotifier` from the daemon code.... not really a huge change but cleans up the core at least.

>  I am concerned about is the fact we're leaking the VK pod controller workqueue to outside parties, and that seems a not-so-good thing to do.

What are your concerns here?
Why do you consider it "leaking"?

pires:
> Why do you consider it "leaking"?

Because it's an implementation detail of the pod controller.

cpuguy83:
But it's not giving it the queue, it's giving it a callback function.

cpuguy83:
per @pires, good with this considering async is opt-in at the provider level.

Merging. We can iterate on any potential issues this causes after.
Thanks!

PR #519, opened 2019-02-05
Make tracing interface to coalesce logging/tracing

### LINKED ISSUE #404: consolidate logging and tracing

cpuguy83:
Allows us to share data between the tracer and the logger so we can simplify log/trace handling where we generally want data to go both places.

Fixes #404

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -30,20 +30,21 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"go.opencensus.io/trace"
+	"github.com/virtual-kubelet/virtual-kubelet/log"
+	logruslogger "github.com/virtual-kubelet/virtual-kubelet/log/logrus"
+	"github.com/virtual-kubelet/virtual-kubelet/manager"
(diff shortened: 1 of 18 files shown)


cpuguy83:
Checking with ACI team on this error

```
--- FAIL: TestGetContainerGroup (0.11s)
	client_test.go:256: api call to https://management.azure.com/subscriptions/ae43b1e3-c35d-4c8c-bc0d-f148b4c52b78/resourceGroups/virtual-kubelet-tests-39f7bf/providers/Microsoft.ContainerInstance/containerGroups/virtual-kubelet-test-container-group?api-version=2018-09-01: got HTTP response status code 404 error code "ResourceNotFound": The Resource 'Microsoft.ContainerInstance/containerGroups/virtual-kubelet-test-container-group' under resource group 'virtual-kubelet-tests-39f7bf' was not found.
```

cpuguy83:
The thing is, it's less a library and more of a framework which can be taken piecemeal.
People can choose to disable logging, actually the default is a no-op logger on the framework side.

cpuguy83:
❤️ Thanks for taking a look.

PR #526, opened 2019-02-22
Add node provider interface

### LINKED ISSUE #427: Feature: use node Lease API
In versions of Kubernetes prior to 1.13, `NodeStatus` is the heartbeat from the `Node`.

Kubernetes 1.13 introduces the node-leases, a lighter, more scalable heartbeat indicator. Node leases are renewed frequently while NodeStatus is reported from node to master only when there is some change or enough time has passed.

More on this Kubernetes feature: https://github.com/kubernetes/enhancements/issues/589

cpuguy83:
This starts the work of having a `NodeProvider` which is responsible for
providing node details.
It splits the responsibilities of node management off to a new
controller.

The primary change here is to add the framework pieces for node
management and move the VK CLI to use this new controller.

It also adds support for node leases where available. This can be
enabled via the command line (disabled by default), but may fall back if
we find that leaess aren't supported on the cluster.

Closes #427


rbitia:
Hey @cpuguy83 how do you expect providers to implement the interface, with accurate numbers or with whatever they see fit? And do you expect some to implement the interface and some to not?

cpuguy83:
@rbitia For the use case of things like ACI, I honestly more expect users to specify the size of the node they want to expose on the cluster. In such a case the node provider lives in the daemon implementation rather than within the provider itself. Actually this is why I've pulled this up into the cmd even though we're still getting that data from the configured provider: https://github.com/virtual-kubelet/virtual-kubelet/blob/a741b868474ac921b1d045c88c3e53fc90b2d24d/cmd/node.go#L19-L21

So basically, this generic do-everything provider goes away (from the core framework) to make things more flexible to integrate with, and the stuff in `cmd/` in this repo is one use-case of such an integration.
In this case it's up to the implementation (`cmd/`) to define an interface that it expects to use in order to configure the core node and pod controllers (framework).

rbitia:
I see so as a user of a provider I myself could go in and specify that the node size would be x cores or x mem?

cpuguy83:
> I myself could go in and specify that the node size would be x cores or x mem?

Yes, I think this is likely something good to add.

cpuguy83:
Marking this ready for review. PTAL. ❤️ 🎉

cpuguy83:
Also introducing gotest.tools/assert here, which is a much simpler assertion framework but also way more flexible than testify. I plan to open a PR to convert the existing provider tests which are using testify over to gotest.tools/assert.

PR #527, opened 2019-02-23
Support building an allow-list of providers

cpuguy83:
Flip the boolean for provider build tags.

Also adds a sub-command to check which providers are compiled in.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,11 @@ DOCKER_IMAGE := virtual-kubelet
 exec := $(DOCKER_IMAGE)
 github_repo := virtual-kubelet/virtual-kubelet
 binary := virtual-kubelet
-build_tags := "netgo osusergo $(VK_BUILD_TAGS)"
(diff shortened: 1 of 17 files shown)


cpuguy83:
I hadn't really thought to just switch these around. In this case we can just do the opposite and have a default `make build` include all providers except those in an exclude list.

The main reason I did the negation build tags was just to get something in to be able to configure it without breaking people and spending a bunch of time on it.
I think this could be flipped around easily enough. The main thing is I didn't want to break the default case of just a plain `make build`.
Does that change (using `aws_provider` instead of `no_aws_provider`) work for you if we still default to building all providers, at least for now?

pires:
But of course. That's why I gave my _approval_ in the first place ;)

cpuguy83:
@pires OK, updated this to flip around those tags. PTAL.

cpuguy83:
Fixed conflicts, all green. Merging. 🎉 🌮

PR #550, opened 2019-03-27
Remove sfmesh provider

cpuguy83:
This is not being supported and as such it was requested (internally) to
be removed.


cpuguy83:
ping @dkkapur @rbitia

cpuguy83:
Ah, I was looking for where that list is...

rbitia:
And one more thing, @lucperkins has it a list on the website

cpuguy83:
Should be all good here.

PR #555, opened 2019-03-30
Add cri as possible provider in helm chart

cpuguy83:
When cri is chosen, containerd is run in the pod.

This makes spinning things up to test CRI a bit simpler.
Not exactly production ready but it's a start.


cpuguy83:
This *almost* works. Having issues injecting CNI config properly.
Right now the CNI config is injected, however for whatever reason cri-containerd is complaining there aren't enough networks (it's looking for at least 2). Of course the default config here only has 1 network but I have an override for that locally.

cpuguy83:
I also want to provide a rootless option just because it'd be cool.

cpuguy83:
I'm going to close this, this should probably go into a separate chart.

PR #557, opened 2019-04-01
Patch the node status instead of updating it.

yashdesai93:
Virtual-kubelet updates the node status periodically.
This change proposes we use the `Patch` API instead of `Update`,
to update the node status.
This avoids overwriting any node updates made by other controllers
in the system, for example an attach-detach controller.
Patch API does a strategic merge instead of overwriting
the entire object, which ensures parallel updates don't overwrite
each other.

Note: `PatchNodeStatus` reduces the time precision to the seconds-level
and therefore I corrected the test for this.

consider two controllers:
CONTROLLER 1 (virtual kubelet)                       | CONTROLLER 2
oldNode := nodes.Get(nodename)                       |
                                                     | node := nodes.Get(nodename)
                                                     | // update status with attached volumes info
                                                     | updateNode := Nodes.UpdateStatus(node)
// update vkubelet info on node status               |
latestNode := Nodes.UpdateStatus(oldNode)            |
<-- latestNode does not contain the volume info added by second controller.

with my patch change:

CONTROLLER 1 (virtual kubelet)                       | CONTROLLER 2
oldNode := Nodes.Get(nodename)                       |
                                                     | node := Nodes.Get(nodename)
                                                     | // update status with attached volumes info
                                                     | updateNode := Nodes.UpdateStatus(node)
node := oldNode.DeepCopy()                           |
// update vkubelet info on node status               |
latestNode := util.PatchNodeStatus(oldNode, node)    |
<-- latestNode contains the volume info added by second controller.

Testing Done: make test


vendor/k8s.io/kubernetes/pkg/util/node/node.go
```diff
@@ -0,0 +1,203 @@
...
+	patchBytes := []byte(fmt.Sprintf(`{"spec":{"podCIDR":%s}}`, raw))
+
+	if _, err := c.CoreV1().Nodes().Patch(string(node), types.StrategicMergePatchType, patchBytes); err != nil {
+		return fmt.Errorf("failed to patch node CIDR: %v", err)
+	}
+	return nil
+}
+
+// PatchNodeStatus patches node status.
+func PatchNodeStatus(c v1core.CoreV1Interface, nodeName types.NodeName, oldNode *v1.Node, newNode *v1.Node) (*v1.Node, []byte, error) {
```
cpuguy83: I don't like that we're pulling in the entire core interface for this function.
Can we just copy/paste these two functions (`PatchNodeStatus` and `preparePatchBytesforNodeStatus`) but just specify the nodes client as an argument instead of the whole corev1?

vkubelet/node.go
```diff
@@ -328,18 +332,20 @@ func UpdateNodeStatus(ctx context.Context, nodes v1.NodeInterface, n *corev1.Nod
...
 
-	updated, err := nodes.UpdateStatus(node)
+	// Patch the node status to merge other changes on the node.
+	updated, _, err := nodeutil.PatchNodeStatus(c, types.NodeName(n.Name), oldNode, node)
 	if err != nil {
 		return nil, err
 	}
 
-	log.G(ctx).WithField("resourceVersion", updated.ResourceVersion).Debug("updated node status in api server")
+	log.G(ctx).WithField("Conditions", updated.Status.Conditions).
```
cpuguy83: let's call this field `node.Status.Conditions`, and maybe also keep `resourceVersion` (but call it `node.resourceVersion`

PR #559, opened 2019-04-01
Resource manager: add service lister and remove unused lock.

yashdesai93:
This will be used to set up env var for services within VK similar to what we do for configmaps and secrets.


cpuguy83:
I don't see where this is being used. Can you talk about the use case?

yashdesai93:
Hi Brian, I want to use this to set up service env var for the pods.

PR #564, opened 2019-04-03
Fix node create after delete

cpuguy83:
node.ResourceVersion must not be set when creating a node.
    This issue prevents vk from resolving issues after the vk node instance
    has been deleted (for whatever reason).


pires:
Why was this merged with no reviews?

cpuguy83:
Are you around to review things?

We can fix if there is an issue.
Bug fix is minor. Biggest thing was getting e2e working locally.

pires:
There is an issue #693

PR #568, opened 2019-04-04
Add support for go modules and rebuild vendor directory

hectorj2f:
I took the initiative and added support for go modules for the project. I adapted the required code and removed the existing dependencies to dep.

I ran the `e2e` tests on my minikube and they worked.


hectorj2f:
Ouch! It seems something might be wrong inside the circleci container, I'll check what is wrong.

cpuguy83:
Please see https://github.com/virtual-kubelet/virtual-kubelet/pull/514 for prior work on this.

scripts/validate/gomod.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+exit_code=0
+
+echo "==> Running go mod verify check <=="
+
+GO111MODULE=on go mod verify || exit_code=1
```
cpuguy83: I don't think this is doing what we need it to do.

hectorj2f: I made some changes to the code according to what you initially proposed in #514.

I am facing some error in the builds such as https://circleci.com/gh/hectorj2f/virtual-kubelet/29?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link and https://circleci.com/gh/hectorj2f/virtual-kubelet/28?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link which doesn't seem fully related to any mistake in my changes. Perhaps I am missing something in relation how the CI of this project works.

Besides, I cannot access to the build in https://virtual-kubelet.visualstudio.com/d2adf84f-67a6-4304-a3b5-0235942fdba2/_build/results?buildId=584. So it is hard to debug any potential issue :/.

cpuguy83:
```
/go/src/github.com/virtual-kubelet/virtual-kubelet/vendor/github.com/vmware/govmomi/vim25/debug (vendor tree)
	/usr/local/go/src/github.com/vmware/govmomi/vim25/debug (from $GOROOT)
	/go/src/github.com/vmware/govmomi/vim25/debug (from $GOPATH)
```

hectorj2f:
@cpuguy83 I am finding that `e2e` test seems to be a bit flaky, I managed to pass it https://circleci.com/gh/virtual-kubelet/virtual-kubelet/2760 when re-run it without making any relevant changes to the code 🤔. Do you have any thoughts on that ?

cpuguy83:
These tests aren't generally flaky... super weird.

hectorj2f:
okay, I'll investigate it 🤔

sargun:
FWIW: Circle-CI has been mucking with their environment recently.

.circleci/config.yml
```diff
@@ -2,7 +2,9 @@ version: 2
 jobs:
   validate:
     docker:
-      - image: circleci/golang:1.10
+      - image: circleci/golang:1.11
```
sargun: Any reason not to use Go 1.12, where the module cache has been ironed out a little bit better.

sargun: I think it's also a lot less work, given a lot of the places where you're explicitly opting into Go module support, 1.12 does automatically.

.circleci/config.yml
```diff
@@ -65,8 +69,22 @@ jobs:
       MINIKUBE_WANTUPDATENOTIFICATION: false
       MINIKUBE_WANTREPORTERRORPROMPT: false
       SKAFFOLD_VERSION: v0.18.0
+      GO111MODULE: "on"
```
sargun: Above, don't you want to build this outside the global gopath? In future versions of Go, it wont enable GO modules by default if you build with the project checked out in the gopath.

.circleci/config.yml
hectorj2f: @sargun I will give a try to `1.12` and remove the unnecessary flags.

sargun:
@hectorj2f This will become easier in a bit: https://github.com/virtual-kubelet/virtual-kubelet/pull/600

sargun:
And: https://github.com/virtual-kubelet/virtual-kubelet/pull/606

.circleci/config.yml
hectorj2f: I will have to keep it:
```
circleci@17ecfd2c0929:/go/src/github.com/virtual-kubelet/virtual-kubelet$ go version
go version go1.12.5 linux/amd64
circleci@17ecfd2c0929:/go/src/github.com/virtual-kubelet/virtual-kubelet$ go mod tidy
go: modules disabled inside GOPATH/src by GO111MODULE=auto; see 'go help modules'
```

sargun: You should move the project to build outside of the global gopath. Basically, if you change working dir to something like /builds or /foo, it should fix this.

.circleci/config.yml
```diff
@@ -65,8 +69,22 @@ jobs:
...
+      GOVERSION: 1.12.1
```
sargun: If we're doing 1.12, why not 1.12.5?

sargun: Also, isn't golang already installed from the docker image?

.circleci/config.yml
cpuguy83: I'd rather not change how we build just to support modules.
You *can* build outside of gopath and things will just work, or you can use GO111MODULE=on and it will work.
In any case, I think the `Makefile` should set this (`GO111MODULE ?= on`)
As I recall in the future `on` will be the default even when in GOPATH, however even this has been postponed.

.circleci/config.yml
hectorj2f: It installs an old version of golang, that is why I had to install the golang version here. 

No problem on using 1.12.5 ...

Makefile.e2e
```diff
@@ -24,7 +24,7 @@ bin/e2e:
 	@mkdir -p bin/e2e
 
 bin/e2e/virtual-kubelet: bin/e2e
-	GOOS=linux GOARCH=amd64 $(MAKE) OUTPUT_DIR=$(@D) build
+	GO111MODULE=on GOOS=linux GOARCH=amd64 $(MAKE) OUTPUT_DIR=$(@D) build
```
cpuguy83: Let's set this like `GO111MODULE ?= on` in this Makefile.

Makefile
```diff
@@ -5,6 +5,8 @@ exec := $(DOCKER_IMAGE)
 github_repo := virtual-kubelet/virtual-kubelet
 binary := virtual-kubelet
 
+export GO111MODULE := on
```
cpuguy83: Please use `?=`

scripts/validate/gomod.sh
```diff
@@ -0,0 +1,22 @@
...
+else
+  echo "go mod tidy passed."
+fi
+
+git diff --exit-code go.mod go.sum || exit_code=$?
+
+if [ ${exit_code} -eq 0 ]; then
+	exit 0
+fi
+
```
cpuguy83: I think we need to check vendor as well.

Makefile
```diff
@@ -49,12 +51,12 @@ release: build goreleaser
 
 ##### =====> Utility targets <===== #####
 
-.PHONY: clean test list cover format docker deps
+.PHONY: clean test list cover format docker vendor
 
-deps: setup
+vendor: setup
```
cpuguy83: No need for `setup` here

Makefile
```diff
@@ -171,6 +172,8 @@ Q := $(if $V,,@)
 
 $(gobin_tool):
 	GO111MODULE=off go get -u github.com/myitcv/gobin
+	# NOTE: Add the GOPATH to the PATH to find gobin
+	export PATH=$(shell go env GOPATH)/bin:$(PATH)
```
cpuguy83: This is not full-proof as `GOPATH` can be a list.
Probably it's best to leave this to the caller to make sure this is setup.

hectorj2f: I didn't like this change, I removed. The problem comes from 
`gobin_tool ?= $(shell which gobin || echo $(GOPATH)/gobin)` which should be `gobin_tool ?= $(shell which gobin || echo $(GOPATH)/bin/gobin)` with `$(GOPATH)/bin/gobin`. That way it'll find the gobin path. Also, I believe we should change the references to gobin in the Makefile to then use `$(gobin_tool)`. Do you like that solution ?

sargun: @hectorj2f yeah, not using $(gobin_tool) was a mistake on my part.

sargun:
@hectorj2f Can you pull in my service account config: https://github.com/sargun/virtual-kubelet/blob/no-more-hard-deletes/hack/skaffold/virtual-kubelet/base.yml 

It looks like you may have updated a k8s library which depends on slightly more perms?

hectorj2f:
@sargun Done. That fixed the bug in the `e2e` tests with the help of the PR #608

go.mod
```diff
@@ -0,0 +1,166 @@
...
+	go.uber.org/zap v1.9.1 // indirect
+	golang.org/x/net v0.0.0-20190311183353-d8887717615a
+	golang.org/x/sync v0.0.0-20190227155943-e225da77a7e6
+	google.golang.org/grpc v1.19.0
+	gopkg.in/asn1-ber.v1 v1.0.0-20181015200546-f715ec2f112d // indirect
+	gopkg.in/gorethink/gorethink.v4 v4.1.0 // indirect
+	gopkg.in/mgo.v2 v2.0.0-20180705113604-9856a29383ce // indirect
+	gopkg.in/ory-am/dockertest.v2 v2.2.3 // indirect
+	gotest.tools v0.0.0-20181223230014-1083505acf35
+	k8s.io/api v0.0.0-20181213150558-05914d821849
```
sargun: Do we want to lock these at v1.13.X as well?

vendor/github.com/vmware/vic/lib/apiservers/engine/backends/cache/container_cache.go
```diff
@@ -0,0 +1,199 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
sargun: Why is it pulling in vic?

sargun:
It seems like there a bunch of implicit version changes in here. I think some of them might have to be changed. How do you feel about that?

Makefile
```diff
@@ -152,20 +153,20 @@ allpackages = $(if $(__allpackages),,$(eval __allpackages := $$(_allpackages)))$
 
 .PHONY: goimports
 goimports: $(gobin_tool)
-	gobin -d $(goimports)
+	$(gobin_tool) -d $(goimports)
```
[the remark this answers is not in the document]
sargun: As a side-note, do you wanna break this out, as a undo-my-dumb changes fix?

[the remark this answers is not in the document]
hectorj2f: #614 👈  :)

vendor/github.com/vmware/vic/lib/apiservers/engine/backends/cache/container_cache.go
hectorj2f: Fixed

go.mod
hectorj2f: Updated

.circleci/config.yml
```diff
@@ -65,8 +69,22 @@ jobs:
...
       - checkout
+      - run:
+          name: Install go
+          command: |
+            go version
+            go env GOROOT
+            mkdir tmp
+            cd tmp
+            sudo rm -rf /usr/local/go
+            wget https://storage.googleapis.com/golang/go$GOVERSION.$OS-$ARCH.tar.gz
```
cpuguy83: Why are we installing go?

Makefile
```diff
@@ -34,9 +36,10 @@ safebuild:
 .PHONY: build
 build: build_tags := netgo osusergo
 build: OUTPUT_DIR ?= bin
-build: authors
+build: vendor authors
```
cpuguy83: build should not require vendor since we commit vendor locally.

Makefile
```diff
@@ -34,9 +36,10 @@ safebuild:
...
 	@echo "Building..."
-	$Q CGO_ENABLED=0 go build -a --tags '$(shell scripts/process_build_tags.sh $(build_tags) $(VK_BUILD_TAGS))' -ldflags '-extldflags "-static"' -o $(OUTPUT_DIR)/$(binary) $(if $V,-v) $(VERSION_FLAGS) ./cmd/$(binary)
+	# TODO: consider moving to a GOPROXY in the future
+	$Q GOFLAGS=-mod=vendor CGO_ENABLED=0 go build -a --tags '$(shell scripts/process_build_tags.sh $(build_tags) $(VK_BUILD_TAGS))' -ldflags '-extldflags "-static"' -o $(OUTPUT_DIR)/$(binary) $(if $V,-v) $(VERSION_FLAGS) ./cmd/$(binary)
```
cpuguy83: Can you just set `-mod=vendor` inline instead of an env var?

Makefile
```diff
@@ -49,12 +52,12 @@ release: build goreleaser
 
 ##### =====> Utility targets <===== #####
 
-.PHONY: clean test list cover format docker deps
+.PHONY: clean test list cover format docker vendor
 
-deps: setup
+vendor:
```
cpuguy83: Can you add another target `mod` which runs `go mod tidy`?

Makefile.e2e
cpuguy83: No need for this change?

cpuguy83:
Also need to remove Gopkg.lock

.circleci/config.yml
sargun: I think the machine image apparently has an older version of Go? Maybe we could cache, or use a newer machine image.

cpuguy83: I don't know how that's possible because this is a container image.

hectorj2f:
@cpuguy83 I wanted to share one concern I have when trying to use the "same" versions used in Gopkg* files. 

After following the steps `go mod init` + `go mod download` I get a list of modules which match the ones defined in Gopkg. However once I run `go mod tidy` (it would be part of the validate scripts in `scripts/validate/gomod.sh`) some of these dependencies would be updated-pruned. Some of these new changes after running `go mod tidy` would enforce using a newer version of `k8s.io/api` which doesn't contain the package `k8s.io/api/admissionregistration/v1alpha1`. 

Should we comment `go mod tidy` execution in `scripts/validate/gomod.sh` ? or should I upgrade the code where `k8s.io/api/admissionregistration/v1alpha1` is replaced by `k8s.io/api/admissionregistration/v1beta1` ?

sargun:
I think that without moving the providers out of core, this is nearly impossible. I tried to split it into two go "things" one go module under vkubelet/ and one using the Gopkg files under `providers/`. Unfortunately, this seems to break Go tooling.

I think the solutions might be aggressively turning vkubelet into a library, and creating self-contained executables per provider.

Even if we put each provider into their own go repo, and then these transitively reference those, there is the messiness of ensuring that each provider's deps are able to be upgraded to go modules.

hectorj2f:
@cpuguy83 I addressed your comments. The kubernetes dependencies are now using `kubernetes-1.13.1`.

cpuguy83:
The syntax we discussed on the phone goes outside of `require`... e.g.

```
require (
  k8s.io/client-go <random version>
)

replace k8s.io/client-go => k8s.io/client-go kubernetes-1.13.1
```

This would need to be done for multiple of the k8s deps.
I'm not sure what other deps may require this.

pires:
I had to do it for an operator and it was tricky, but eventually worked. https://github.com/travelaudience/cloudsql-postgres-operator/blob/master/go.mod#L5-L13

hectorj2f:
> The syntax we discussed on the phone goes outside of require... e.g.

Thanks @cpuguy83. I'll take care of these changes today.

pires:
@hectorj2f is probably wise to wait for the providers to be removed, something @cpuguy83 and @sargun were discussing yesterday, over slack.

hectorj2f:
@cpuguy83 @sargun can you confirm what @pires just mentioned ? so I don't work on this unnecessarily  :/

cpuguy83:
@hectorj2f Yes, planning to move out providers this week.

pires:
@hectorj2f it's time! 🎉

cpuguy83:
Carrying this PR in https://github.com/virtual-kubelet/virtual-kubelet/pull/671

Thanks @hectorj2f

PR #573, opened 2019-04-10
Set container env var using services.

yashdesai93:
This change introduces setting container env var using services.
It also adds tests for the same.

### DIFF
--- a/vkubelet/env.go
+++ b/vkubelet/env.go
@@ -8,10 +8,14 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	apivalidation "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/tools/record"
 	podshelper "k8s.io/kubernetes/pkg/apis/core/pods"
+	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 	fieldpath "k8s.io/kubernetes/pkg/fieldpath"
+	"k8s.io/kubernetes/pkg/kubelet/envvars"
 
 	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"github.com/virtual-kubelet/virtual-kubelet/manager"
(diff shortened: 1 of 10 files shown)


vkubelet/env.go
```diff
@@ -347,6 +400,29 @@ loop:
 			continue loop
 		}
 	}
+
+	// enableServiceLinks defaults to true.
+	enableServiceLinks := true
+	if pod.Spec.EnableServiceLinks != nil {
```
cpuguy83: The kubelet errors out if this is nil, I think we should do the same.

vkubelet/env.go
```diff
@@ -89,6 +95,53 @@ func populateContainerEnvironment(ctx context.Context, pod *corev1.Pod, containe
...
+	for i := range services {
+		service := services[i]
+		// ignore services where ClusterIP is "None" or empty
+		if !v1helper.IsServiceIPSet(service) {
+			continue
+		}
+		serviceName := service.Name
+
+		// We always want to add environment variabled for master services
+		// from the master service namespace, even if enableServiceLinks is false.
```
cpuguy83: I understand this is basically a copy/paste from kubelet, but can we come up with a better than "master"?

vkubelet/env.go
```diff
@@ -89,6 +95,53 @@ func populateContainerEnvironment(ctx context.Context, pod *corev1.Pod, containe
...
+		if service.Namespace == metav1.NamespaceDefault && masterServices.Has(serviceName) {
+			if _, exists := serviceMap[serviceName]; !exists {
+				serviceMap[serviceName] = service
+			}
+		} else if service.Namespace == ns && enableServiceLinks {
+			serviceMap[serviceName] = service
+		}
+	}
+
+	mappedServices := []*corev1.Service{}
```
cpuguy83:
```suggestion
	mappedServices := make([]*corev1.Service, 0, len(serviceMap))
```

vkubelet/env.go
yashdesai93: My only concern was that some e2e tests might fail, but let's see.

yashdesai93:
@cpuguy83 failing `pod.Spec.EnableServiceLinks` being nil led to the timeout. I haven't figured out the exact cause, but I have added a TODO and for now use the default when nil. I am hoping that's okay.

PR #574, opened 2019-04-15
Use I/O stream for provider logs interface

cpuguy83:
Providers must still update the implementation to actually gain any
benefit here, but this makes the provider interface a bit more sane.


cpuguy83:
The only provider I've really changed is aws... which I was considering just dropping back to minimal changes.

pires:
Yes, please do.

cpuguy83:
Updated.

pires:
@cpuguy83 please fix the conflict and let's have this merged.

PR #578, opened 2019-04-19
Refactor exec interface

cpuguy83:
This removes the dependence on remotecommand in providers as well as the
need to expose provider ID's for the sake of the ExecInContainer API.

### DIFF
--- a/vkubelet/api/exec.go
+++ b/vkubelet/api/exec.go
@@ -1,7 +1,8 @@
 package api
 
 import (
-	"fmt"
+	"context"
+	"io"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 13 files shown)


vkubelet/api/exec.go
```diff
@@ -26,30 +37,121 @@ func PodExecHandlerFunc(backend remotecommand.Executor) http.HandlerFunc {
...
 
 		idleTimeout := time.Second * 30
 		streamCreationTimeout := time.Second * 30
 
-		remotecommand.ServeExec(w, req, backend, fmt.Sprintf("%s-%s", namespace, pod), "", container, command, streamOpts, idleTimeout, streamCreationTimeout, supportedStreamProtocols)
+		ctx, cancel := context.WithCancel(context.TODO())
+		defer cancel()
+
+		exec := &containerExecContext{ctx: ctx, b: backend, pod: pod, namespace: namespace, container: container}
+		remotecommand.ServeExec(w, req, exec, "", "", container, command, streamOpts, idleTimeout, streamCreationTimeout, supportedStreamProtocols)
```
cpuguy83: I would ultimately like to have not used this function, however all the stream setup is handled in here and most of it is not exported.
There's just a lot to handle in here and I did not want to duplicate it and have potential divergence in the future.
This includes support for different websocket protocol versions as well as SPDY.

Issue #583, opened 2019-04-24
Pod status is inconsistent with containers status cause a nil pointer dereference error.


zlseu-edu:
---

### Environment summary

Provider : Alibaba

Version : master

K8s Master Info : ACS

Install Method : kubectl

### Issue Details

Examples are as below. `Nginx` Container will exit after executing the command `ls -l`, all pods will in `CrashLoopBackOff` state. There has a small probability of `VK` crash again and again.

``` yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 10
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx
        command: ["/bin/sh"]
        args: ["ls", "-l"]
        ports:
        - containerPort: 80
      nodeSelector:
        kubernetes.io/role: agent
        beta.kubernetes.io/os: linux
        type: virtual-kubelet
      tolerations:
      - key: virtual-kubelet.io/provider
        operator: Exists
```

### Repo Steps

Crash info show a dereference nil poniter error in `c.State.Running.StartedAt` of file `vkubelet/pod.go` line `200` :

``` go
// Only change the status when the pod was already up
// Only doing so when the pod was successfully running makes sure we don't run into race conditions during pod creation.
if pod.Status.Phase == corev1.PodRunning || pod.ObjectMeta.CreationTimestamp.Add(time.Minute).Before(time.Now()) {
    // Set the pod to failed, this makes sure if the underlying container implementation is gone that a new pod will be created.
    pod.Status.Phase = corev1.PodFailed
    pod.Status.Reason = "NotFound"
    pod.Status.Message = "The pod status was not found and may have been deleted from the provider"
    for i, c := range pod.Status.ContainerStatuses {
        // Skip waiting and Terminated status
        if c.State.Waiting != nil || c.State.Terminated != nil {
            continue
        }
        pod.Status.ContainerStatuses[i].State.Terminated = &corev1.ContainerStateTerminated{
            ExitCode:    -137,
            Reason:      "NotFound",
            Message:     "Container was not found and was likely deleted",
            FinishedAt:  metav1.NewTime(time.Now()),
            StartedAt:   c.State.Running.StartedAt,
            ContainerID: c.ContainerID,
        }
        pod.Status.ContainerStatuses[i].State.Running = nil
    }
[...]

PR #584, opened 2019-04-24
fix dereference nil pointer error when updatePodStatus

zlseu-edu:
A pod in running state doesn't means its container also in running states. Like issue #583 
I catch some logs show a pod in running state but container in Waiting state:

```
{nginx 
{ContainerStateWaiting{Reason:CrashLoopBackOff,Message:Back-off 1m20s restarting failed container=nginx pod=default-nginx-deployment-7d6f46c7fb-bmp8w_eci-2ze4nwuvbq3zw94g39z5(XbHMbjeBvZwcJXyXLoet-default-nginx-deployment-7d6f46c7fb-bmp8w),} nil nil} 
{nil nil &ContainerStateTerminated{ExitCode:127,Signal:0,Reason:Error,Message:,StartedAt:2019-04-24 14:31:48 +0000 UTC,FinishedAt:2019-04-24 14:31:48 +0000 UTC,ContainerID:,}} 
false 4 nginx  eci://a657e387435a1b32d829058bfe7741d80ee5bf6f7505d27823ddeb7a3a9b3000}
```

Logic in `updatePodStatus` should skipping the `waiting` and `terminated` state, then change `running` to `terminated`.

### DIFF
--- a/vkubelet/pod.go
+++ b/vkubelet/pod.go
@@ -192,6 +192,10 @@ func (s *Server) updatePodStatus(ctx context.Context, pod *corev1.Pod) error {
 			pod.Status.Reason = "NotFound"
 			pod.Status.Message = "The pod status was not found and may have been deleted from the provider"
 			for i, c := range pod.Status.ContainerStatuses {
+				// Skip waiting and Terminated status
+				if c.State.Waiting != nil || c.State.Terminated != nil {
+					continue
+				}
 				pod.Status.ContainerStatuses[i].State.Terminated = &corev1.ContainerStateTerminated{
 					ExitCode:    -137,
 					Reason:      "NotFound",


zlseu-edu:
I signed it

pires:
If any container in a pod is not running, then the pod cannot be in state running.

vkubelet/pod.go
```diff
@@ -192,6 +192,10 @@ func (s *Server) updatePodStatus(ctx context.Context, pod *corev1.Pod) error {
 			pod.Status.Reason = "NotFound"
 			pod.Status.Message = "The pod status was not found and may have been deleted from the provider"
 			for i, c := range pod.Status.ContainerStatuses {
+				// Skip waiting and Terminated status
+				if c.State.Waiting != nil || c.State.Terminated != nil {
+					continue
+				}
```
pires: I don't understand how this fixes _dereference nil pointer_.

zlseu-edu:
> If any container in a pod is not running, then the pod cannot be in state running.

Please [look the stack info](https://github.com/virtual-kubelet/virtual-kubelet/issues/583#issuecomment-486490494)

vkubelet/pod.go
```diff
@@ -192,6 +192,10 @@ func (s *Server) updatePodStatus(ctx context.Context, pod *corev1.Pod) error {
 			pod.Status.Reason = "NotFound"
 			pod.Status.Message = "The pod status was not found and may have been deleted from the provider"
 			for i, c := range pod.Status.ContainerStatuses {
+				// Skip waiting and Terminated status
+				if c.State.Waiting != nil || c.State.Terminated != nil {
```
cpuguy83: Instead of skipping, let's just change how we are setting these values.
When in Terminated or Running we should have a StartedAt, and in Terminated it should also have a proper terminated time as well.

zlseu-edu: Agree. I rewrite the bugfix code (`still skip the waiting state`).

cpuguy83:
Do you think you can write a unit test for this?

PR #596, opened 2019-05-03
Do not swallow errors from the zpages server silently

sargun:
If the zpages server exits for any reason, or is unable to bind,
rather than exiting silently, throw an error.


cmd/virtual-kubelet/commands/root/tracing.go
```diff
@@ -98,5 +98,7 @@ func setupZpages(ctx context.Context) {
 	}
 	mux := http.NewServeMux()
 	zpages.Handle(mux, "/debug")
-	http.ListenAndServe(p, mux)
+	err := http.ListenAndServe(p, mux)
+	// This should never terminate
+	log.G(ctx).WithError(err).Fatal("Zpages server exited")
```
cpuguy83: Couple of things here...
- We should check for an error.
- I think ListenAndServe will always return an error unless something changed recently.
- Can this return an error instead of calling Fatal?

sargun: 👍

sargun: If it always returns an error, why would we check for an error?

sargun:
Changed it to return an Error instead of Fatal. Not sure what you mean by "Checking" the error, given that anything that returns from ListenAndServe is not good.

sargun:
I mean, from the docs:
>  ListenAndServe always returns a non-nil error.

cpuguy83:
We could switch this to listen and then serve and only do the serve in a goroutine.

sargun:
@cpuguy83  I changed this to Listen, then run

cmd/virtual-kubelet/commands/root/tracing.go
```diff
@@ -96,7 +97,14 @@ func setupZpages(ctx context.Context) {
...
+	listener, err := net.Listen("tcp", p)
+	if err != nil {
+		log.G(ctx).WithError(err).Error("Cannot bind to ZPAGES PORT, cannot setup listener")
+	}
 	mux := http.NewServeMux()
 	zpages.Handle(mux, "/debug")
-	http.ListenAndServe(p, mux)
+	go func() {
+		// This should never terminate, if it does, it will always terminate with an error
+		log.G(ctx).WithError(http.Serve(listener, mux)).Error("Zpages server exited")
```
cpuguy83: I'd rather not log an error if it's not an error.

> After Shutdown or Close, the returned error is ErrServerClosed

We should be able to check for this and only log the error if it's not that error.

cmd/virtual-kubelet/commands/root/tracing.go
```diff
@@ -96,7 +97,14 @@ func setupZpages(ctx context.Context) {
 	if p == "" {
 		log.G(ctx).Error("Missing ZPAGES_PORT env var, cannot setup zpages endpoint")
 	}
+	listener, err := net.Listen("tcp", p)
+	if err != nil {
+		log.G(ctx).WithError(err).Error("Cannot bind to ZPAGES PORT, cannot setup listener")
```
cpuguy83: This should be an early return.

cmd/virtual-kubelet/commands/root/tracing.go
```diff
@@ -96,7 +97,14 @@ func setupZpages(ctx context.Context) {
 	if p == "" {
 		log.G(ctx).Error("Missing ZPAGES_PORT env var, cannot setup zpages endpoint")
```
cpuguy83: Hmm... think we are missing an early return here?

PR #586, opened 2019-04-25
[WIP] Support configmap/secret update notifications

cpuguy83:

PR #597, opened 2019-05-03
Add the concept of startup timeout

sargun:
This adds two concepts, where one encompasses the other.

### Startup timeout
Startup timeout is how long to wait for the entire kubelet
to get into a functional state. Right now, this only waits
for the pod informer cache for the pod controllerto become
in-sync with API server, but this could be extended to other
informers (like secrets informer).

### Wait For Startup
This changes the behaviour of the virtual kubelet to wait
for the pod controller to start before registering the node.

It is to avoid the race condition where the node is registered,
but we cannot actually do any pod operations.

### DIFF
--- a/cmd/virtual-kubelet/commands/root/flag.go
+++ b/cmd/virtual-kubelet/commands/root/flag.go
@@ -78,7 +78,7 @@ func installFlags(flags *pflag.FlagSet, c *Opts) {
 	flags.StringVar(&c.TraceSampleRate, "trace-sample-rate", c.TraceSampleRate, "set probability of tracing samples")
 
 	flags.DurationVar(&c.InformerResyncPeriod, "full-resync-period", c.InformerResyncPeriod, "how often to perform a full resync of pods between kubernetes and the provider")
-
+	flags.DurationVar(&c.StartupTimeout, "startup-timeout", c.StartupTimeout, "How long to wait for the virtual-kubelet to start")
 }
 
 func getEnv(key, defaultValue string) string {
--- a/cmd/virtual-kubelet/commands/root/root.go
+++ b/cmd/virtual-kubelet/commands/root/root.go
@@ -17,6 +17,7 @@ package root
 import (
 	"context"
 	"os"
+	"time"
 
 	"github.com/cpuguy83/strongerrors"
 	"github.com/pkg/errors"
@@ -54,6 +55,9 @@ This allows users to schedule kubernetes workloads on nodes that aren't running 
 }
 
 func runRootCommand(ctx context.Context, c Opts) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	if ok := providers.ValidOperatingSystems[c.OperatingSystem]; !ok {
 		return strongerrors.InvalidArgument(errors.Errorf("operating system %q is not supported", c.OperatingSystem))
 	}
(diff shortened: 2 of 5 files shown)


sargun:
The test that failed in Circle CI succeeded locally.

cmd/virtual-kubelet/commands/root/root.go
```diff
@@ -160,12 +161,23 @@ func runRootCommand(ctx context.Context, c Opts) error {
...
 
+	// TODO: Consider waiting for up to startup timeout seconds for this to close
+	//
+	// currently, startup timeout is just used in waiting for the VK to start. It has its own timeout
+	// mechanism, so this is redundant if we add timeouts here.
+	//
+	// Even though it wont close the channel if the VK fails to start, it will cause a fatal log
+	// message to be emitted, shutting down the process
+	if c.WaitForStartup {
+		<-started
```
cpuguy83: I needed something similar for #586 
I decided to add a `.Ready()` function instead of accepting a channel.
WDYT? https://github.com/virtual-kubelet/virtual-kubelet/pull/586/files#diff-c334be5faaed7b87440b6090f6dd3f6fR96

cmd/virtual-kubelet/commands/root/flag.go
```diff
@@ -78,6 +78,8 @@ func installFlags(flags *pflag.FlagSet, c *Opts) {
 	flags.StringVar(&c.TraceSampleRate, "trace-sample-rate", c.TraceSampleRate, "set probability of tracing samples")
 
 	flags.DurationVar(&c.InformerResyncPeriod, "full-resync-period", c.InformerResyncPeriod, "how often to perform a full resync of pods between kubernetes and the provider")
+	flags.DurationVar(&c.StartupTimeout, "startup-timeout", c.StartupTimeout, "How long to wait for the virtual-kubelet to start")
+	flags.BoolVar(&c.WaitForStartup, "wait-for-startup", c.WaitForStartup, "Wait for pod controller to start, before starting up the node")
```
cpuguy83: I'm not sure we need this. If the startup times out we should just exit.

sargun: I don't know what the project's perspective is on breaking behaviour. If we do something like waiting on a channel, or a Ready() method, then the node wont come up until the podcontroller is started. 

If you think it's okay changing behaviour, I'm all for making things simpler.

cmd/virtual-kubelet/commands/root/root.go
sargun: I like the Ready() approach better, and I'm okay with changing it. I just feel weird that it introduces state to the struct that's not necessary, and IMHO, unless there is a good reason to add members to a struct, don't do it.

sargun:
One question: in your PR, your `Ready()` waits for pod deletion to finish. That is bounded by the provider. If we have a timeout over that, it could be problematic. Do we want to break out the multiple channels?

i.e.
```
InSync() // All informers in sync
Ready() // Providers are synched and ready to go
Running() // We're launching pods and such
```

sargun:
I changed it to use the ready-style interface.

cmd/virtual-kubelet/commands/root/root.go
```diff
@@ -178,6 +189,22 @@ func runRootCommand(ctx context.Context, c Opts) error {
 	return nil
 }
 
+func waitForVK(ctx context.Context, time time.Duration, vk *vkubelet.Server) error {
+	ctx, cancel := context.WithTimeout(ctx, time)
+	defer cancel()
+
+	// The VK / PC close this channel once they are ready. We wait on it before we start the node
```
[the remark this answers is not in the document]
sargun: Pushed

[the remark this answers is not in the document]
sargun: Tried to rephrase it to be clearer

PR #595, opened 2019-05-03
Add the ability to configure klog from VK

sargun:
All of Kubernetes logging is based on klog. Klog currently does
not output any logging information to logrus, so you're flying
somewhat blind to Kubernetes internals.

This exposes the full set of configurables that klog offers,
but decorates (prefixes) the klog configuration with "klog".

The help output looks something like this:

```
      --disable-taint                         disable the virtual-kubelet node taint
      --enable-node-lease                     use node leases (1.13) for node heartbeats
      --full-resync-period duration           how often to perform a full resync of pods between kubernetes and the provider (default 1m0s)
  -h, --help                                  help for virtual-kubelet
      --klog.alsologtostderr                  log to standard error as well as files
      --klog.log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --klog.log_dir string                   If non-empty, write log files in this directory
      --klog.log_file string                  If non-empty, use this log file
      --klog.logtostderr                      log to standard error instead of files
      --klog.skip_headers                     If true, avoid header prefixes in the log messages
      --klog.stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
      --klog.v Level                          log level for V logs
      --klog.vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
      --kubeconfig string                     kube config file to use for connecting to the Kubernetes API server (default "/Users/sargun/.kube/config")
      --log-level string                      set the log level, e.g. "debug", "info", "warn", "error" (default "info")
      --metrics-addr string                   address to listen for metrics/stats requests (default ":10255")
      --namespace string                      kubernetes namespace (default is 'all')
      --nodename string                       kubernetes node name (default "virtual-kubelet")
      --os string                             Operating System (Linux/Windows) (default "Linux")
      --pod-sync-workers int                  set the number of pod synchronization workers (default 10)
      --provider string                       cloud provider
      --provider-config string                cloud provider configuration file
      --trace-exporter strings                sets the tracing exporter to use, available exporters: [jaeger ocagent]
      --trace-sample-rate string              set probability of tracing samples
      --trace-service-name string             sets the name of the service used to register with the trace exporter (default "virtual-kubelet")
      --trace-tag map                         add tags to include with traces in key=value form
```

PR #599, opened 2019-05-03
Introduce Soft Deletes to allow for Terminal Status Updates

sargun:
This introduces the concept of soft deletes, so that the
virtual-kubelet can make a call to API server to update
the pod and set the terminal state.

The mechanism of action is that if soft deletes are enabled
(by the user), the kubelet will then not actually delete
the pod from apiserver, but instead set an annotation
indicating that it has been deleted. Then, on subsequent
updates, it will not make the callback to the provider
to repeat deleting the pod.

Although this is a user-facing option in this PR, it also
makes it so that when you initialize the pod controller, you
can state whether you want this behaviour or not.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,10 @@ jobs:
           command: |
             make e2e
       - run:
+          name: Run the end-to-end test suite
+          command: |
+            make e2e NODE_NAME=vkubelet-mock-1 STATS_PORT=10256
+      - run:
           name: Collect logs on failure from vkubelet-mock-0
           command: |
             kubectl logs vkubelet-mock-0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ jobs:
       CHANGE_MINIKUBE_NONE_USER: true
       GOPATH: /home/circleci/go
       KUBECONFIG: /home/circleci/.kube/config
-      KUBERNETES_VERSION: v1.12.3
+      KUBERNETES_VERSION: v1.13.4
       MINIKUBE_HOME: /home/circleci
       MINIKUBE_VERSION: v0.30.0
       MINIKUBE_WANTUPDATENOTIFICATION: false
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,14 @@ jobs:
               sleep 1;
             done
       - run:
+          name: Watch pods
+          command: kubectl get pods -o json --watch
+          background: true
+      - run:
+          name: Watch nodes
+          command: kubectl get nodes -o json --watch
+          background: true
+      - run:
           name: Run the end-to-end test suite
           command: |
             make e2e
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,9 +103,14 @@ jobs:
           command: |
             make e2e
       - run:
-          name: Collect logs on failure
+          name: Collect logs on failure from vkubelet-mock-0
           command: |
             kubectl logs vkubelet-mock-0
+          when: on_fail
+      - run:
+          name: Collect logs on failure from vkubelet-mock-1
+          command: |
+            kubectl logs vkubelet-mock-1
           when: on_fail
 
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,6 +102,11 @@ jobs:
           name: Run the end-to-end test suite
           command: |
             make e2e
+      - run:
+          name: Collect logs on failure
+          command: |
+            kubectl logs vkubelet-mock-0
+          when: on_fail
 
 workflows:
   version: 2
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -36,6 +36,8 @@ var (
 	taintValue string
 	// taintEffect is the effect of the taint that is expected to be associated with the virtual-kubelet node to test.
 	taintEffect string
+	// Wait for Kubelet pod to come up before running tests
+	waitForKubelet bool
 )
 
 func init() {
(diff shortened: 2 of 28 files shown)


vkubelet/pod.go
```diff
@@ -12,11 +13,14 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 )
 
+const softDeleteKey = "com.github.virtualKubelet.virtualKubelet.softDelete"
```
pires: As an annotation key, this format seems weird to me. Given there's the `virtual-kubelet.io` domain too, I suggest changing to `softdelete.pod.virtual-kubelet.io`.

vkubelet/pod.go
```diff
@@ -94,25 +98,94 @@ func (s *Server) deletePod(ctx context.Context, namespace, name string) error {
...
+		return err
+	}
+	// If the pod deletion was successful, or the provider did not know about the pod
+	// then go ahead and tombstone the pod, otherwise we allow the pod to persist in
+	// API server, and if it gets another update, we will retry deleting it
+	// from the provider.
+	//
+	// Reasoning: A provider may get into an inconsistent state, and a restart might
+	// be necessary go get it back into a good state. We would rather allow for graceful
+	// termination if possible
```
pires: So assuming the VK and/or provider restarted, where in the code exactly will the reconciliation happen?

vkubelet/pod.go
```diff
@@ -94,25 +98,94 @@ func (s *Server) deletePod(ctx context.Context, namespace, name string) error {
...
+}
+
+// terminatePod terminates the pod in the provider. It assumes the API
+// server pod has been marked for deletion
+func (s *Server) terminatePod(ctx context.Context, pod *corev1.Pod) error {
+	ctx, span := trace.StartSpan(ctx, "terminatePod")
+	defer span.End()
+	ctx = addPodAttributes(ctx, span, pod)
+	ctx = span.WithField(ctx, "softDeletes", s.softDeletes)
+
```
pires: I suggest adding some logging from this point onwards to increase visibility into what's going on.

vkubelet/pod.go
sargun: What about:
virtual-kubelet.io/softDelete?

That jives with: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

> Annotations are key/value pairs. Valid annotation keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/).

pires: Oh, right, I forgot the key part of the annotation 🤦‍♂

test/e2e/main_test.go
```diff
@@ -45,17 +47,20 @@ func init() {
 	flag.StringVar(&taintKey, "taint-key", defaultTaintKey, "the key of the taint that is expected to be associated with the virtual-kubelet node to test")
 	flag.StringVar(&taintValue, "taint-value", defaultTaintValue, "the value of the taint that is expected to be associated with the virtual-kubelet node to test")
 	flag.StringVar(&taintEffect, "taint-effect", defaultTaintEffect, "the effect of the taint that is expected to be associated with the virtual-kubelet node to test")
-	flag.Parse()
```
sargun: Running flag.Parse in init causes issues because other packages check for Parsed() before they add their flags (in their inits)

test/e2e/main_test.go
```diff
@@ -45,17 +47,20 @@ func init() {
...
 func TestMain(m *testing.M) {
+	flag.Parse()
 	// Set sane defaults in case no values (or empty ones) have been provided.
 	setDefaults()
 	// Create a new instance of the test framework targeting the specified node.
 	f = framework.NewTestingFramework(kubeconfig, namespace, nodeName, taintKey, taintValue, taintEffect)
-	// Wait for the virtual-kubelet pod to be ready.
-	if err := f.WaitUntilPodReady(namespace, nodeName); err != nil {
-		panic(err)
+	if waitForKubelet {
```
sargun: I can remove this from the PR if you want, it's just something that is useful when debugging outside of the cluster

.circleci/config.yml
```diff
@@ -102,6 +102,11 @@ jobs:
           name: Run the end-to-end test suite
           command: |
             make e2e
+      - run:
+          name: Collect logs on failure
+          command: |
+            kubectl logs vkubelet-mock-0
```
cpuguy83: 👍

vkubelet/pod.go
```diff
@@ -218,11 +303,64 @@ func (s *Server) updatePodStatus(ctx context.Context, pod *corev1.Pod) error {
...
+	// We don't really have a better option, as this method is only called from the async pod notifier,
+	// which provides (ordered)
+	pod.ObjectMeta.ResourceVersion = ""
+	patch, err := json.Marshal(pod)
+	if err != nil {
+		span.SetStatus(ocstatus.FromError(err))
+		return pkgerrors.Wrap(err, "Unable to serialize patch JSON")
+	}
+
+	newPod, err := s.k8sClient.CoreV1().Pods(pod.Namespace).Patch(pod.Name, types.MergePatchType, patch, "status")
```
sargun: This should restrict the patch from only applying to "status" subresource. I'm not 100% sure about this behaviour as the documentation on merge patch (non-strategic) + subresources seems to be...vague.

sargun:
PTAL:
This changed from a fairly simple hack to a pretty big hack. 

It touches a bunch of places, and I can break up the PR if y'all want:
* Changes the way that all of the queue / concurrency stuff works. In two major ways:
1. Before the pattern was to send strings across the queue, and then the receiver would than act on the string. This exposed: https://github.com/virtual-kubelet/virtual-kubelet/issues/602
2. Rather than sending strings, it sends closures.

* It adds a bunch of ergonomics stuff to the testing code.
* It changes Mock into an async provider
* It changes the functional tests to depend on the soft delete pattern
* It calls `PATCH` on both pods, and pod statuses, and this in turn makes it so that it's a newly added permission
* It gets the pod statuses from the notifier at the time of notification, and not calling back -- this has the one "cost" that the (async) provider must not unnecessarily callback without an updated pod status.

sargun:
Also, https://github.com/virtual-kubelet/virtual-kubelet/pull/595 is in here. I can remove that once that's merged. The only reason is that I use it in the test bits for logs.

cpuguy83:
As far as I'm aware, the way deletes work in k8s is the `DeletionTimestamp` gets set, which basically means "the pod will be listed at least until `pod.DeletionTimestamp`".

sargun:
superseded by https://github.com/virtual-kubelet/virtual-kubelet/pull/612

PR #606, opened 2019-05-07
Move to lowercase Sirupsen/logrus

sargun:


sargun:
MacOS does not like them being in the same commit.

sargun:
PTAL, @pires @cpuguy83

PR #611, opened 2019-05-09
Add the /runningpods/ api endpoint

sargun:
This adds an API endpoint from the kubelet (/runningpods/). It is
an endpoint on kubelet which is considered a "debug" endpoint, so
it might be worth exposing through the options, but by default
it is exposed in most k8s configs AFAICT.

### DIFF
--- a/vkubelet/api/pods.go
+++ b/vkubelet/api/pods.go
@@ -1,0 +1,53 @@
+package api
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/cpuguy83/strongerrors"
+	"github.com/virtual-kubelet/virtual-kubelet/log"
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+)
+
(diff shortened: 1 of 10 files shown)


vkubelet/api/pods.go
```diff
@@ -0,0 +1,53 @@
...
+	scheme := runtime.NewScheme()
+	v1.SchemeBuilder.AddToScheme(scheme)
+	codecs := serializer.NewCodecFactory(scheme)
+
+	return handleError(func(w http.ResponseWriter, req *http.Request) error {
+		ctx := req.Context()
+		ctx = log.WithLogger(ctx, log.L)
+		pods, err := p.GetPods(ctx)
+		if err != nil {
+			return strongerrors.System(err)
```
sargun: @cpuguy83 No idea if this is the right one to use.

cpuguy83: The provider should give the correct error rather than try to figure it out here.

sargun: Will handlError handle it correctly if they don't return a strongerror?

sargun:
@cpuguy83 any opinions?

PR #613, opened 2019-05-10
Light up UpdatePod

jeremyrickard:
This PR updates the vkublet/pod.go createOrUpdate(..) method to actually handle
updates. 

The createOrUpdatePod gets the Pod from the provider as before, but now if it exists the method
checks the hash of the spec against the spec of the new pod instead of simply returning. If the spec has changed, it then calls provider.UpdatePod(). Associated test added that makes a new Test Server, Makes a Mock Provider and uses those to test both create and update branches. 

Note: no providers implement UpdatePod currently, so there isn't much change here. 

This PR also makes a change to the Server struct to swap from kuberentes.Clientset to kubernetes.Interface to better facilitate testing with fake ClientSet.


providers/mock/mock.go
```diff
@@ -71,6 +71,32 @@ func NewMockProvider(providerConfig, nodeName, operatingSystem string, internalI
 	return &provider, nil
 }
 
+// NewMockProviderMockConfig creates a new MockProvider with the given Mock Config
```
jeremyrickard: Added this just to simplify making a new provider for testing without needing to make files and deal with file io)

vkubelet/pod.go
```diff
@@ -28,60 +30,87 @@ func addPodAttributes(ctx context.Context, span trace.Span, pod *corev1.Pod) con
...
+	// NOTE: Some providers return a non-nil error in their GetPod implementation when the pod is not found while some other don't.
+	// Hence, we ignore the error and just act upon the pod if it is non-nil (meaning that the provider still knows about the pod).
+	if pp, _ := s.provider.GetPod(ctx, pod.Namespace, pod.Name); pp != nil {
+		// Pod Update Only Permits update of:
+		// - `spec.containers[*].image`
+		// - `spec.initContainers[*].image`
+		// - `spec.activeDeadlineSeconds`
+		// - `spec.tolerations` (only additions to existing tolerations)
+		// compare the hashes of the pod specs to see if the specs actually changed
+		expected := hashPodSpec(pp.Spec)
```
sargun: Couldn't this get into a weird situation where if the provider does something like async updates that the update function is called, but not applied immediately? Shouldn't we try to dedup the updates (i.e. keep track of the last update we did?)

jeremyrickard: I'm not sure that state keeping makes sense in the Kubernetes world. The API server is the source of truth / keeper of state. If we get an update and the spec has changed, that should signify that another update needs to be made by the provider. This is essentially the same behavior that occurs in a regular kubelet, except at a little different granularity (i.e. the kubelet diffs each container within the update). In the case of updating a pod, it's actually quite limited in what you can change: image (the containers and the init containers), tolerations, and active deadline. I think the more common thing to occur would be something like with a replica set or a deployment, which would result in pod create/deletes and not updates. Adding de-dupe on these seems like adding complexity to VK for a possibly infrequently used case. Providers might implement that logic if it makes sense. 

@cpuguy83 any thoughts?

cpuguy83: Agree that we don't want to keep state like this.
The reason being is that the k8s API is a declarative API, if a pod state updates 5x before we ever apply any changes, we should only apply the last state as it is the current expected state.
This is why the queue is setup to prefer storing keys to storing point-in-time objects.

vkubelet/pod.go
```diff
@@ -28,60 +30,87 @@ func addPodAttributes(ctx context.Context, span trace.Span, pod *corev1.Pod) con
...
+		if actual := hashPodSpec(pod.Spec); actual != expected {
+			log.G(ctx).Info("Pod %s exists, updating pod in provider", pp.Name)
```
cpuguy83:
```suggestion
			log.G(ctx).Debugf("Pod %s exists, updating pod in provider", pp.Name)
```

PR #617, opened 2019-05-14
Fix permission denied warnings during tests

sargun:
For some reason the kubelet tries to do things that it gets perm
denied. This changes the base.yml in the skaffold file to have
the permissions. In turn, this broke the node deletion test. So,
instead of looking for the deletion event (imperative), we wait
for the node UID to change (declarative).

### DIFF
--- a/test/e2e/node_test.go
+++ b/test/e2e/node_test.go
@@ -3,49 +3,61 @@
 package e2e
 
 import (
+	"context"
 	"testing"
 	"time"
 
 	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	watchapi "k8s.io/apimachinery/pkg/watch"
 )
 
 // TestNodeCreateAfterDelete makes sure that a node is automatically recreated
 // if it is deleted while VK is running.
 func TestNodeCreateAfterDelete(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	podList, err := f.KubeClient.CoreV1().Pods(f.Namespace).List(metav1.ListOptions{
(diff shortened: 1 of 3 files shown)


sargun:
@cpuguy83 This should be ready to review.

test/e2e/node_test.go
```diff
@@ -3,36 +3,56 @@
...
+
+	podList, err := f.KubeClient.CoreV1().Pods(f.Namespace).List(metav1.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector("spec.nodeName", f.NodeName).String(),
+	})
+
+	assert.NilError(t, err)
+	assert.Assert(t, is.Len(podList.Items, 0), "Kubernetes does not allow node deletion with dependent objects (pods) in existence: %v", podList.Items)
+
 	chErr := make(chan error, 1)
 	chDone := make(chan struct{})
```
cpuguy83: delete

test/e2e/node_test.go
```diff
@@ -3,36 +3,56 @@
...
 			select {
-			case <-chDone:
+			// The test has exited, go ahead and exit
+			case <-ctx.Done():
 				return true, nil
 			default:
 			}
 
-			if deleted {
-				return f.WaitUntilNodeAdded(e)
```
cpuguy83: Can this function by removed?

test/e2e/node_test.go
```diff
@@ -44,8 +64,10 @@ func TestNodeCreateAfterDelete(t *testing.T) {
 
 	select {
 	case <-timer.C:
+		t.Logf("Observed deletion: %v", deleted)
```
cpuguy83: Are these logs necessary since this is the end of the test anyway?

test/e2e/node_test.go
sargun: Isn't this worth leaving in? If the node for whatever reason has a dependent object, this test wont pass. (if any of the above tests get into a weird state)

cpuguy83: As far as I can tell it's not being used.

sargun:
@cpuguy83 PTAL.

PR #618, opened 2019-05-14
Remove unused Azure scripts

cpuguy83:
This were previously used for setting up azure creds for ACI specific
tests, but are no longer needed since that has moved to a separate repo.


cpuguy83:
Self-merging. If there is an issue we can revert.
Note: I have deleted the env settings from the Circle job already.

PR #627, opened 2019-05-17
Convert mock provider to async provider

sargun:


sargun:
@cpuguy83 Are you sure the async podnotifier interface works?

sargun:
I saw the two dependent changes get merged in. If there are any other changes needed, I should be able to get to this tomorrow.

PR #628, opened 2019-05-17
Use Running Pods Endpoint for testing

sargun:
* Use the VKUBELET API for fetching the pods during testing, which
   makes it easier to run testing without wiring up an entire cluster
 * Runningpods does not require each provider to implement the
   optional stats interface for testing.
  * This also splits out some of the tests (Disentangles them)
  * This also changes it to use the test name in the pod name.

### DIFF
--- a/test/e2e/basic_test.go
+++ b/test/e2e/basic_test.go
@@ -8,33 +8,35 @@ import (
 	"time"
 
 	"github.com/virtual-kubelet/virtual-kubelet/vkubelet"
-	v1 "k8s.io/api/core/v1"
+	"gotest.tools/assert"
+	"k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
 )
 
 const (
-	// deleteGracePeriodForProvider is the amount of time we allow for the provider to react to deletion of a pod before proceeding to assert that the pod has been deleted.
-	deleteGracePeriodForProvider = 100 * time.Millisecond
+	// deleteGracePeriodForProvider is the maximum amount of time we allow for the provider to react to deletion of a pod
+	// before proceeding to assert that the pod has been deleted.
+	deleteGracePeriodForProvider = 1 * time.Second
 )
 
 // TestGetStatsSummary creates a pod having two containers and queries the /stats/summary endpoint of the virtual-kubelet.
 // It expects this endpoint to return stats for the current node, as well as for the aforementioned pod and each of its two containers.
 func TestGetStatsSummary(t *testing.T) {
-	// Create a pod with prefix "nginx-0-" having three containers.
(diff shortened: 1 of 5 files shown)


sargun:
@jeremyrickard @cpuguy83 What do you think about this? I can further split it up if you want, but it makes it so we don't rely upon the backend implementing the stats interface for testing.

test/e2e/basic_test.go
```diff
@@ -334,3 +351,14 @@ func findPodInPodStats(summary *v1alpha1.Summary, pod *v1.Pod) (int, error) {
 	}
 	return -1, fmt.Errorf("failed to find pod \"%s/%s\" in the slice of pod stats", pod.Namespace, pod.Name)
 }
+
+// findPodInPodStats returns the index of the specified pod in the .pods field of the specified PodList object.
+// It returns error if the pod doesn't exist in the podlist
+func findPodInPods(pods *v1.PodList, pod *v1.Pod) error {
```
cpuguy83: Nit: You can return a bool here instead of an error and simplify the asserts based on this.
It also seem like when we get the pod list all we ever want to do is check if the pod we have is running.
Maybe this can just be part of the framework? e.g. `assert.Assert(t, f.PodIsRunning(pod))`

test/e2e/basic_test.go
```diff
@@ -49,165 +51,188 @@ func TestGetStatsSummary(t *testing.T) {
...
+	// Forcibly delete the pod.
+	if err := f.DeletePodImmediately(pod.Namespace, pod.Name); err != nil {
+		t.Logf("Last saw pod in state: %+v", podLast)
 		t.Fatal(err)
 	}
+	t.Log("Force deleted pod: ", pod.Name)
+
 	// Wait for the delete event to be ACKed.
-	if err := <-pod0Ch; err != nil {
+	if err := <-podCh; err != nil {
```
cpuguy83: Select with timeout?
I haven't looked at `WaitUntilPodDeleted` but it would be nice if that took a cancellable context.

test/e2e/basic_test.go
```diff
@@ -49,165 +51,188 @@ func TestGetStatsSummary(t *testing.T) {
...
 	go func() {
-		// Wait for the "nginx-0-X" pod to be reported as having been deleted.
-		if err := f.WaitUntilPodDeleted(pod0.Namespace, pod0.Name); err != nil {
+		// Close the podCh channel, signaling we've observed deletion of the pod.
+		defer close(podCh)
+
+		var err error
+		// Wait for the pod to be reported as having been deleted.
+		podLast, err = f.WaitUntilPodDeleted(pod.Namespace, pod.Name)
+		if err != nil {
```
cpuguy83: Why not always send `err`?

test/e2e/basic_test.go
sargun: I tried to keep this to the semantics of findPodInPodStats. If you're fine with the change, I'm all for it.

cpuguy83: Yeah since we're changing I think it's good.

cpuguy83:
Comments can be handled separately. This is a good change.

PR #629, opened 2019-05-18
Remove `Server` object

cpuguy83:
This had some weird shared responsibility with the PodController.
Instead just move the functionality to the PodController.


cpuguy83:
ping @sargun

cpuguy83:
Note that this no longer has the `InSync` call since the caller can wait on the cache to be in sync.

cpuguy83:
I was considering adding another object (but as a separate PR)... probably called `Agent`... which basically manages the `PodController` and `Node`... so basically a bunch of the work that the CLI currently does.

vkubelet/podcontroller.go
```diff
@@ -22,21 +22,22 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cpuguy83/strongerrors"
+
```
sargun: Why the newline?

cpuguy83: Took did this for whatever reason

cpuguy83: *Tool

cpuguy83: *Tool

cpuguy83:
Ok, now this actually works.

PR #630, opened 2019-05-18
Upgrade to go1.12

sargun:


manager/resource.go
```diff
@@ -1,7 +1,7 @@
 package manager
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
pires: This renders already to `v1` so why the redundant alias? I suggest changing it to `corev1`, which can be found in this repo too, sometimes.

I also suggest we have a separate PR that normalizes this across the entire code base.

cpuguy83: goimports often does this.

sargun: This came due to upgrading from go1.10 to go1.12 goimports. Go imports changed their style in 1.11 IIRC where in some cases it makes the alias explicit versus implicit.

sargun: @pires I wanted to make sure that the version of Go and the Go coding style were upgraded in line.

pires: OK I didn't know about the behavior you describe. It's OK then.

PR #633, opened 2019-05-20
Handle not found case for status update

cpuguy83:
If we don't handle the "pod not found" case then we end up with the pod
getting re-queued over and over until the max retries are hit. It also
blocks the queue for other pod status updates for that pod
namespace/name.

### DIFF
--- a/vkubelet/pod.go
+++ b/vkubelet/pod.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	pkgerrors "github.com/pkg/errors"
+	"github.com/virtual-kubelet/virtual-kubelet/errdefs"
 	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"github.com/virtual-kubelet/virtual-kubelet/trace"
 	corev1 "k8s.io/api/core/v1"
@@ -206,7 +207,7 @@ func (pc *PodController) updatePodStatus(ctx context.Context, pod *corev1.Pod) e
 	ctx = addPodAttributes(ctx, span, pod)
 
 	status, err := pc.provider.GetPodStatus(ctx, pod.Namespace, pod.Name)
-	if err != nil {
+	if err != nil && !errdefs.IsNotFound(err) {
 		span.SetStatus(err)
 		return pkgerrors.Wrap(err, "error retreiving pod status")
 	}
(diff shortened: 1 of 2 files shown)


cpuguy83:
I wonder if the pod status queue should be using the pod UID instead of namespace/name.

vkubelet/pod.go
```diff
@@ -204,7 +205,7 @@ func (s *Server) updatePodStatus(ctx context.Context, pod *corev1.Pod) error {
 	ctx = addPodAttributes(ctx, span, pod)
 
 	status, err := s.provider.GetPodStatus(ctx, pod.Namespace, pod.Name)
-	if err != nil {
+	if err != nil && !strongerrors.IsNotFound(err) {
```
sargun: Can you add to the docs somewhere that `GetPod` / `GetPodStatus` should return a `NotFoundError`?

cpuguy83: See #635

sargun:
How can you guarantee pod names are unique? Can't you delete a pod, and recreate it with the same name?

pires:
Pod names are unique. You can't have two pods with the same name in the API. Now, it may be possible for multiple different instances of the VK to have each their instance of a pod with the same name, but that only means the _illegal_ VK instance(s) are handling stale data.

sargun:
@pires Kind of. If you have a network partition, a pod can be deleted and recreated with the same name. Although this may seem weird, it can happen if people are using pod names a kind of "singleton" key.

sargun:
Again, it's okay for VK1.0 to use namespace/name.

vkubelet/queue.go
```diff
@@ -55,7 +55,7 @@ func handleQueueItem(ctx context.Context, q workqueue.RateLimitingInterface, han
 		if err := handler(ctx, key); err != nil {
 			if q.NumRequeues(key) < maxRetries {
 				// Put the item back on the work queue to handle any transient errors.
-				log.G(ctx).Warnf("requeuing %q due to failed sync: %v", key, err)
+				log.G(ctx).WithError(err).Warnf("requeuing %q due to failed sync", key)
```
jeremyrickard: 👍

pires:
I don't think Kubernetes has core primitives that prevent a scenario where certain kubelets (or any other controllers) stop seeing an(y) API server and continue to run workloads that the kube-scheduler will event reschedule to other nodes (visible from an API perspective).
Don't forget API servers are stateless. They don't care about network partitions. For as long as an API server has access to an healthy etcd, it has visibility into _everything_ that is stored there, including (un)healthy nodes.

PR #634, opened 2019-05-20
Use custom mock in vkubelet tests

### LINKED ISSUE #632: Dependency hell / import loop from mock to vkubelet package
If we want to use any of the interfaces from the vkubelet in mock (for example, if you want to assert that you implement the PodNotifier interface), you cannot, because pod_test in vkubelet uses the mock provider to test.

This is due to a combination of:
* https://github.com/virtual-kubelet/virtual-kubelet/pull/626
* https://github.com/virtual-kubelet/virtual-kubelet/pull/613

It'd be nice if either pod_test wasn't in the vkubelet package, the pod notifier interface was moved out of the vkubelet package.

cpuguy83:
Don't use actual "mock" provider in vkubelet to reduce awkward import
cycles.
Really the "mock" provider is for e2e tests. Instead create our own mock
in the vkubelet package specifically for unit tests which can also be
more easily customized to test edge cases.

Closes #632

### DIFF
--- a/vkubelet/pod_test.go
+++ b/vkubelet/pod_test.go
@@ -2,9 +2,11 @@ package vkubelet
 
 import (
 	"context"
+	"path"
 	"testing"
 
-	"github.com/virtual-kubelet/virtual-kubelet/providers/mock"
+	"github.com/cpuguy83/strongerrors"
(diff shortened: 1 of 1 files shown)


cpuguy83:
ping @sargun

PR #635, opened 2019-05-20
Define errdefs package

cpuguy83:
Posting this to get people's thoughts before I go too far down the rabbit hole of implementing error types.

---

Creates an `errdefs` package which holds all the error definitions. Providers should use these error definitions in order for the vk core to understand error values.


errdefs/notfound.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"errors"
+	"fmt"
+)
+
+// NotFound is an error interface which denotes whether the opration failed due
+// to a the resource not being found.
+type ErrNotFound interface {
+	NotFound() bool
```
cpuguy83: Note that this is a little different than github.com/cpuguy83/strongerrors which only defines such interfaces like so:

```go
type ErrNotFound interface {
  NotFound()
}
```

I haven't personally decided if I like it more, less or the same... probably about the same.

cpuguy83: The main advantage here is error implementations can short circuit the error traversal.

rbitia:
@cpuguy83 are you waiting on a review?

cpuguy83:
Yep.

errdefs/invalid.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (e *invalidInputError) Error() string {
+	return e.cause.Error()
+}
+
+func (e *invalidInputError) InvalidInput() bool {
+	return true
+}
+
+func (e *invalidInputError) Cause() error {
```
sargun: This blows the stack because it's infinitely recursive AFAICT:
https://play.golang.org/p/n9J-YPfZpcM

errdefs/invalid.go
```diff
@@ -0,0 +1,69 @@
...
+
+// InvalidInput is an error interface which denotes whether the opration failed due
+// to a the resource not being found.
+type ErrInvalidInput interface {
+	InvalidInput() bool
+	error
+}
+
+type invalidInputError struct {
+	cause error
```
sargun: Any reason to not just leave this as an embedding, so Error() doesn't have to be reimplemented?

errdefs/invalid.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (e *invalidInputError) InvalidInput() bool {
+	return true
+}
+
+func (e *invalidInputError) Cause() error {
+	return e.Cause()
+}
+
+// AsInvalidInput wraps the passed in error to makeit of type ErrInvalidInput
```
sargun: "makeit" -> "make it"

errdefs/invalid.go
```diff
@@ -0,0 +1,69 @@
...
+//
+// Callers should make sure the passed in error has exactly they error message
```
sargun: they -> the.

errdefs/wrapped.go
```diff
@@ -0,0 +1,10 @@
+package errdefs
+
+// Causal is an error interface for errors which have wrapped another error
+// in a non-opaque way.
+//
+// This pattern is used by github.com/pkg/errors
+type Causal interface {
```
sargun: Given pkg/errors is going to be standardized in go1.13, should we do our own, if it's going to be "obsoleted" in a few months? 
See: https://github.com/golang/go/issues/29934

errdefs/notfound.go
```diff
@@ -0,0 +1,69 @@
...
+// which implements the `NotFound` interface.
+func IsNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
+	if e, ok := err.(ErrNotFound); ok {
+		return e.NotFound()
+	}
+
+	if e, ok := err.(Causal); ok {
```
sargun: Refer to other comment about pkg/errors.

sargun: Also, why should Casual be a public interface?

providers/aws/fargate/cluster.go
```diff
@@ -276,7 +276,7 @@ func (c *Cluster) GetPod(namespace string, name string) (*Pod, error) {
 	tag := buildTaskDefinitionTag(c.name, namespace, name)
 	pod, ok := c.pods[tag]
 	if !ok {
-		return nil, strongerrors.NotFound(fmt.Errorf("pod %s/%s is not found", namespace, name))
+		return nil, errdefs.NotFoundf("pod %s/%s is not found", namespace, name)
```
sargun: It looks like we use both `namespace/pod`, and `namespace-pod` in error messages. Any reason not to standardize on `namespace/pod`?

providers/azure/aci.go
```diff
@@ -777,7 +777,7 @@ func (p *ACIProvider) GetContainerLogs(ctx context.Context, namespace, podName,
 	}
 
 	if cg.Tags["NodeName"] != p.nodeName {
-		return nil, strongerrors.NotFound(errors.New("got unexpected pod node name"))
+		return nil, errdefs.NotFound("got unexpected pod node name")
```
sargun: nit:
"got pod with unexpected node name" reads more clearly.

trace/opencensus/opencensus.go
```diff
@@ -46,7 +47,30 @@ func (s *span) End() {
...
+
+	if err == nil {
+		status.Code = octrace.StatusCodeOK
+		s.s.SetStatus(status)
+		return
+	}
+
+	switch {
+	case errdefs.IsNotFound(err):
+		status.Code = octrace.StatusCodeNotFound
```
sargun: Does it make more sense to have an (anonymous) interface a la:

```
type ocstatusError interface {
    OCStatus() octrace.Status
}
```
And then errdefs can just implement it?

trace/opencensus/opencensus.go
```diff
@@ -46,7 +47,30 @@ func (s *span) End() {
...
+	case errdefs.IsInvalidInput(err):
+		status.Code = octrace.StatusCodeInvalidArgument
+		// TODO: other error types
```
sargun: Which other error types?

errdefs/wrapped.go
cpuguy83: pkg/errors is a package used all over the place.
Having a solution in go (or really it's currently in golang.org/x/exp/errors) does not make the usage of pkg/errors go away.
Once the interface for x/exp/errors is finalized we should definitely support it here.

errdefs/invalid.go
cpuguy83: I want to be able to implement the causal interface, embedding `error` I don't think we'd be able to get at the underlying error type.

errdefs/invalid.go
cpuguy83: Oops :)

cpuguy83: Clearly missing a test there. Will fix.

providers/azure/aci.go
cpuguy83: Let's handle changes to messages separately, if you don't mind.

errdefs/notfound.go
cpuguy83: I think I made it public for documentation purposes.

cpuguy83: But it may not be necessary.

errdefs/invalid.go
sargun:
```
package main

import (
	"fmt"
	"errors"
)

func main() {
	var e error = &invalidInputError{errors.New("Sargun's error")}
	fmt.Println(e)
	causalError, ok := e.(Causal)
	fmt.Println(ok)
	fmt.Println(causalError.Cause())
}

type Causal interface {
	Cause() error
	error
}


type ErrInvalidInput interface {
	InvalidInput() bool
	error
}

type invalidInputError struct {
	error
}

func (e *invalidInputError) Cause() error {
	return e.error
}
```

Implements ^ the error interface, and don't have to restate Error. Also, can fetch the underlying error.

trace/opencensus/opencensus.go
cpuguy83: This ties errors to the opencensus implementation which is not great.
Since we already know the kind of error and can do the translation, this seems a bit heavy to add.

errdefs/notfound.go
sargun: It's awkward right -- if someone wants to write their own wrap methods which may eventually drill-down to a errdefs error, they want to know how to implement this interface. Do we want people doing that?  

The route that pkgerrors goes is that it defines the interface in the docstring, which I think is good enough.

errdefs/wrapped.go
sargun: I think the comment was misunderstood. Rather than defining our own causal interface, I was going to suggest using the causal interface from `pkg/errors`.

errdefs/invalid.go
cpuguy83: 🐑 
Wow yes, not sure what I was thinking.

errdefs/wrapped.go
cpuguy83: Oh... yes... you mean use `errors.Cause`
I think I didn't want to take a dependency on pkg/errors in this package for such a small thing.

pires: I don't see anything wrong with introducing said dependency if it's something we're doing down the line.

errdefs/notfound.go
sargun: @cpuguy83 Can you give an example of where short-circuiting / overriding error traversal would be valuable?

sargun:
lgtm.

errdefs/notfound.go
cpuguy83: So when we do something like:

```go
select {
case IsNotFound(err):
case IsInvalidInput(err):
case IsSomethingElse(err):
}
```

For each case (until it finds a match, of course), it will traverse the entire causal chain.

But if the err I pass in implements not found where it just returns `false`, it doeesn't need to traverse that chain.

trace/opencensus/opencensus.go
cpuguy83: Could be things like `Conflict`, etc.

pires:
@cpuguy83 @sargun I can't follow this discussion right now in a way I can meaningfully contribute. I think you are going deep and no decision will be made on a whim, so I trust you to make the better call here. I am therefore excluding myself from the reviewer list. Thank you for trusting, nonetheless 🙇

cpuguy83:
Thanks!

PR #638, opened 2019-05-22
env: fix env order error due to save env to map

shidao-ytt:
there are circular reference of env in k8s. like a=1, b=$(a),
if the env order is error, like b=$(a), a=1, the env b can't be
resolved correct. use env slice replace with map to keep env order
correct

### DIFF
--- a/vkubelet/env.go
+++ b/vkubelet/env.go
@@ -92,7 +92,7 @@ func populateContainerEnvironment(ctx context.Context, pod *corev1.Pod, containe
 	// This is in accordance with what the Kubelet itself does.
 	// https://github.com/kubernetes/kubernetes/blob/v1.13.1/pkg/kubelet/kubelet_pods.go#L557-L558
 	container.EnvFrom = []corev1.EnvFromSource{}
-	container.Env = mergeEnvironments(envFrom, env)
+	container.Env = append(envFrom, env...)
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


cpuguy83:
```
--- FAIL: TestEnvOverridesEnvFrom (0.10s)
    env_internal_test.go:692: assertion failed: 
        --- pod.Spec.Containers[0].Env
        +++ →
        Sort({[]v1.EnvVar})[1->?]:
        	-: v1.EnvVar{Name: "FOO", Value: "__foo__"}
        	+: <non-existent>
        
```

tty5:
the test case fail reason is 
when save envs in map, the new env object will replace the old env.
and now, change to use slice to save envs, the new env and old env are all saved in pod.Spec.Containers[0].env

in k8s pod Spec, there are two env has same name is valid, so i would like to modify the test case to correct.

how about u think?

yashdesai93:
In the current implementation  we maintain order of env by order of traversal of `container.Env`. So an env var is dependent on a later variable, it will be unresolved. I don't see what's the problem.

yashdesai93:
If you look at https://github.com/virtual-kubelet/virtual-kubelet/pull/643 it takes care of env var composed of existing env vars.

tty5:
> If you look at #643 it takes care of env var composed of existing env vars.

the difference between this PR and #643 is whether expanse envs in virtual kubelet

an other choice is transfer the original envs to provider. in this condition, need keep the order of envs. if not, the env expansion will be wrong

cpuguy83:
Closing since this seems to have gone stale and never passed tests.

PR #639, opened 2019-05-22
Bugfix: Azure ACI provider memory limits round

jahstreet:
Description:
Azure ACI memory limits have to be of a multiplier of 0.1G, just like for memory requests. I propose to calculate memory limits like we do it for memory requests.

This PR fixes errors when user sets memory limits to a specific value which is not a multiplier of 0.1G and allows user to set memory limits in power-of-two format, eg: Mi, Gi, etc.

A point to discuss here maybe and idea to ceil (round up) memory requests and limits to the 0.1G multiplied value. At present those values are rounded down which causes to create a Container in ACI with a memory value lower then user may want to request.

Signed-off-by: Aliaksandr Sasnouskikh <jahstreetlove@gmail.com>

### DIFF
--- a/providers/azure/aci.go
+++ b/providers/azure/aci.go
@@ -1220,11 +1220,11 @@ func (p *ACIProvider) getContainers(pod *v1.Pod) ([]aci.Container, error) {
 				cpuLimit = float64(container.Resources.Limits.Cpu().MilliValue()) / 1000.00
 			}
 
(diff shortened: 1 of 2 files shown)


srrengar:
@mostafaez can you take a look

cpuguy83:
Carried this in virtual-kubelet/azure-aci#6

Issue #641, opened 2019-05-22
virtual-kubelet has different DNS resolution behaviour to standard kubelet  


grahamhayes:
---

### Environment summary

Provider (e.g. ACI, AWS Fargate)
ACI
Version (e.g. 0.1, 0.2-beta)
0.9.0
K8s Master Info (e.g. AKS, ACS, Bare Metal, EKS)
AKS
Install Method (e.g. Helm Chart, )
All
### Issue Details
When creating a pod, I expect `/etc/resolv.conf` to be populated like a standard AKS kubelet (or plain upstream kubernetes).

By default, the search domain is left blank, and while it can be overridden by a dnsConfig section on the pod spec, it means that off the shelf software that generates its own pod specs cannot use the DNS based service discovery it relies on.

### Repo Steps

```
apiVersion: v1
kind: Pod
metadata:
  name: helloworld-3
spec:
  containers:
  - image: microsoft/aci-helloworld
    imagePullPolicy: Always
    name: helloworld
    resources:
      requests:
        memory: 1G
        cpu: 1
    ports:
    - containerPort: 80
      name: http
      protocol: TCP
    - containerPort: 443
      name: https
  nodeSelector:
    kubernetes.io/role: agent
    beta.kubernetes.io/os: linux
    type: virtual-kubelet
  tolerations:
  - key: virtual-kubelet.io/provider
    operator: Exists
  - key: azure.com/aci
    effect: NoSchedule
```

Then (using the ACI interface on the portal) 

`cat /etc/resolv.conf` or try to connect to a service via the cluster dns name with no dns suffixes.

Romiko:
Any update on this being resolved? Excuse the pun.

dkkapur:
Redirecting this to @macolso @MikkelHegn for latest updates!

PR #642, opened 2019-05-22
Allow virtual-kubelet to use cluster domain

grahamhayes:
This allows `--cluster-domain` to be passed to virtual kubelet like a
traditional kublet, and use this to generate search-domains for
`/etc/resolv.conf`

* Set default `cluster-domain` to `cluster-local` to match current kubelet
* Added an example usage for the Azure provider
* * Only apply to pods with `DNSClusterFirst` to match kubelet
* * Merge search-domains with any set in the `dnsConfig`
* * Set `ndots` to the default 5

Related: #641

Signed-off-by: Graham Hayes <gr@ham.ie>

### DIFF
--- a/cmd/virtual-kubelet/commands/root/flag.go
+++ b/cmd/virtual-kubelet/commands/root/flag.go
@@ -60,6 +60,7 @@ func (mv mapVar) Type() string {
 func installFlags(flags *pflag.FlagSet, c *Opts) {
 	flags.StringVar(&c.KubeConfigPath, "kubeconfig", c.KubeConfigPath, "kube config file to use for connecting to the Kubernetes API server")
 	flags.StringVar(&c.KubeNamespace, "namespace", c.KubeNamespace, "kubernetes namespace (default is 'all')")
+	flags.StringVar(&c.KubeClusterDomain, "cluster-domain", c.KubeClusterDomain, "kubernetes cluster-domain (default is 'cluster.local')")
 	flags.StringVar(&c.NodeName, "nodename", c.NodeName, "kubernetes node name")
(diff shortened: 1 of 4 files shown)


grahamhayes:
Using https://cloud.docker.com/repository/docker/grahamhayes/virtual-kubelet + the helm chart for AKS in the repo this was running, and generated the right `/etc/resolv.conf` when dealing with a mix of DNS config options.

cpuguy83:
It makes sense to have such an option, but we are working to remove providers from this repo and expect all providers to build their own daemon implementations.

grahamhayes:
OK - that makes sense - but for now this could still merge (for the config option at the root), and the code in there for ACI could merge here and be moved out to the external repo? 
Or I can implement the config option here, and if you point me to the ACI external repo I can add it there - whichever makes it easiest.

pires:
@cpuguy83 I am OK with whatever you decide here. Neither option strikes me as an obstacle for the (near) future changes.

cpuguy83:
Let's move this over to https://github.com/virtual-kubelet/azure-aci since this code is no longer here.

The tricky part here is having VK provide a common CLI implementation that all the providers can use.
It actually may not be that bad now since all the providers are now gone and provider registration is now done explicitly in `package main` instead of burried in a `func init()` in the library code.

Thanks!

cpuguy83:
Actually if you want to update this to just add the CLI/config bits we can try importing the CLI code as a library in azure-aci instead of how it currently is (copied in).

cpuguy83:
I've updated the azure-aci repo to use the vk CLI as a lib so if we add the flag here we can just update to take advantage of it there.

grahamhayes:
@cpuguy83 updated to just add the flag and the config option

grahamhayes:
virtual-kubelet/azure-aci#8 is the Azure portion of this

Issue #321, opened 2018-08-15
VK Node will be removed if VK Pod is unable to connect to API server


robbiezhang:
---

### Environment summary

Provider (e.g. ACI, AWS Fargate, Hyper)
ACI
Version (e.g. 0.1, 0.2-beta)
0.4.1
K8s Master Info (e.g. AKS, ACS, Bare Metal, EKS)
AKS
Install Method (e.g. Helm Chart, )
Helm Chart
### Issue Details
Since the VK node is a virtual node, which means it’s not in the cloud provider control plane, the k8s node_controller will delete it from the apiserver, 40 seconds (depending on the controller setting) after the VK pod report its status.
https://github.com/kubernetes/kubernetes/blob/337dfe0a9cde3894eb6a26f9184df659d54007c6/pkg/controller/cloud/node_controller.go#L285

rbitia:
#427

pires:
I'll look into this.

pires:
By looking at the code linked above (`node_controller.go`), this seems to happen only when:
- cloud-provider feature is enabled;
-  a node, or in this case the VK, reports its `node.status.conditions["Ready"]="false"`.

I'll talk to some people in sig/cloud-provider and propose the controller skips based on some annotation, e.g. `cloud.kubernetes.io/unmanaged`.

pires:
I've talked to one co-leads for sig-cloud-provider and we've agreed for me to come up with a proposal PR following https://github.com/kubernetes/kubernetes/pull/70344.

pires:
Here's a tentative solution https://github.com/kubernetes/kubernetes/pull/73171.

rbitia:
Hey @pires do you need some air cover to help push the upstream PR. If so we can have Lachie from our side help out.

andrewsykim:
After having more time to think through the discussion in https://github.com/kubernetes/kubernetes/pull/73171, I think a more reasonable solution is to embed a virtual kubelet ID into the node's provider ID. Maybe something like `aws://us-east-1a//virtual-kubelet-<fargate-id>`. Then as part of [EnsureInstanceExistsByProviderID](https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L168), the AWS cloud provider  
 can return `false` if the provider ID matches the format for virtual kubelet. This way we don't have to leak information about the virtual kubelet upstream but each provider can opt into this option if they want to support it. Thoughts @pires @rbitia?

[...]

PR #648, opened 2019-05-30
Support error handler callback for node status

cpuguy83:
This moves the logic for re-creating the a missing node up into the CLI.
We can make this optional, but for now I've just preserved existing
functionality.


cmd/virtual-kubelet/commands/root/root.go
```diff
@@ -143,6 +144,21 @@ func runRootCommand(ctx context.Context, c Opts) error {
...
+			log.G(ctx).Debug("node not found")
+			newNode := pNode.DeepCopy()
+			newNode.ResourceVersion = ""
+			_, err = client.CoreV1().Nodes().Create(newNode)
+			if err != nil {
+				return err
+			}
+			log.G(ctx).Debug("created new node")
+			return nil
+		}),
```
pires: Interesting thought, this may have an impact on #321.

PR #534, opened 2019-03-05
FieldRef feature for DownwardAPI

VineethReddy02:
Added FieldRef to VK, as the part of DownwardAPI

PR #650, opened 2019-06-04
Introduce PodInfoProvider interface to expose podinfo.

yashdesai93:
Downward API allows us to expose the hostIP, podIP and
node's allocatable resources using environment variables.
This info is not known to vkubelet, but only the provider,
and that too not before the `provider.CreatePod` request has been made.

Proposal:
1. Introduce PodInfoProvider interface to expose pod related info:

// PodInfoProvider is an optional interface that providers can implement to expose pod info.
type PodInfoProvider interface {
   // GetAllocatable returns the allocatable resource list.
   GetAllocatable() v1.ResourceList

   // GetPodIP returns the pod's IP.
   GetPodIP(pod *v1.Pod) string

   // GetHostIP returns the host's IP.
   GetHostIP(pod *v1.Pod) string
}

2. Make vkubelet/env.go's code a different lib/pkg and move
   the call for `populateEnvironmentVariables` from vkubelet/pod.go to within
   the provider's `CreatePod`.

3. Use the `PodInfoProvider` interface within the new env pkg, to figure out pod related info:

func PopulateEnvironmentVariables(ctx context.Context,
                                  pod *corev1.Pod,
                                  rm *manager.ResourceManager,
                                  recorder record.EventRecorder,
                                  podInfo providers.PodInfoProvider) error

Reasoning:
Moving envvar processing into a separate pkg, makes the code more modular to begin with.

Vkubelet layer (pod.go) does not have enough info: podIP, hostIP, or
allocatable resources during the current population step and there
is no way the provider has this info before the `CreatePod` request.

Another solution could be to let things be and let provider handle
the remaining downward API, but there are issues with that approach:
If we try to split up populating within vkubelet and provider, then we don't
allow for reuse of a lot of the code within vkubelet/env.go.
Also composing env vars within vkubelet becomes an issue, since
composing should be done only after all the previous env vars have
been mapped:

env     - name: MY_POD_IP
          valueFrom:
            fieldRef:
              fieldPath: status.podIP
          name: SOME_SERVICE_ADDRESS
          value: "${MY_POD_IP}" <------ provider needs to compose all env vars.


pires:
I think this needs to come out of a draft before we can review. At least, I see zero changes.

pires:
Here are my thoughts:
- VK has access to secrets and configmaps needed to populate pod env vars, [the provider doesn't](https://github.com/virtual-kubelet/virtual-kubelet/blob/master/node/pod.go#L71) unless it explicitly sets it up;
- While the above remains true, the VK **must** perform env var expansion;
- VK can determine most of [the Downward API capabilities](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api) but the `status.podIP`. Actually, part of the Downward API is already expanded (#534), so if anything is missing and can be determined by the VK, we could expand on that;
- In the case of references the VK can't determine, such as `status.podIP`, I don't think the VK needs to be involved at all. The VK will update a pod's statuses, including when `status.podIP` is set but there's no need for it to expand the env var. This is the provider's responsibility. And where does it really matters? In the pod containers, whatever that is for your provider.

cpuguy83:
So here's what I'm thinking:

Currently the pod reconiler handles downward API semantics, but really it has nothing to do with the the recociler at all.
We could just provide helpers for applying the downard API, but then every provider will implement this the same way anyway, so it seems weird to push the responsibility out.
So...

I'm thinking we can add a configuration to the `PodController` to have it take a `PodApplier`:

```go
type PodApplier interface {
    Apply(context.Context, *corev1.Pod)
}
```

The downward API would be applied through this `PodApplier` interface, which would be set by default.
Consumers of the pod controller can set up their own applier when creating the pod controller. We'd provide helpers to create downward API appliers which the consumer can mix in with their own applier.

pires:
As mentioned over slack, I like the idea of being able to specify multiple `Applier` to provide some sort of chain of appliers. If it's just one applier wrapping other _appliers_, it seems one-too-many indirection layers.

uzuku:
@cpuguy83 @pires Hi, I wonder if the proposal has gained any ground since the last post, as I found the exposure of pod info like pod.IP was quite a common requirement for many applications. Thanks!

justnoise:
I discussed this issue with @cpuguy83 today in slack. The feeling was that the best way to apply environment variables would be to create helpers that providers could call.  E.G. `downwardapi.ResolveConfigMapRefs(ctx, &pod, configMapLister)`.

With this setup the feeling is that an `Applier` interface is no longer necessary.

My lingering question is how an existing provider that currently relies on `PodController.createOrUpdatePod()` to call `populateEnvironmentVariables()` would fare with that type of change.  What are the compatibility guarantees between VK releases?  Is there a good way to opt-in to the new behavior?  One advantage of using Appliers would allow VK to provide a default behavior that matches the current implementation.

cpuguy83:
> Is there a good way to opt-in to the new behavior?

I think we can make the extra listers optional in the pod controller config.
If people feel like it's needed we can be really explicit and have a bool to completely disable the current behavior.

> What are the compatibility guarantees between VK releases?

I do not want to introduce breaking changes. So behavior changes need to be opt-in somehow.
I think if the listers (services, config maps, secrets) are optional we can just not resolve references for the missing lister(s).

cpuguy83:
There are a couple of WIP PR's to introduce downward API helpers so providers can apply this to the pod instead of expecting the core runtime to do it.

See #827 

Going to close this one.

Thanks!

PR #651, opened 2019-06-04
Fix bad recursion in errdefs check

cpuguy83:


cpuguy83:
Going to self-merge since this an important fix. Can follow-up if someone thinks it is needed.
Thanks!

sargun:
Lgtm. Wondering if there is anything we can do to prevent this.

PR #654, opened 2019-06-04
Update node defaults

### LINKED ISSUE #652: Node defaults should be adjusted
These were best guess at what should be default, I think they should be updated (and functional options adjusted) before 1.0

cpuguy83:
Do not enable leases by default in the node constructor.
Simplify node constructor to not require a lease client when leases may
not even be enabled.

Updates node status update interval to the kubelet default of 10s (was
5s in vk).

Closes #652

### DIFF
--- a/cmd/virtual-kubelet/commands/root/root.go
+++ b/cmd/virtual-kubelet/commands/root/root.go
@@ -35,6 +35,7 @@ import (
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/kubernetes/typed/coordination/v1beta1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -140,13 +141,17 @@ func runRootCommand(ctx context.Context, c Opts) error {
 		"watchedNamespace": c.KubeNamespace,
 	}))
 
+	var leaseClient v1beta1.LeaseInterface
+	if c.EnableNodeLease {
+		leaseClient = client.CoordinationV1beta1().Leases(corev1.NamespaceNodeLease)
+	}
+
 	pNode := NodeFromProvider(ctx, c.NodeName, taint, p)
 	node, err := vkubelet.NewNode(
 		vkubelet.NaiveNodeProvider{},
 		pNode,
-		client.CoordinationV1beta1().Leases(corev1.NamespaceNodeLease),
 		client.CoreV1().Nodes(),
-		vkubelet.WithNodeDisableLease(!c.EnableNodeLease),
+		vkubelet.WithNodeEnableLeaseV1Beta1(leaseClient, nil),
 		vkubelet.WithNodeStatusUpdateErrorHandler(func(ctx context.Context, err error) error {
 			if !k8serrors.IsNotFound(err) {
 				return err
--- a/vkubelet/node.go
+++ b/vkubelet/node.go
@@ -45,8 +45,11 @@ type NodeProvider interface {
 //
 // Use the node's `Run` method to register and run the loops to update the node
 // in Kubernetes.
-func NewNode(p NodeProvider, node *corev1.Node, leases v1beta1.LeaseInterface, nodes v1.NodeInterface, opts ...NodeOpt) (*Node, error) {
-	n := &Node{p: p, n: node, leases: leases, nodes: nodes, chReady: make(chan struct{})}
+//
+// Note: When if there are multiple NodeOpts which apply against the same
+// underlying options, the last NodeOpt will win.
+func NewNode(p NodeProvider, node *corev1.Node, nodes v1.NodeInterface, opts ...NodeOpt) (*Node, error) {
+	n := &Node{p: p, n: node, nodes: nodes, chReady: make(chan struct{})}
 	for _, o := range opts {
 		if err := o(n); err != nil {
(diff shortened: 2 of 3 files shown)


vkubelet/node.go
```diff
@@ -189,6 +188,12 @@ func (n *Node) Run(ctx context.Context) error {
 	if n.disableLease {
 		log.G(ctx).Info("Node leases not supported, falling back to only node status updates")
 	}
+	if n.leases == nil {
+		// now that everything is initialized, just set disableLease to true since
+		// there is no lease client set.
+		// this just makes things easier to read later instead of having a nil check
+		n.disableLease = true
+	}
```
pires: Can this `if` block move up to become an `if-else` block which is started in line 171?

vkubelet/node.go
```diff
@@ -58,12 +58,20 @@ func NewNode(p NodeProvider, node *corev1.Node, leases v1beta1.LeaseInterface, n
 // NodeOpt are the functional options used for configuring a node
 type NodeOpt func(*Node) error
 
-// WithNodeDisableLease forces node leases to be disabled and to only update
-// using node status
-// Note that this will force status updates to occur on the ping interval frequency
-func WithNodeDisableLease(v bool) NodeOpt {
+// WithNodeEnableLeaseV1Beta1 enables support for v1beta1 leases.
+// If client is nil, leases will not be enabled.
+// If baseLease is nil, the a default base lease will be used.
```
pires: typo: _the a_

vkubelet/node.go
cpuguy83: I didn't want to log the message about falling back if no node lease client was even supplied.

cpuguy83: Ok, I have a better solution, pushed now.

cmd/virtual-kubelet/commands/root/root.go
```diff
@@ -140,13 +141,17 @@ func runRootCommand(ctx context.Context, c Opts) error {
 		"watchedNamespace": c.KubeNamespace,
 	}))
 
+	var leases v1beta1.LeaseInterface
+	if c.EnableNodeLease {
+		leases = client.CoordinationV1beta1().Leases(corev1.NamespaceNodeLease)
```
pires: Suggestion: `leaseClient` as `leases` is unclear to me.

vkubelet/node.go
```diff
@@ -169,27 +168,25 @@ func (n *Node) Run(ctx context.Context) error {
...
-
-		n.lease = l
+	if n.leases == nil {
+		n.disableLease = true
+		return n.controlLoop(ctx)
 	}
 
-	if n.disableLease {
+	n.lease = newLease(n.lease)
+	setLeaseAttrs(n.lease, n.n, n.pingInterval*5)
```
pires: Why `pingInterval*5`?

cpuguy83: This makes sure the lease is valid for plenty of time.
I just picked what seemed like something good enough since the ping interval is typically very small and I want to make sure the lease doesn't go invalid because of a temporary problem.

Can update if you think of a better value.

vkubelet/node.go
```diff
@@ -136,7 +135,7 @@ type Node struct {
 
 // The default intervals used for lease and status updates.
 const (
-	DefaultPingInterval         = 5 * time.Second
+	DefaultPingInterval         = 10 * time.Second
```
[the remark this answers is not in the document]
cpuguy83: This is the default kubelet uses (from what I can tell) for status updates.

vkubelet/node.go
```diff
@@ -169,27 +168,25 @@ func (n *Node) Run(ctx context.Context) error {
...
-			if errors.IsNotFound(err) {
-				n.disableLease = true
-			} else {
-				return pkgerrors.Wrap(err, "error creating node lease")
-			}
-		}
-		log.G(ctx).Debug("Created node lease")
-
-		n.lease = l
+	if n.leases == nil {
```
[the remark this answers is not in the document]
cpuguy83: We either set `leases` to nil or set `disableLease`. I personally find the bool check to be nicer than a nil check.
It's not really two things determining it. We are disabling leases if no lease client is provided OR if leases are supported on the cluster.

vkubelet/node.go
```diff
@@ -58,12 +58,20 @@ func NewNode(p NodeProvider, node *corev1.Node, leases v1beta1.LeaseInterface, n
 // NodeOpt are the functional options used for configuring a node
```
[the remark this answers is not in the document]
cpuguy83: Add to `NewNode` docstring

cmd/virtual-kubelet/commands/root/root.go
cpuguy83: Done.

vkubelet/node.go
```diff
@@ -58,12 +58,20 @@ func NewNode(p NodeProvider, node *corev1.Node, leases v1beta1.LeaseInterface, n
...
-func WithNodeDisableLease(v bool) NodeOpt {
+// WithNodeEnableLeaseV1Beta1 enables support for v1beta1 leases.
+// If client is nil, leases will not be enabled.
+// If baseLease is nil, a default base lease will be used.
+//
+// The lease will be updated after each successful node ping. To change the
+// lease update interval, you must set the node ping interval.
+// See WithNodePingInterval().
+//
+// This also affects the frequencey of node status updates.
```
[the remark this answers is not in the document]
cpuguy83: Fixed.

vkubelet/node.go
cpuguy83: We can update this as needed (or maybe provide an option override this). This value is what was originally set (of course did double the default ping interval here).

PR #657, opened 2019-06-07
Delete virtual node when virtual kubelet is removed

sashayakovtseva:
Right now when virtual kubelet pod is removed, virtual node is transitioned in a state `NotReady` and remains in k8s node list. This PR fixes that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,9 @@ loganalytics.json
 # VS Code files
 .vscode/
 
+# IntelliJ Goland files
+.idea
+
 # Terraform ignores
 **/.terraform/**
 **/terraform-provider-kubernetes
--- a/cmd/virtual-kubelet/commands/root/root.go
+++ b/cmd/virtual-kubelet/commands/root/root.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"os"
 	"path"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -60,6 +61,9 @@ This allows users to schedule kubernetes workloads on nodes that aren't running 
 }
 
 func runRootCommand(ctx context.Context, c Opts) error {
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
@@ -102,8 +106,15 @@ func runRootCommand(ctx context.Context, c Opts) error {
 	configMapInformer := scmInformerFactory.Core().V1().ConfigMaps()
 	serviceInformer := scmInformerFactory.Core().V1().Services()
 
-	go podInformerFactory.Start(ctx.Done())
-	go scmInformerFactory.Start(ctx.Done())
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		podInformerFactory.Start(ctx.Done())
+	}()
+	go func() {
+		defer wg.Done()
+		scmInformerFactory.Start(ctx.Done())
+	}()
 
 	rm, err := manager.NewResourceManager(podInformer.Lister(), secretInformer.Lister(), configMapInformer.Lister(), serviceInformer.Lister())
 	if err != nil {
@@ -188,7 +199,9 @@ func runRootCommand(ctx context.Context, c Opts) error {
 	}
 	defer cancelHTTP()
 
+	wg.Add(1)
 	go func() {
+		defer wg.Done()
 		if err := pc.Run(ctx, c.PodSyncWorkers); err != nil && errors.Cause(err) != context.Canceled {
 			log.G(ctx).Fatal(err)
 		}
@@ -204,7 +217,9 @@ func runRootCommand(ctx context.Context, c Opts) error {
 		}
 	}
 
+	wg.Add(1)
 	go func() {
+		defer wg.Done()
 		if err := node.Run(ctx); err != nil {
 			log.G(ctx).Fatal(err)
 		}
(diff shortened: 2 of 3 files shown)


sashayakovtseva:
Not sure if node should be silently removes so that test passes, or a test should be updated. 
Let me know what it better!

cpuguy83:
Can you point to where the kubelet handles this?

.gitignore
```diff
@@ -33,6 +33,9 @@ loganalytics.json
 # VS Code files
 .vscode/
 
+# IntelliJ Goland files
```
sargun: Can you split this into a separate PR?

cmd/virtual-kubelet/commands/root/root.go
```diff
@@ -60,6 +61,9 @@ This allows users to schedule kubernetes workloads on nodes that aren't running
 }
 
 func runRootCommand(ctx context.Context, c Opts) error {
+	var wg sync.WaitGroup
+	defer wg.Wait()
```
sargun: This seems like it'd be very confusing to troubleshoot a deadlock. Why not explicitly move it?

cmd/virtual-kubelet/commands/root/root.go
```diff
@@ -102,8 +106,15 @@ func runRootCommand(ctx context.Context, c Opts) error {
...
 	serviceInformer := scmInformerFactory.Core().V1().Services()
 
-	go podInformerFactory.Start(ctx.Done())
-	go scmInformerFactory.Start(ctx.Done())
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		podInformerFactory.Start(ctx.Done())
+	}()
+	go func() {
```
sargun: Why not use errgroup?

vkubelet/node.go
```diff
@@ -233,7 +233,7 @@ func (n *Node) controlLoop(ctx context.Context) error {
 	for {
 		select {
 		case <-ctx.Done():
-			return nil
+			return n.nodes.Delete(n.n.Name, &metav1.DeleteOptions{})
```
sargun: This seems problematic. What if there are still running pods? Does normal kubelet exhibit this behaviour?

sashayakovtseva:
@cpuguy83 Node monitoring code can be found [here](https://github.com/kubernetes/kubernetes/blob/release-1.13/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L651) if that is what you meant.
As I see k8s doesn't remove node automatically there, also based on experiments node will remain in the list.

.gitignore
sashayakovtseva: Sure, see #660 please.

cmd/virtual-kubelet/commands/root/root.go
sashayakovtseva: For sake of simplicity: wait in one place instead of each return.
Another option would be to use wait group for goroutine with `node.Run` only, but leaving the rest goroutines without attention is not nice, I think.

cmd/virtual-kubelet/commands/root/root.go
sashayakovtseva: I haven't heard of it before, can you elaborate? :)
Will it make any difference?

cpuguy83:
I would say then let's follow what the kubelet does.
Since vk is a library, consumers can choose to delete the node or not on shutdown.

pires:
@sashayakovtseva as you correctly pointed out, there's a node lifecycle controller that handles this. That controller, however, is not part of the `kubelet` but of the `controller-manager`. What is the reasoning behind having similar behavior in the VK when the `controller-manager` will do it?

sashayakovtseva:
@pires The initial idea was to make cleanup after pod removal, b/c node was created in pod so in my understanding it should be removed with pod removal as well. 
Does `nodelifecycle` removes `NotReady` nodes? According to source code, it does not, that is why this PR was suggested in the first place.

sashayakovtseva:
@cpuguy83 
> consumers can choose to delete the node or not on shutdown

Do you mean end user will have to remove node manually? Or there is some flag in vk? If not, that is another option how node removal can be implemented.

cpuguy83:
I mean that whatever integration people choose to make, they can handle it outside of the scope of vk core libraries.
I do not think the life-cycle of the node should be tied to a particular pod life-cycle implicitly.
There are good reasons to take down a pod that do not mean the node should go away.
So ultimately, whatever is managing the pod should be responsible for cleaning it up if it shouldn't be there.

pires:
@sashayakovtseva I had missed the fact you're running it as a pod but, still, what's the difference here? Say the `kubelet` didn't exit cleanly, it is marked as `NotReady` and the node stays in the API, right?

I understand what you may want to do here but I still think it shouldn't be the responsibility of the VK to guarantee it exists cleanly. If it doesn't exit cleanly, how can you then guarantee the corresponding node is removed?

sashayakovtseva:
@pires Hm, there are no guarantees in case vk exits abnormally. And the question is, should they be? I mean if it suddenly breaks everything can go wrong, only external monitor can cleanup in that case.

sashayakovtseva:
Thanks everyone for comments and discussion, looks like this PR doesn't bring any value to the virtual-kubelet. Closing this PR, unless anyone else thinks this behavior may be useful.

pires:
@sashayakovtseva but the same applies to the kubelet, right? Maybe I'm missing something, and if that's the case, please, do point it out. Always learning and I've taken lessons from this thread already, so thank you for your contribution(s).

sashayakovtseva:
@pires Yes, when kubelet falls down, node is left in the list afaik. Maybe you are right and there is no much point in cleaning up node from vk pod. 
However, we will use node removal in our fork as it looks convenient for us :) 
Thank you for pointing things out!

PR #658, opened 2019-06-10
Move around some packages

### LINKED ISSUE #637: Move vkubelet package to root
The `vkubelet` package naming is redundant.
It exists because the CLI used to be in the repo root, but this has moved to cmd/virtual-kubelet.

I propose we move the package to the root of the repo for 1.0

cpuguy83:
- Move tracing exporter registration to the CLI b/c in the library we just don't care about exporters since this is a last-mile config

- Rename `vkubelet` package to `node`
    `vkubelet` does not convey any information to the consumers of the
    package.
    Really it would be nice to move this package to the root of the repo,
    but then you wind up with... interesting... import semantics due to the
    repo name... and after thinking about it some, a subpackage is really
    not so bad as long as it has a name that convey's some information.

    `node` was chosen since this package deals with all the semantics of
    operating a node in Kubernetes. This does create some
    redundency/confusion with the `node.Node` object which is really only
    dealing with the node lifecycle in Kubernetes, but this can (and
    probably should) be updated to something more appropriate to what it is.

Closes #637


cmd/virtual-kubelet/commands/root/root.go
```diff
@@ -147,12 +147,12 @@ func runRootCommand(ctx context.Context, c Opts) error {
 	}
 
 	pNode := NodeFromProvider(ctx, c.NodeName, taint, p)
-	node, err := vkubelet.NewNode(
-		vkubelet.NaiveNodeProvider{},
+	nodeRunner, err := node.NewNode(
```
pires: Suggestion:
- `node.NewNodeLifecycleManager`, or
- `node.NewNodeController` to follow similar semantics as `node.NewPodController`.

cpuguy83: I *almost* did this in the PR 😄

cpuguy83: Took your suggestion and change to NodeController

cpuguy83: This does cause golint to complain about stuttering.

pires: golint is too serious.

PR #664, opened 2019-06-12
Remove resourcemanager from conroller public API

cpuguy83:
We still use it internally, but this does not need to be part of the
public API. Instead just have callers pass us the relevent listers and
we create our own resource manager.


cpuguy83:
🤷‍♂ 

Sorry about that.
Maybe it was just "let's do this separately"?

pires:
Maybe just a different perspective back then, or code, or whatever. Honestly, don't care. Just messing with you! 🙇

PR #665, opened 2019-06-12
Add authors to file heads

cpuguy83:
Realized as I was starting to copy some stuff to other repos that we
    should go ahead and add this.

### DIFF
--- a/cmd/virtual-kubelet/commands/root/tracing_register.go
+++ b/cmd/virtual-kubelet/commands/root/tracing_register.go
@@ -1,3 +1,17 @@
+// Copyright © 2017 The virtual-kubelet authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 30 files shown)


AUTHORS
```diff
@@ -1,13 +1,114 @@
...
+Fei Xu <xufei40@huawei.com>
+Hector Fernandez <cpuguy83@gmail.com>
+Hector Fernandez <hfernandez@mesosphere.com>
+Hongbin Lu <cpuguy83@gmail.com>
+Jacob LeGrone <git@jacob.work>
+Jacob LeGrone <junjiez@microsoft.com>
+Jake Bjorke <junjiez@microsoft.com>
+Jeremy Rickard <cpuguy83@gmail.com>
+Jeremy Rickard <jeremy.rickard@microsoft.com>
+Jeremy Rickard <junjiez@microsoft.com>
```
pires: The name:email pairs here seem wrong all over the doc. I'm thinking this can happen because of commits that are co-authored, such as in PR merges.

cpuguy83: Yes I noticed, was going to check out AUTHORS in other repos to see if its expected or not.

cpuguy83: Ok, I changed the authors script to what we use in moby.

cpuguy83:
Looks like it's removed these two:

- Jess Frazelle <acidburn@microsoft.com> (@jessfraz)
- Julien Stroheker <julienstroheker@gmail.com> (@julienstroheker)

Maybe these were co-authored commits or something?

cpuguy83:
I removed the 2nd commit which updates AUTHORS so we can get the file header change in, authors will be updated separately as it seems it will require extra tweaking.

PR #667, opened 2019-06-13
Remove version package

cpuguy83:
This moves version declaration into cmd/virtual-kubelet/main.go
From here values are passed down into the things that need it.


pires:
I've looked into this and I have no opinion. I don't understand it, tbh, and can't recall right now the details of past discussions. I'll also be quite busy over the next few days and don't want to block this from being merged, so I'm removing myself from the reviewers list.

cpuguy83:
The main thing is this is just moving from a global to a value set on the CLI, which is what is versioned.
This is something I've done on both the azure-aci and azure-batch repos b/c it simplifies imports.

Doing it here just means I don't need to keep doing it.

pires:
Thank you for the clarification, it does make tons of sense to me.

Issue #494, opened 2019-01-09
What to do with providers?


cpuguy83:
Right now this repo is full of provider specific code which also has a separate maintenance flow from VK core.
Should this code continue to live in VK?

Part of the direction that I see for VK is to provide a good framework for providers to build their own integrations that integrate well with their infrastructure. Mostly I would expect to see that code live within a provider's org.

I do see some benefit to having a common daemon implementation, but I think realistically this limits what is possible and makes a pretty bloated daemon.

So I want to get a feel for what other maintainers/providers are thinking along these lines. Is there still a benefit to:
1. Keep existing providers in tree
2. Keep adding new providers in tree

This also brings up a slight tangent, should we continue to release an all-inclusive VK image as we do today? If so we need to formalize this release process.

cpuguy83:
Added a bunch of people here, but sure other providers are missing... I didn't see them in the list... maybe @rbitia knows GH handles?

resouer:
My random thoughts is, eventually I believe providers could find a new home like [cloudproviders](https://github.com/kubernetes/cloud-provider-aws) in Kubernetes org. That said, it make a lot sense for VK to follow the same maintenance/migrate pattern in kube community. 

The experience is, having providers in tree for certain period of time do benefit to accumulate issues/PRs/discussion in the same picture, especially when the interface/architecture of the main project is still quickly evolving. And of course the cost is extra maintenance effort, and dragging of release lifecycle.

I would suggest, maybe, we can try to align providers migration process with VK incubation process in CNCF? cc @ritazh

caniszczyk:
You can also host these providers as sub projects of the VK github organization as one path forward, each other their own maintainers/governance per sub project

pires:
+1 for making VK a clean framework anyone can use to build their provider with, and having providers living elsewhere. Eventually, a path for something like a _certified VK provider_ could be specified, which end result could be the provider code moving to the VK github org.

caniszczyk:
My advice to the VK project would be to create a GOVERNANCE.md that dictates how projects are added/removed along with maintainers, e.g.,

https://github.com/envoyproxy/envoy/blob/master/GOVERNANCE.md
https://github.com/coredns/coredns/blob/master/GOVERNANCE.md

zanetworker:
Should gRPC be considered? communication between the daemon and the providers would be a matter of calling the right remote RPC.

pires:
[...]

PR #668, opened 2019-06-14
Remove azurebatch provider

cpuguy83:
This was moved to https://github.com/virtual-kubelet/azure-batch
Relates to #494


cpuguy83:
Replaced by #670

Will keep this open in case ppl want to do this piecemeal.

jeremyrickard:
Prefer #670.

PR #669, opened 2019-06-14
Use object informers instead of listers

cpuguy83:
We'll need these informers for
https://github.com/virtual-kubelet/virtual-kubelet/pull/586

Changing this now means we don't need to make API changes later.

### DIFF
--- a/node/podcontroller.go
+++ b/node/podcontroller.go
@@ -110,33 +110,36 @@ type PodControllerConfig struct {
 
 	Provider PodLifecycleHandler
 
-	// Listers used for filling details for things like downward API in pod spec.
-	ConfigMapLister corev1listers.ConfigMapLister
-	SecretLister    corev1listers.SecretLister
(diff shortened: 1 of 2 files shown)


node/podcontroller.go
```diff
@@ -126,17 +129,17 @@ func NewPodController(cfg PodControllerConfig) (*PodController, error) {
 	if cfg.PodInformer == nil {
 		return nil, errdefs.InvalidInput("must set informer")
 	}
-	if cfg.ConfigMapLister == nil {
+	if cfg.ConfigMapInformer == nil {
 		return nil, errdefs.InvalidInput("must set config map lister")
```
pires: Error definition is wrong.

pires: Same for all previous lister validations below.

pires: I don't even know if it makes sense to log it like this.

cpuguy83: Updated

PR #674, opened 2019-06-19
Make CLI node CLI framework

cpuguy83:
This makes a basic CLI that can be imported by other projects without
exposing low-level implementation details that would paint us into a
corner w.r.t. 1.0 API guarantees.


node/cli/cli.go
```diff
@@ -0,0 +1,108 @@
...
+	s          provider.Store
+	name       string
+	setupTrace TracingSetupFunc
+}
+
+// WithProvider registers a provider which the cli can be initialized with.
+func WithProvider(name string, f provider.InitFunc) CommandOption {
+	return func(c *Command) {
+		if c.s == nil {
+			c.s = provider.NewStore()
```
sargun: Why do you use the store still? It seems weird to continue to support multiple providers?

node/cli/cli.go
```diff
@@ -0,0 +1,108 @@
...
+		o(&c)
+	}
+
+	name := c.name
+	if name == "" {
+		name = filepath.Base(os.Args[0])
+	}
+
+	var flagOpts root.Opts
+	optsErr := root.SetDefaultOpts(&flagOpts)
```
sargun: I still wish that we called root.NewOps instead.

node/cli/cli.go
```diff
@@ -0,0 +1,108 @@
...
+		if preRun != nil {
+			return preRun(cmd, args)
+		}
+		return nil
+	}
+
+	c.cmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", `set the log level, e.g. "debug", "info", "warn", "error"`)
+
+	c.cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		if logLevel != "" {
```
sargun: Do we want to support viper?

node/cli/cli.go
cpuguy83: Yeah, I debated removing it...
Really it'd be nice to be able to use this at least to switch beetwen mockv0 and mockv1 (async).
But I guess we could just as easily create two CLI's.

node/cli/cli.go
cpuguy83: Yeah, we can do that.

node/cli/cli.go
cpuguy83: Not really. At the moment I'm trying to hide as much of the implementation details as possible (so, cobra and pflag).

cpuguy83: Really even this logger config and the tracing config probably needs to be removed because I can't really think of a good way to abstract these and let people use whatever logging or tracing implementation they want.

cpuguy83:
Ok, I'm marking this ready for review.

node/cli/cli.go
pires: Or just one CLI with a flag `--mode={sync,async}` that defaults to whatever is preferred - I'd argue sync is as it's been around for longer.

pires:
Why is the package `node/cli` and not just `cli`?

cpuguy83:
> Why is the package node/cli and not just cli?

Because it's the cli lib for the node.

---

I'm a bit loathe to include this tbh. I built it b/c others seemed interested in this for rapid prototyping.
I'm thinking perhaps this should move to a separate repo. For the mock provider we can import this lib in still.
It would allow us to version separately at least.

node/cli/cli.go
cpuguy83: Perhaps mode is more suitable as a provider specific config passed through the config file.

pires:
I'm all in to move mock provider(s) to a new repo and link it as an example from our docs!

cpuguy83:
Oh I as referring to this CLI rather than mock, but 🤷‍♂ 
Moving mock does make it harder to do e2e testing.

cpuguy83:
Closing, I've moved this to https://github.com/virtual-kubelet/node-cli/
I'll open a PR to move out mock into a separate repo.

PR #676, opened 2019-06-21
Allow Annotations to be set on nodes derived from a Provider

gabrielhartmann:
Today the following fields on the Node are derived from the Provider.
```
NodeStatus:
    Capacity,
    Allocatable,
    Conditions,
    Addresses,
    DaemonEndpoints,
```
```
ObjectMetadata:
    Labels:
        beta.kubernets.io/os,
```

Here we extend this list to also set:
```
ObjectMetadata:
    Annotations,
```

### DIFF
--- a/providers/mock/mock.go
+++ b/providers/mock/mock.go
@@ -12,9 +12,11 @@ import (
 
 	"github.com/virtual-kubelet/virtual-kubelet/errdefs"
 	"github.com/virtual-kubelet/virtual-kubelet/log"
+	"github.com/virtual-kubelet/virtual-kubelet/node"
 	"github.com/virtual-kubelet/virtual-kubelet/node/api"
+	"github.com/virtual-kubelet/virtual-kubelet/providers"
 	"github.com/virtual-kubelet/virtual-kubelet/trace"
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
@@ -32,14 +34,11 @@ const (
 	containerNameKey = "containerName"
 )
 
-// See: https://github.com/virtual-kubelet/virtual-kubelet/issues/632
-/*
 var (
 	_ providers.Provider           = (*MockV0Provider)(nil)
 	_ providers.PodMetricsProvider = (*MockV0Provider)(nil)
-	_ node.PodNotifier         = (*MockProvider)(nil)
+	_ node.PodNotifier             = (*MockProvider)(nil)
 )
-*/
 
 // MockV0Provider implements the virtual-kubelet provider interface and stores pods in memory.
 type MockV0Provider struct {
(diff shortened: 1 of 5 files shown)


providers/mock/mock.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/virtual-kubelet/virtual-kubelet/node"
+	"github.com/virtual-kubelet/virtual-kubelet/providers"
```
pires: Please, move these imports to their rightful place.

providers/mock/mock_test.go
```diff
@@ -1,12 +1,43 @@
...
-/*
-func TestMockLegacyInterface(t *testing.T) {
-	var mlp providers.Provider = &MockLegacyProvider{}
-	_, ok := mlp.(node.PodNotifier)
-	assert.Assert(t, !ok)
+import (
+	"context"
+	"gotest.tools/assert"
+	"github.com/virtual-kubelet/virtual-kubelet/cmd/virtual-kubelet/commands/root"
+	"testing"
```
pires: Same thing about imports here.

providers/mock/mock_test.go
```diff
@@ -1,12 +1,43 @@
...
+func TestNodeFromMockProvider(t *testing.T) {
+	const (
+		NodeName = "nodeName"
+		OperatingSystem = "os"
+		InternalIp = "1.2.3.4"
+		DaemonEndpointPort = 42
+		Version = "v0.1"
+	)
+
+	expectedAnnoations := map[string]string {
```
pires: typo

gabrielhartmann:
Screen Shot 2019-06-24 at 9 37 39 AM

@pires I believe I'm good to go on the CLA front.

cpuguy83:
I think I would prefer if we can just have the provider configure the node object itself rather than having a bunch of functions to do it that we have to keep adding to the interface.

gabrielhartmann:
@cpuguy83 

> I think I would prefer if we can just have the provider configure the node object itself rather than having a bunch of functions to do it that we have to keep adding to the interface.

Agreed.  Given that this PR is an extension of the current mechanism for constructing a node should I consider your comment a blocker?  As this is my first PR I preferred not to make larger changes to the node construction approach.

cpuguy83:
@gabrielhartmann Correct, I don't think we should add it just to remove it quickly. Did you want to take on the work of updating the interface to handle this? Feel free to ping me in Slack about this.

pires:
@cpuguy83 I didn't want to block 1.0 for long so to give us time to think the problem thorough in a next iteration. Guess by not accepting this, you're disagreeing.

cpuguy83:
@pires I think for 1.0, we can just say that cmd and cmd related packages ("providers" package) are not semver, so there may be breaking changes to those packages.

I think this is good rather than trying to rush an un-used/untested interface.

pires:
I think this addition is somewhat important yet trivial to review and the author provided tests in it. While labels and taints are important for scheduling, annotations are where providers can freely provide metadata they care about. That's why I think it's something we should have in 1.0.

cpuguy83:
Are we supporting the CLI for 1.0? I don't think I want to do this.

PR #680, opened 2019-06-26
Add provider method to pass node configuration

### LINKED ISSUE #678: Allow Runtime Definition of Node Annotations
---

### Environment summary

Provider: Generic
Version: N/A
K8s Master Info: Bare metal
Install Method: Custom

### Issue Details

Authors of Providers are unable to provide Annotations on Node metadata in the current `NodeFromProvider` function in `node.go`.  These Annotations are potentially important to arbitrary Kubernetes deployments.

Various fields are currently populated in the status and metadata subsections of the Node object today.  Further extension of the current mechanism of Node definition from the provider is undesirable.  See PR #676.  A new method of defining nodes with input from the provider authors is therefore needed.

### Repo Steps
N/A

cpuguy83:
This eliminates all the one-off calls to for specific things like
`Capacity` and other things.
Instead we configure defaults in the CLI and then pass the full node
object to the provider to change as needed.

This makes sure the provider can change whatever they need to without
having to add tons of methods to the provider interface.

Closes #678

ping @gabrielhartmann

### DIFF
--- a/cmd/virtual-kubelet/commands/root/node.go
+++ b/cmd/virtual-kubelet/commands/root/node.go
@@ -25,6 +25,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const osLabel = "beta.kubernetes.io/os"
+
 // NodeFromProvider builds a kubernetes node object from a provider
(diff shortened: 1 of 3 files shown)


gabrielhartmann:
LGTM

PR #690, opened 2019-07-02
Add "TerminatePod" to lifecycle interface

cpuguy83:
This splits the termination of a pod from the deletion of it.
It allows providers to update the final pod status before deleting it is
deleted form k8s.

Pod is terminated if pod.Phase == Running and pod.DeletionTimestamp is
not nil.

When the provider is ready for the pod to be deleted, the pod status
should be udpated so it's not "Running", but whatever is appropriate for
the exist statuses of the containers.
Once the status is updated a new event will be triggered from k8s which
will trigger us to call `DeletePod` as before.

### DIFF
--- a/providers/mock/mock.go
+++ b/providers/mock/mock.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/virtual-kubelet/virtual-kubelet/errdefs"
@@ -47,10 +48,13 @@ type MockV0Provider struct {
 	operatingSystem    string
 	internalIP         string
 	daemonEndpointPort int32
-	pods               map[string]*v1.Pod
-	config             MockConfig
-	startTime          time.Time
-	notifier           func(*v1.Pod)
+
+	mu   sync.Mutex
+	pods map[string]*v1.Pod
+
+	config    MockConfig
+	startTime time.Time
+	notifier  func(*v1.Pod)
 }
 
 // MockProvider is like MockV0Provider, but implements the PodNotifier interface
@@ -208,9 +212,10 @@ func (p *MockV0Provider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 		})
 	}
 
+	p.mu.Lock()
 	p.pods[key] = pod
 	p.notifier(pod)
-
+	p.mu.Unlock()
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


cpuguy83:
ping @gabrielhartmann

providers/mock/mock.go
```diff
@@ -229,49 +234,87 @@ func (p *MockV0Provider) UpdatePod(ctx context.Context, pod *v1.Pod) error {
...
+}
+
+func (p *MockV0Provider) terminatePod(ctx context.Context, pod *v1.Pod) {
 	now := metav1.Now()
-	delete(p.pods, key)
 	pod.Status.Phase = v1.PodSucceeded
-	pod.Status.Reason = "MockProviderPodDeleted"
-
-	for idx := range pod.Status.ContainerStatuses {
+	pod.Status.Reason = "MockProviderPodTermianted"
```
sargun: termianted -> Terminated

providers/mock/mock.go
```diff
@@ -229,49 +234,87 @@ func (p *MockV0Provider) UpdatePod(ctx context.Context, pod *v1.Pod) error {
...
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	pod, exists := p.pods[key]
+	if !exists {
+		return errdefs.NotFound("pod not found")
+	}
+
+	if pod.Status.Phase == v1.PodRunning {
```
sargun: Would it be expected all providers do this?

cpuguy83: No, I need to account for this.

cpuguy83: Or more precisely, I need to check if the provider implements TerminatePod (and split TerminatePod from the main interface).

I do expect all providers to set the phase of the pod, though.

sargun:
I do not like this approach because it hoists the responsibility of keeping track of "state" on the provider. Instead, I think that the VK should be responsible for keeping track of which pods that are in the APIServer, and what pods are in the provider. It should then try to reconcile them as best as possible.

Specifically, the way I see it working is rather than a separate terminatePod/deletePod is that the PC would reconcile the pods in apiserver and the provider at startup time. 

If a pod is marked for deletion in apiserver, but it exists in the provider, I would rather see us call deletePod (or terminatePod), and use the task status that gets passed back by the asynchronous notification mechanism (notifyPods?). Otherwise, after terminatePods, the provider still has to keep the pod status in memory, until deletePods. 

Alternatively, if we were able to use the return of deletePod to generate a new pod status, that would also be valuable, in that it wouldn't require another round trip to the provider.

sargun:
(P.S. I have bandwidth now, if you're open the hoisting state into VK, I am more than happy to scrap something together)

cpuguy83:
@sargun What you are asking for is precisely what `TerminatePod` does, unless I'm mistaken.

VK gets a notification about the pod deletion so we call TerminatePod (if the pod is running) then wait for the provider to notify us of the new pod status, vk updates the pod status in k8s which triggers a new notification from k8s about the pod status change which then prompts VK to call DeletePod.

cpuguy83:
VK should not be holding any pod state except what is in the API server (this is what the pod informer cache does). It's up to the provider to tell VK about the state change (or it could do it directly, I suppose), but VK pushes that state to the API server.
VK is *always* acting upon the desired state from the API server.

sargun:
The queue and podinformer keep state. The queue keeps track of "there is a change" and "needs to be (re)tried". 

I think that the problem I have is that this requires that providers build tombstoning logic. The tombstoning logic is such that there is a difference in lifecycle between "stop" and "delete". The reason is that between terminatepod and deletepod, the podstatus must be preserved. I think that making every Provider implementation take on this responsibility is suboptimal, and instead the transition should be just "managed" by VK, and the callback should be "transition to terminal status". 

If you believe instead that every provider should implement this, then I believe this approach is good.

cpuguy83:
@sargun I do not see how VK can keep track of terminal state (such as container exit codes, pod success/failure, etc).
The provider is already expected to track the pod's state which is why VK either listens for pod state notifications (from the provider) or probes the provider for pod state periodically.

cpuguy83:
Discussed this with @sargun on the community call and I now understand his concerns.
Closing this as it does not look like it's the right approach.

PR #691, opened 2019-07-03
Don't ignore all GetPod errors on deletePod

cpuguy83:
A legit error could be passed here and we don't want to ignore it.


cpuguy83:
We all missed it!

PR #694, opened 2019-07-03
node: set label to enable selection

### LINKED ISSUE #692: How to distinguish virtual-node from kubernetes node
### Environment summary

Provider: ECI，Alibaba Cloud

### Issue Details
Is there any specific label or other symbol to distinguish virtual-node from kubernetes node.   
<a href="https://github.com/kubernetes-incubator/cluster-proportional-autoscaler">cluster-proportional-autoscaler</a> is a component autoscaler for coredns and others. It will calculate the replicas based on cores. So we need to ignore the virtual-node, otherwise there will be so many replicas.

pires:
By setting Node resource label `beta.kubernetes.io/instance-type` to `virtual-kubelet`, we are enabling controllers to filter nodes that are virtual-kubelet instances.
Two examples of controllers that may benefit from this are:
- cloud-controller-manager node controller, which deletes Node resources that
are not present in the cloud-provider;
- cluster-autoscaler, which scales up/down worker machines to satisfy a
Kubernetes cluster needs.

Fixes #692

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -33,6 +33,53 @@ import (
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
+const (
+	// InstanceTypeLabelKey is part of the built-in Node labels.
+	// See https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#interlude-built-in-node-labels
+	InstanceTypeLabelKey = "beta.kubernetes.io/instance-type"
+
+	// InstanceTypeLabelValue will clearly brand the VK node for proper parsing
+	// such as required by cloud-providers.
+	// As an example, see https://github.com/virtual-kubelet/virtual-kubelet/issues/692
+	InstanceTypeLabelValue = "virtual-kubelet"
+
+	// OSLabelKey is part of the built-in Node labels.
+	// See https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#interlude-built-in-node-labels
+	OSLabelKey = "beta.kubernetes.io/os"
+)
+
+// NewNodeSpec builds a Kubernetes Node resource.
+func NewNodeSpec(name string, taint *corev1.Taint, version string) *corev1.Node {
+	taints := make([]corev1.Taint, 0)
+
+	if taint != nil {
+		taints = append(taints, *taint)
+	}
+
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Labels: map[string]string{
+				"type":                   "virtual-kubelet",
+				"kubernetes.io/role":     "agent",
+				"kubernetes.io/hostname": name,
+				InstanceTypeLabelKey:     InstanceTypeLabelValue,
+			},
+		},
+		Spec: corev1.NodeSpec{
+			Taints: taints,
+		},
+		Status: corev1.NodeStatus{
+			NodeInfo: corev1.NodeSystemInfo{
+				Architecture:   "amd64",
+				KubeletVersion: version,
+			},
+		},
+	}
+
+	return node
+}
+
 // NodeProvider is the interface used for registering a node and updating its
 // status in Kubernetes.
 //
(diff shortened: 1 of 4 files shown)


node/node.go
```diff
@@ -64,6 +75,13 @@ type NodeProvider interface {
 // underlying options, the last NodeControllerOpt will win.
 func NewNodeController(p NodeProvider, node *corev1.Node, nodes v1.NodeInterface, opts ...NodeControllerOpt) (*NodeController, error) {
 	n := &NodeController{p: p, n: node, nodes: nodes, chReady: make(chan struct{})}
+
+	// Apply instance type label.
+	labels := make(map[string]string)
```
cpuguy83: Can we move this to the default generated node config?

pires: Makes sense, will do!

node/node.go
```diff
@@ -33,6 +33,53 @@ import (
...
+
+	if taint != nil {
+		taints = append(taints, *taint)
+	}
+
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Labels: map[string]string{
+				"type":                   "virtual-kubelet",
```
pires: In case you're wondering, I didn't want to remove this because someone out there may be using it already.

cpuguy83:
A couple of thoughts:

- This, along with a few other functions (like UpdateNodeStatus, PatchNodeStatus) should maybe be moved to `node/nodeutil`
- We should think about this as a default node spec and let the caller set taints on the returned node.
- We could maybe let `NewNode` take opts like `... func(*v1.Node)`  if we do want to have an API where the caller passes everything in at once.
- I'm not sure `instance-type=virtual-kubelet` makes sense and would generally be up to the implementer to deal with this... but I'm also not familiar with how CCM or the autoscaler works. An instance-type seems right, but I'm not sure we can define an appropriate default.

pires:
I agree with your feedback and here's some thoughts too.

- I tried to make as little change as necessary to deliver the feature within the 1.0 release window. I never pondered changing how a `Node` is built. I like the _opts_ idea but I'd suggest we do that post-1.0.
- The `instance-type` label seemed to me like a good placeholder. I am not convinced it is/it is not the right placeholder.
- While the providers themselves can change the `Node` before actually starting the `NodeController`, the question here is the user can only do it after the `Node` is created, adding up to the burden on ops. With this change, and for as long as providers don't overwrite the `instance-type` label, users have a way to filter all instances of `virtual-kubelet` in their clusters.

cpuguy83:
@pires I'm thinking I would like to push this post-1.0 just because there's nothing in the current setup that prevents someone from setting instance-type already. In order to get instance-type all the provider repos will need updates one way or the other.

> the question here is the user can only do it after the Node is created, adding up to the burden on ops

I'm not sure I understand this statement.
In my commentary I'm more referring to:

```go
n := node.NewNode()
n.Spec.Taints = taints
node.NewController(ctx, node, ...)
```

Rather than the node actually being created in k8s.

pires:
> > the question here is the user can only do it after the Node is created, adding up to the burden on ops
> 
> I'm not sure I understand this statement.

See the issue that spawned this PR, #692. Basically, users who have VKs in their clusters, cannot segregate them from `Node` instances that are backed by kubelet (or others). I just wanted to provide a way they can have it.

Now, I understand providers would need to have new releases but I think that's expectable as VK reaches 1.0.

Everything else we discussed in this issue can be post-1.0 but this I'd like to see in 1.0.

cpuguy83:
@pires My main concern (outside of long term support) here is we are creating a default that assumes something special about a node created from this package when what makes it special is the provider itself rather than VK. This is why I've pushed taints and other things outside of this package, meanwhile it is relatively trivial for a provider to set `instance-type` on the node object that they pass into the `NodeController`.

pires:
I don't disagree but it needs go be done explicitly by provider and the user will have to account for all providers specific labels, when this is only one label for _everybody_ using VK (unless they overwrite it).

cpuguy83:
Fair enough. Can we update `NewNodeSpec` to either take no options or functional options?

cpuguy83:
Also really thinking this should go into a `node/nodeutils` package.

cpuguy83:
Per Slack discussion, pulling this from the 1.0 milestone. This is not a breaking change and can be added in a subsequent release. Providers can set this explicitly as needed until then.

pires:
After a little more thought, I agree with @cpuguy83 opinion that settings labels and taints is sole responsibility of the provider.

PR #687, opened 2019-07-01
Use the new  node-cli framework for cmd

cpuguy83:

PR #697, opened 2019-07-03
Move CLI related packages into internal

cpuguy83:
We don't want people to import these packages, so move these out into
private packages.


cpuguy83:
Doing this as an alternative to #687. That PR is working, but modules is being problematic (e.g. `go mod tidy` keeps acting like it is missing modules when it's not... everything builds fine but `go mod tidy` keeps having problems with `golang.org/x/time/rate`

PR #698, opened 2019-07-03
Update README.md to account for recent changes

cpuguy83:
This is preparing for 1.0


cpuguy83:
Yeah it could use some more love.
Also realized we need a package doc in the root of the repo.

PR #703, opened 2019-07-16
fix several typo

zhuangqh:
Signed-off-by: zhuangqh <zhuangqhc@gmail.com>


zhuangqh:
> I feel called out :)
> 
> These all look good, except the one for "exec", that one is good as is.

done

PR #706, opened 2019-07-24
Add a test which tests the e2e lifecycle of the pod controller

sargun:
This uses the mock provider, so I moved the mock provider to a
location where the node test can use it.

### DIFF
--- a/node/lifecycle_test.go
+++ b/node/lifecycle_test.go
@@ -15,6 +15,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/watch"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
@@ -414,8 +415,51 @@ func testCreateStartDeleteScenario(ctx context.Context, t *testing.T, s *system)
 	}
 }
 
-func newPod() *corev1.Pod {
-	return &corev1.Pod{
+func BenchmarkCreatePods(b *testing.B) {
+	sl := logrus.StandardLogger()
+	sl.SetLevel(logrus.ErrorLevel)
+	newLogger := logruslogger.FromLogrus(logrus.NewEntry(sl))
+
+	ctx := context.Background()
+	ctx = log.WithLogger(ctx, newLogger)
+
+	assert.NilError(b, wireUpSystem(ctx, newMockProvider(), func(ctx context.Context, s *system) {
+		benchmarkCreatePods(ctx, b, s)
+	}))
+}
+
+func benchmarkCreatePods(ctx context.Context, b *testing.B, s *system) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	errCh := s.start(ctx)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		pod := newPod(randomizeUID, randomizeName)
+		_, err := s.client.CoreV1().Pods(pod.Namespace).Create(pod)
+		assert.NilError(b, err)
+		select {
+		case err = <-errCh:
+			b.Fatalf("Benchmark terminated with error: %+v", err)
+		default:
+		}
+	}
+}
+
+type podModifier func(*corev1.Pod)
+
+func randomizeUID(pod *corev1.Pod) {
+	pod.ObjectMeta.UID = uuid.NewUUID()
+}
+
+func randomizeName(pod *corev1.Pod) {
+	name := fmt.Sprintf("pod-%s", uuid.NewUUID())
+	pod.Name = name
+}
+
+func newPod(podmodifiers ...podModifier) *corev1.Pod {
+	pod := &corev1.Pod{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Pod",
 			APIVersion: "v1",
@@ -433,4 +477,8 @@ func newPod() *corev1.Pod {
 			Phase: corev1.PodPending,
 		},
 	}
+	for _, modifier := range podmodifiers {
+		modifier(pod)
+	}
+	return pod
 }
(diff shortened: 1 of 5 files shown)


sargun:
I'm beginning on adding this kind of test to be able to verify VK behaviour, ideally without the e2e tests, to get some baseline "spec" of the behaviour of the VK. 

More than a straight code review, I'm interested in what you think of this approach to testing.

node/lifecycle_test.go
```diff
@@ -0,0 +1,211 @@
+package node
+
+// Right now this test lives here because it needs the mock provider, which is an internal module.
+// If we move it out of this part of the tree, we can also moves this test.
```
gabrielhartmann: ```... we can also _move_ this test```

node/lifecycle_test.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+func init() {
+	klog.InitFlags(nil)
+}
+
+// fakeDiscardingRecorder discards all events. Silently.
+type fakeDiscardingRecorder struct {
+}
+
```
gabrielhartmann: Might be nice to log the NOOP function calls below.

node/lifecycle_test.go
```diff
@@ -0,0 +1,211 @@
...
+				return pod.Status.Phase == corev1.PodRunning, nil
+			})
+
+		watchErrCh <- watchErr
+	}()
+
+	select {
+	case <-ctx.Done():
+		t.Fatal("Context ended early")
+	case err = <-podControllerErrCh:
```
gabrielhartmann: Why are there 2 success cases here?

node/lifecycle_test.go
```diff
@@ -0,0 +1,211 @@
...
+		assert.NilError(t, err)
+	case err = <-watchErrCh:
+		assert.NilError(t, err)
+
+	}
+
+	watcher, err = client.CoreV1().Pods(testNamespace).Watch(listOptions)
```
gabrielhartmann: Is there a race here with the watcher?  Can we miss the deleted event or do events wait for watchers?

node/lifecycle_test.go
```diff
@@ -0,0 +1,211 @@
...
+		PodClient:         client.CoreV1(),
+		PodInformer:       podInformer,
+		EventRecorder:     fakeRecorder,
+		Provider:          mockProvider,
+		ConfigMapInformer: configMapInformer,
+		SecretInformer:    secretInformer,
+		ServiceInformer:   serviceInformer,
+	}
+
+	pc, err := NewPodController(config)
```
gabrielhartmann: From about here up, we're mostly setting up.  It would be nice factor out of the test a lot of the setup.

node/lifecycle_test.go
```diff
@@ -0,0 +1,211 @@
...
+				return pod.Name == p.ObjectMeta.Name, nil
+			})
+
+		watchErrCh <- watchErr
+	}()
+
+	go podInformerFactory.Start(ctx.Done())
+	go scmInformerFactory.Start(ctx.Done())
+
+	_, e := client.CoreV1().Pods(testNamespace).Create(&p)
```
gabrielhartmann: Please add comments for the major test sections / assertions.  e.g. `// CREATE POD`

node/lifecycle_test.go
sargun: That's the idea once we get enough feedback about the test. I plan on adding more tests with this approach shortly.

cpuguy83:
Are you looking for quicker turn-around for testing than the current e2e suite can provide?

In general it seems fine, there seems to be some(?) overlap with what's in pod_test.go.

sargun:
The e2e test has a lot more breadth (and is slow). I'm trying to create a minimal test which will be able to test multiple kinds of flows. For example, if you find this test approach "okay", then I want to add things like the dangling pods test.

node/lifecycle_test.go
```diff
@@ -0,0 +1,280 @@
...
+		"message":     fmt.Sprintf(messageFmt, args...),
+	}).Infof("Received annotated event")
+}
+func TestPodLifecycle(t *testing.T) {
+	// We don't do the defer cancel() thing here because t.Run is non-blocking, so the parent context may be cancelled
+	// before the children are finished and there is no way to do a "join" and wait for them without using a waitgroup,
+	// at which point, it doesn't seem much better.
+	ctx := context.Background()
+
+	newLogger := logruslogger.FromLogrus(logrus.NewEntry(logrus.StandardLogger()))
```
cpuguy83: Could make a logger for `t.Log`.

sargun: I find it confusing because if you try to compare klog vs. t.log messages they're in a different order.

cpuguy83: Yep. I remember I tried this now.

node/lifecycle_test.go
pires: Aren't these failure cases?

sargun: Yeah, they should always return errors so the assertion will faill.

node/lifecycle_test.go
```diff
@@ -414,8 +415,51 @@ func testCreateStartDeleteScenario(ctx context.Context, t *testing.T, s *system)
 	}
 }
 
-func newPod() *corev1.Pod {
-	return &corev1.Pod{
+func BenchmarkCreatePods(b *testing.B) {
```
pires: What is the goal of adding a benchmark?

node/lifecycle_test.go
pires: Events are delivered to watchers. If no watcher or watcher loses the event, a lister is to be used to fetch latest state. Therefore, events do not demand delivery guarantees and controllers should account for that.

node/lifecycle_test.go
sargun: It's mostly for memory allocations, and not for throughput. When you pass `-benchmem`, you can at least find out how many allocs are required per pod?

pires: Right, but what's the plan here? Is it to keep track of the allocations and signal when we increase/decrease number of allocs? I'm fine with any tests and benchmarks but I'd like to understand what the project will be gaining from this being introduced.

sargun: The discussion I've been having with @cpuguy83 is that we should set a scalability target. The intent of this test is to help us verify we can hit that scalability number. The one we decided on was 100k pods @ some known amount of memory. The purpose of this benchmark is to help us identify that number. 

Prior to any large changes, people should post before / after numbers and ensure we don't breach our #.

gabrielhartmann:
LGTM

node/lifecycle_test.go
pires: That's enough for me, thanks. Now, I wonder if the fact the test uses a fake client and an _empty_ control-plane has any impact - I think it does.

PR #708, opened 2019-07-25
Add a little bit of documentation to NotifyPods

sargun:
As far as I can tell, based on the implementation in MockProvider
NotifyPods is called with the mutated pod. This allows us to
take a copy of the Pod object in NotifyPods, and make it so
(eventually) we don't need to do a callback to GetPodStatus.

### DIFF
--- a/node/podcontroller.go
+++ b/node/podcontroller.go
@@ -69,9 +69,13 @@ type PodLifecycleHandler interface {
 // of pod status updates asynchronously.
 type PodNotifier interface {
 	// NotifyPods instructs the notifier to call the passed in function when
-	// the pod status changes.
+	// the pod status changes. It should be called when a pod's status changes.
 	//
-	// NotifyPods should not block callers.
+	// The provided pointer to a Pod is guaranteed to be used in a read-only
+	// fashion. The provided pod's PodStatus should be up to date when
+	// this function is called.
+	//
+	// NotifyPods will not block callers.
 	NotifyPods(context.Context, func(*corev1.Pod))
 }
 


sargun:
@pires @cpuguy83 This is a "simple" clarification. Let me know if it's (in)accurate. I'd like to add it for our own interest as well, because it will prevent us from breaking this rule in the future.

node/podcontroller.go
```diff
@@ -69,7 +69,13 @@ type PodLifecycleHandler interface {
 // of pod status updates asynchronously.
 type PodNotifier interface {
 	// NotifyPods instructs the notifier to call the passed in function when
-	// the pod status changes.
+	// the pod status changes. It should be called pod mutated with the new
```
pires: I can't follow what this means. Maybe you ate some words or something?

node/podcontroller.go
```diff
@@ -69,7 +69,13 @@ type PodLifecycleHandler interface {
...
+	// pod status.  It may be a pointer to the provider's version of the pod,
+	// and the pointer will not be dereferenced outside of the synchronous
+	// NotifyPods function.
```
pires: How are these two statements related? Also, is there any other scenario where the argument is not a pointer to the provider's version of the Pod?

sargun: A couple things:
* The pod doesn't need to be cloned (DeepCopy) before being passed to notifyPods
* The pod's podstatus should be updated to reflect the podstatus that you want to use to send to API Server

Basically, the intent is not to have the provider just calling notifypods with the pod that got passed from CreatePod, and then only surface the podstatus in getpodstatus. It's really ugly to be honest, in GetPodStatus, GetPods, and NotifyPods, but it means we don't have to callback to GetPodStatus after NotifyPods and we can avoid a nasty race condition.

sargun: Perhaps, you can suggest better words?

node/podcontroller.go
gabrielhartmann:
```
It should be called when a pod's status changes.
```

node/podcontroller.go
gabrielhartmann:
```
The provided pointer to a Pod is guaranteed to be used in a read-only fashion.
```

node/podcontroller.go
```diff
@@ -69,7 +69,13 @@ type PodLifecycleHandler interface {
...
+	//
+	// In addition, it is not necessary to preserve the ResourceVersion
+	// from the last imperative call.
```
gabrielhartmann: This is confusing and can probably be dropped without loss of clarity.

node/podcontroller.go
```diff
@@ -69,9 +69,13 @@ type PodLifecycleHandler interface {
 // of pod status updates asynchronously.
 type PodNotifier interface {
 	// NotifyPods instructs the notifier to call the passed in function when
-	// the pod status changes.
+	// the pod status changes. It should be called when a pod's status changes.
 	//
-	// NotifyPods should not block callers.
+	// The provided pointer to a Pod is guaranteed to be used in a read-only
+	// fashion. The provided pod's PodStatus should be up to date when
+	// NotifyPods is called
```
pires:
```suggestion
	// this function is called.
```

sargun:
@pires I made the change you suggested.

pires:
@cpuguy83 @rbitia I can't re-trigger the failed job so I can't merge this.

sargun:
I can.

PR #710, opened 2019-07-26
Introduce PodLifecycleHandlerV0 interface

sargun:
Introduce PodLifecycleHandlerV0 interface
    
    This breaks out the PodLifecycleHandlerV0 interface. The PodNotifier
    interface gives a false sense of what we expect people to implement.
    
    In addition, it marks the PodLifecycleHandlerV0 as deprecated.
    
    Lastly, it breaks out the wrapper code. The benefit of breaking
    out the wrapper code is that now people can customize their
    reconcilation loop time based on their provider's capabilities.

### DIFF
--- a/node/legacy_wrapper.go
+++ b/node/legacy_wrapper.go
@@ -1,0 +1,79 @@
+package node
+
+import (
+	"context"
+	"time"
+
+	pkgerrors "github.com/pkg/errors"
+	"github.com/virtual-kubelet/virtual-kubelet/log"
+	"github.com/virtual-kubelet/virtual-kubelet/trace"
+	v1 "k8s.io/api/core/v1"
+)
+
+type legacyPodLifecycleHandlerWrapper struct {
+	PodLifecycleHandlerV0
+	notifier      func(*v1.Pod)
+	reconcileTime time.Duration
+}
+
+func (wrapper *legacyPodLifecycleHandlerWrapper) NotifyPods(ctx context.Context, notifier func(*v1.Pod)) {
+	wrapper.notifier = notifier
+}
+
+// updatePodStatuses syncs the providers pod status with the kubernetes pod status.
+func (wrapper *legacyPodLifecycleHandlerWrapper) updatePodStatuses(ctx context.Context) {
+	ctx, span := trace.StartSpan(ctx, "updatePodStatuses")
+	defer span.End()
+
+	pods, err := wrapper.GetPods(ctx)
+	if err != nil {
+		err = pkgerrors.Wrap(err, "error getting pod list")
+		span.SetStatus(err)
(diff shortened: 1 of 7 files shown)


sargun:
I'd like to get your feedback on what you think of this. It's kind of two (intermingled) changes. They can be separated into two commits, but I think they would need to be part of the same PR.

cpuguy83:
Perhaps instead of calling it `legacy`, `v0` is more intentional and doesn't make it weird to embed?

sargun:
@pires @cpuguy83 This is ready for review.

node/legacy_wrapper.go
```diff
@@ -0,0 +1,79 @@
...
+	for {
+		t := time.NewTimer(wrapper.reconcileTime)
+		defer t.Stop()
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-t.C:
+				t.Stop()
```
cpuguy83: Timer is already stopped here

node/legacy_wrapper.go
```diff
@@ -0,0 +1,79 @@
...
+		wrapper.notifier(pod)
+	}
+}
+
+func (wrapper *legacyPodLifecycleHandlerWrapper) run(ctx context.Context) {
+	for {
+		t := time.NewTimer(wrapper.reconcileTime)
+		defer t.Stop()
+
+		for {
```
cpuguy83: What's the purpose of this double loop?

sargun: Errr, this was from a change before I made run() a private function.

node/legacy_wrapper.go
sargun: Same here. remenant of an earlier generation.

sargun:
I think it's better to put run() in a hidden interface and have the pod controller start it up. Originally, I had the user start it up, but I ran into a couple issues:
1. If NotifyPods started before the notifier was, this led to some messiness that then required synchronization. 
2. There was a . race . condition I exposed where if you started calling GetPods + NotifyPods when deleteDanglingPods was doing its thing, you could easily get into a race. I don't exactly remember what triggered the race detector, but I can dig it up.


I strongly prefer #2, and suggest moving this package out. We can solve the notifypods race problem by just adding a light bit of synchronization, and changing the code in the pod controller to only set notifyPods *after* deleteDanglingPods happens.

sargun:
The only downside I see of doing it in a new package is if we still keep tests in node/, they wont be able to use the wrapper, because the wrapper will depend on node/.

cpuguy83:
I do not like that this is exposing new interfaces which have unexported types and is generally unusable externally.
So I would like to handle the wrapping without exposing new things.

Changing the Handler interface really is a separate discussion, but I'm not sure what the benefit of doing it is.

PR #711, opened 2019-07-26
Setup event handler at Pod Controller creation time

sargun:
This seems to avoid a race conditions where at pod informer
startup time, the reactor doesn't properly get setup.

It also refactors the root command example to start up
the informers after everything is wired up.


pires:
I wanted a review from @cpuguy83.

sargun:
Do you want me to revert?

cpuguy83:
Can you explain why moving the queue setup helps?

pires:
I think it was done for not initializing before other validations pass, such as resource manager, etc.

sargun:
When .Start() is called on the podinformerfactory, it calls .`.Run()` on the informer (`podsInformer.Informer()`). There is supposed to be code, that _after_ the podinformerfactory is . started, it will start the informers. 

For some reason, there is a race where the informer is not started if it is being created / added while the podinformerfactory is being started.

PR #716, opened 2019-08-05
Upgrade Kubernetes Libraries

sargun:
This upgrades the kubernetes libraries to those that are used in
1.15. The purpose of this is so we can use modern versions / features of those libraries while we test things. Also, it's hygenic in terms of support.


pires:
I don't know how I can validate these changes as very unfortunately the versions are not (easily) human-readable.
Anyway, shouldn't we then test these APIs by bumping version [here](https://github.com/virtual-kubelet/virtual-kubelet/blob/master/go.mod#L56), [here](https://github.com/virtual-kubelet/virtual-kubelet/blob/master/cmd/virtual-kubelet/main.go#L41) and [here](https://github.com/virtual-kubelet/virtual-kubelet/blob/master/.circleci/config.yml#L48)?

Also, tests wouldn't pass if we had leases enabled by default because coordination API group is alpha in Kubernetes 1.13.

sargun:
There are bugs I'm running into with fake client-go with our current version of the k8s libraries. I can bump those references, but is it really neccessary?

pires:
Honestly, I'd bump everything, including the Kubernetes version we use for e2e, to latest 1.15. After all, when 1.16 is released, 1.13 upstream support will be no more.

I understand this was not your intention when you started the PR, but why not do it?

sargun:
@pires There's a bunch of small interface stuff (switching from *Beta to non-beta), but I do not think we should change those yet. Let me know if this works for you.

sargun:
I also added the commands I ran to the commit. The replaces are copy & paste from the kubernetes go.mod.

sargun:
I think we're hitting a build tagging issue. On my mac, it wants to include this prom dep. Whereas remotely (on Linux) it doesn't. What do y'all think about making it so it only fails on *missing* deps, as opposed to "too many" deps?

pires:
I suggest we simply remove this step from CI. @cpuguy83 ?

sargun:
@pires @cpuguy83 Bar the go.mod issue, what's your opinion on the rest?

sargun:
I'm just going to roll this into https://github.com/virtual-kubelet/virtual-kubelet/pull/718

PR #718, opened 2019-08-07
Fixup race conditions, and lifecycle tests

sargun:
* Fix the deletion test to actually test the pod is deleted
 * Fix the update pods test to update a value which is allowed
   to be updated


sargun:
I think we could probably get rid of the deletion test function in `pod_test.go` after this.

node/lifecycle_test.go
```diff
@@ -90,11 +93,19 @@ func TestPodLifecycle(t *testing.T) {
 
 	ctx = log.WithLogger(ctx, log.L)
 
+	waitForPodToBePermanentlyDeletedDeletionFunc := func(ev watch.Event) (bool, error) {
```
gabrielhartmann: This function doesn't wait.
`isPodDeletedFunc`

node/lifecycle_test.go
```diff
@@ -103,13 +114,39 @@ func TestPodLifecycle(t *testing.T) {
...
+	t.Run("createStartDeleteScenarioWithDeletionErrorNotFound", func(t *testing.T) {
+		// Wait for the pod to be started
+		mp := newMockProvider()
+		mp.errorOnDelete = errdefs.NotFound("not found")
+		assert.NilError(t, wireUpSystem(ctx, mp, func(ctx context.Context, s *system) {
+			testCreateStartDeleteScenario(ctx, t, s, waitForPodToBePermanentlyDeletedDeletionFunc)
+		}))
+	})
+
+	t.Run("createStartDeleteScenarioWithDeletionRandomError", func(t *testing.T) {
```
gabrielhartmann: It's not obvious to me what this test is supposed to do.  Retry of delete should succeed?

node/lifecycle_test.go
```diff
@@ -103,13 +114,39 @@ func TestPodLifecycle(t *testing.T) {
...
+		mp := newMockProvider()
+		// Wait for the pod to be started
+		deletionFunc := func(ev watch.Event) (bool, error) {
+			for atomic.LoadUint64(&mp.attemptedDeletes) < 2 {
```
gabrielhartmann: 1.  This is a super hot loop, ok...
1.  If `attemptedDeletes` never reaches 2 we block forever?

Do these tests timeout?  A more informative test failure of "Didn't reach attempted deletes in X tries" is easier to debug later.

node/lifecycle_test.go
```diff
@@ -446,14 +474,15 @@ func testUpdatePodWhileRunningScenario(ctx context.Context, t *testing.T, s *sys
...
 	go func() {
-		_, watchErr := watchutils.UntilWithoutRetry(ctx, watcher,
+		newPod, watchErr := watchutils.UntilWithoutRetry(ctx, watcher,
 			// Wait for the pod to be started
 			func(ev watch.Event) (bool, error) {
 				pod := ev.Object.(*corev1.Pod)
 				return pod.Status.Phase == corev1.PodRunning, nil
 			})
-
+		p = newPod.Object.(*corev1.Pod).DeepCopy()
```
gabrielhartmann: Does DeepCopy grab a lock?

node/lifecycle_test.go
```diff
@@ -473,9 +502,16 @@ func testUpdatePodWhileRunningScenario(ctx context.Context, t *testing.T, s *sys
 	}
 
 	// Update the pod
+	version, err := strconv.Atoi(p.ResourceVersion)
+	if err != nil {
+		panic(err)
```
gabrielhartmann: Is a test assertion or `Fatal` error more appropriate here?

node/mock_test.go
```diff
@@ -157,7 +164,14 @@ func (p *mockV0Provider) DeletePod(ctx context.Context, pod *v1.Pod) (err error)
 	}
 
 	p.notifier(pod)
-
+	if p.realNotifier == nil {
+		// The pods reconciliation (GetPodStatus) should be called in under 1 minute
```
gabrielhartmann: This will get removed once the notifier uses the passed in pod instead of calling the provider again right?

node/pod.go
```diff
@@ -57,11 +56,16 @@ func (pc *PodController) createOrUpdatePod(ctx context.Context, pod *corev1.Pod)
...
 	})
 
+	pod = pod.DeepCopy()
 	if err := populateEnvironmentVariables(ctx, pod, pc.resourceManager, pc.recorder); err != nil {
 		span.SetStatus(err)
 		return err
 	}
 
+	// We have to use a  different pod that we pass to the provider than the one that gets used in handleProviderError
+	// because the provider may have manipulated the pod in a separate goroutine while we were doing work
```
gabrielhartmann:
```
... may have manipulated the pod in a separate goroutine while we were doing work
```
to
```
... may manipulate the pod 
```

node/lifecycle_test.go
sargun: It's supposed to ensure that the deletes are retried. I guess I could also check if the pod is (not) deleted in API server erroneously.

node/lifecycle_test.go
sargun: the tests run with a timeout. Unfortunately, the timeout is a panic.

node/lifecycle_test.go
sargun: No, it doesn't need to. The pod objects returned by the fake client-go  should be safe to *read* (but not neccessarily write) to.

node/mock_test.go
sargun: Kind of. When it's being run in "Legacy" mode, it'll still need this.

sargun: Part of me is almost like, maybe we should just actively desupport legacy. It's really not that hard to use the NotifyPods interface.

node/pod.go
sargun: fixed

sargun:
This is quickly becoming a "mega" PR. I'll wait for this to go through before proceeding with the work to:
* Make NotifyPods better
* Write the "official" wrapper for v0 providers.

node/lifecycle_test.go
pires: But why is this needed in the first place?

node/node_test.go
```diff
@@ -54,7 +54,9 @@ func testNodeRun(t *testing.T, enableLease bool) {
 	if enableLease {
 		opts = append(opts, WithNodeEnableLeaseV1Beta1(leases, nil))
 	}
-	node, err := NewNodeController(testP, testNode(t), nodes, opts...)
+	testNode := testNode(t)
+	testNodeCopy := testNode.DeepCopy()
```
pires: Why is this needed?

go.mod
```diff
@@ -38,26 +29,57 @@ require (
...
+
+replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.0.0-20190805143126-cdb999c96590
+
+replace k8s.io/metrics => k8s.io/metrics v0.0.0-20190805143318-16b07057415d
+
+replace k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.0.0-20190805142637-3b65bc4bb24f
+
+replace k8s.io/code-generator => k8s.io/code-generator v0.0.0-20190612205613-18da4a14b22b
+
+replace k8s.io/client-go => k8s.io/client-go v0.0.0-20190805141520-2fe0317bcee0
```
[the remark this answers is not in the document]
sargun: This was generated via:
```
go get k8s.io/kubernetes@v1.15.2 2> fail
for i in $(cat fail|grep unknown|cut -f1 -d@|cut -f2 -d" "); do echo "replace ${i} => ${i} kubernetes-1.15.2";done
```

For some reason, the go.mod in kubernetes has a bunch of these entries that depend on `v0.0.0`:  https://github.com/kubernetes/kubernetes/blob/v1.15.2/go.mod#L167-L191

node/lifecycle_test.go
```diff
@@ -369,39 +404,32 @@ func testCreateStartDeleteScenario(ctx context.Context, t *testing.T, s *system)
...
 		watchErrCh <- watchErr
 	}()
 
-	// Delete the pod
+	// Delete the pod via deletiontimestamp
 
 	// 1. Get the pod
 	currentPod, err := s.client.CoreV1().Pods(testNamespace).Get(p.Name, metav1.GetOptions{})
 	assert.NilError(t, err)
+	currentPod = currentPod.DeepCopy()
```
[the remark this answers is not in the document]
sargun: To ensure we're working on a local copy of the pod.

node/node_test.go
sargun: To ensure we're working on a local copy of the pod.

sargun: If we use . testNode directly, it's problematic because NewNodeController refers to the pointer, and modifies that object. This causes the race detector to get unhappy.

node/lifecycle_test.go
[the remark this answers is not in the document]
pires: But why is it needed here? Is the pod being changed concurrencially?

node/node_test.go
pires: Ah, so you're trying to please the race detector. A comment would have avoided this discussion, so I suggest adding said comment, please.

node/lifecycle_test.go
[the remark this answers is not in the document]
pires: Neververmind, I get it that you're trying to please the race detector. Same suggestion as below is applicable here too.

sargun:
@pires Comments added about pleasing the race detector.

node/lifecycle_test.go
gabrielhartmann: The name of the test doesn't say, "testDeleteRetry" nor does any comment.

node/lifecycle_test.go
gabrielhartmann: 🆒

node/mock_test.go
gabrielhartmann: I agree with you but probably not in this PR.

PR #724, opened 2019-08-12
Upgrade k8s

sargun:


sargun:
There's a couple things here:

* Go module support (in K8s natively)
* K8s 1.13 is going to go non-supported soonish
* It fixes some weird corner cases.

PR #726, opened 2019-08-12
Fix race conditions in node/pod.go

sargun:


node/pod.go
```diff
@@ -231,13 +236,20 @@ func (pc *PodController) updatePodStatus(ctx context.Context, pod *corev1.Pod) e
 		return pkgerrors.Wrap(err, "error retrieving pod status")
 	}
 
+	// Copy the status, so we do not reference it, in case the provider changes it
+	status = status.DeepCopy()
```
cpuguy83: Is this not racy in and of itself if the provider could change it?
We should document this in the interface.

sargun: This *should* also go away as part of NotifyPods.w

PR #728, opened 2019-08-13
Remove usage of atomics in tests

sargun:
It turns out that running atomic.Read(...) in a tight loop breaks
Golang. The goroutine would never yield control over the scheduler,
so we ended up getting into a situation where the test would get
stuck forever. This moves to a different model, in which
there is a condition var, instead of atomics in loops.

### DIFF
--- a/node/mock_test.go
+++ b/node/mock_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/virtual-kubelet/virtual-kubelet/errdefs"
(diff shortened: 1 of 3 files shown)


node/mock_test.go
```diff
@@ -18,11 +17,48 @@ var (
...
+	}
+}
+
+func (w *waitableInt) read() int {
+	defer w.cond.L.Unlock()
+	w.cond.L.Lock()
+	return w.val
+}
+
+func (w *waitableInt) until(f func(int) bool) {
```
cpuguy83: Can this leak goroutines in failure cases?
Seems like we may need a way to cancel this.

node/mock_test.go
```diff
@@ -18,11 +17,60 @@ var (
...
+	defer close(c)
+	go func() {
+		select {
+		case <-ctx.Done():
+			w.cond.Broadcast()
+		case <-c:
+		}
+	}()
+
+	for {
```
cpuguy83:
```go
w.cond.L.Lock()
defer w.cond.L.Unlock()

for !f(w.val) {
    if err := ctx.Err() != nil {
        return err
    }
    w.cond.Wait()
}
return nil
```

node/mock_test.go
```diff
@@ -18,11 +17,60 @@ var (
...
+}
+
+func (w *waitableInt) read() int {
+	defer w.cond.L.Unlock()
+	w.cond.L.Lock()
+	return w.val
+}
+
+func (w *waitableInt) until(ctx context.Context, f func(int) bool) error {
+	c := make(chan struct{})
```
cpuguy83: Maybe we can simplify this:

```go
ctx, cancel := context.WithCancel(ctx)
defer cancel()
go func() {
    <-ctx.Done()
    w.cond.Broadcast()
}()
```

node/mock_test.go
sargun: Won't that deadlock on itself? because when w.cond.Wait returns, it’ll lock w.cond.L. and on the 2nd loop, it’ll lock twice.

cpuguy83: Wait unlocks when it's called and re-locks when it returns.
Here's the official example:

>  Wait atomically unlocks c.L and suspends execution of the calling goroutine. After later resuming execution, Wait locks c.L before returning. Unlike in other systems, Wait cannot return unless awoken by Broadcast or Signal.
>
> Because c.L is not locked when Wait first resumes, the caller typically cannot assume that the condition is true when Wait returns. Instead, the caller should Wait in a loop: 
```
c.L.Lock()
for !condition() {
    c.Wait()
}
... make use of condition ...
c.L.Unlock()
```

sargun: nevermind

PR #732, opened 2019-08-13
Move location of eventhandler registration

sargun:
This moves the event handler registration until after the cache
is in-sync.

It makes it so we can use the log object from the context,
rather than having to use the global logger

The cache race condition of the cache starting while the reactor
is being added wont exist because we wait for the cache
to startup / go in sync prior to adding it.

### DIFF
--- a/node/podcontroller.go
+++ b/node/podcontroller.go
@@ -173,11 +173,31 @@ func NewPodController(cfg PodControllerConfig) (*PodController, error) {
 		k8sQ:            workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "syncPodsFromKubernetes"),
 	}
 
-	// Set up event handlers for when Pod resources change.
+	return pc, nil
+}
+
+// Run will set up the event handlers for types we are interested in, as well as syncing informer caches and starting workers.
+// It will block until the context is cancelled, at which point it will shutdown the work queue and wait for workers to finish processing their current work items.
+func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) error {
+	defer pc.k8sQ.ShutDown()
+
+	podStatusQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "syncPodStatusFromProvider")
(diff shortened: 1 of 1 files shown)


cpuguy83:
Weren't these recently moved out of Run?

sargun:
Yeah, I didn't realize I could move it below sync. I just had to make it so it didn't run concurrently with start.

node/podcontroller.go
```diff
@@ -161,14 +161,35 @@ func NewPodController(cfg PodControllerConfig) (*PodController, error) {
...
+// It will block until the context is cancelled, at which point it will shutdown the work queue and wait for workers to finish processing their current work items.
+func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) error {
+	defer pc.k8sQ.ShutDown()
+
+	podStatusQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "syncPodStatusFromProvider")
+	pc.runSyncFromProvider(ctx, podStatusQueue)
+	pc.runProviderSyncWorkers(ctx, podStatusQueue, podSyncWorkers)
+	defer podStatusQueue.ShutDown()
+
+	// Wait for the caches to be synced *before* starting workers.
```
cpuguy83: "Workers" are started, we just haven't added event handlers.

sargun: Let me rearrange this code to reflect the truth. I don’t actually see any downside of starting the workers for k8s->provider. The other way might be problematic though.

node/podcontroller.go
```diff
@@ -161,14 +161,35 @@ func NewPodController(cfg PodControllerConfig) (*PodController, error) {
 		resourceManager: rm,
 		ready:           make(chan struct{}),
 		recorder:        cfg.EventRecorder,
-		k8sQ:            workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "syncPodsFromKubernetes"),
+
+		k8sQ: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "syncPodsFromKubernetes"),
```
gabrielhartmann: weird formatting

PR #733, opened 2019-08-13
Add a generic reactor to lifecycle_test to bump resource version

sargun:
All updates in our tests should have the behaviour that best
reflects what API server does.


cpuguy83:
@pires PTAL

PR #734, opened 2019-08-13
Do not mutate pods, nor hand off pod references to provider

sargun:
This moves to a model where any time that pods are given to a
provider, it uses a DeepCopy, as opposed to a reference. If the
provider mutates the pod, it prevents it from causing issues
with the informer cache.


sargun:
@jeremyrickard it looks like spew vs reflect.deepequal return different results in regards to nil values.

cpuguy83:
PTAL at https://godoc.org/github.com/google/go-cmp/cmp instead of reflect.

sargun:
Validated with:

```
go test -c -gcflags="-N -l" -race  ./node/ && KUBE_CACHE_MUTATION_DETECTOR=true ./node.test -test.count=10 -test.run='TestPodLifecycle/.*/mockProvider' -test.v  2>&1|tee test.log
```

node/pod.go
```diff
@@ -57,32 +58,44 @@ func (pc *PodController) createOrUpdatePod(ctx context.Context, pod *corev1.Pod)
...
 		// - `spec.activeDeadlineSeconds`
 		// - `spec.tolerations` (only additions to existing tolerations)
+		// - `objectmeta.labels`
+		// - `objectmeta.annotations`
 		// compare the hashes of the pod specs to see if the specs actually changed
-		expected := hashPodSpec(pp.Spec)
-		if actual := hashPodSpec(pod.Spec); actual != expected {
-			log.G(ctx).Debugf("Pod %s exists, updating pod in provider", pp.Name)
-			if origErr := pc.provider.UpdatePod(ctx, pod); origErr != nil {
+		if !cmp.Equal(podFromProvider.Spec.Containers, podForProvider.Spec.Containers) ||
```
cpuguy83: Can this all be moved into a helper?
How come we are looking at specific fields like this?

sargun: @jeremyrickard Why did you initially do it like this, was it just performance?

PR #735, opened 2019-08-13
Add methods to wait for pod controller exit and errors

cpuguy83:
Adds Done() and Err() to pod controller

    Allows callers to wait for pod controller exit in addition to readiness.
    This means the caller does not have to deal handling errors from the pod
    controller running in a gorutine since it can wait for exit via `Done()`
    and check the error with `Err()`

### DIFF
--- a/cmd/virtual-kubelet/internal/commands/root/root.go
+++ b/cmd/virtual-kubelet/internal/commands/root/root.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"os"
 	"path"
-	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -207,11 +206,16 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
 	}()
 
 	if c.StartupTimeout > 0 {
-		// If there is a startup timeout, it does two things:
-		// 1. It causes the VK to shutdown if we haven't gotten into an operational state in a time period
-		// 2. It prevents node advertisement from happening until we're in an operational state
-		err = waitFor(ctx, c.StartupTimeout, pc.Ready())
-		if err != nil {
+		ctx, cancel := context.WithTimeout(ctx, c.StartupTimeout)
+		log.G(ctx).Info("Waiting for pod controller / VK to be ready")
+		select {
+		case <-ctx.Done():
+			cancel()
+			return ctx.Err()
+		case <-pc.Ready():
+		}
+		cancel()
+		if err := pc.Err(); err != nil {
 			return err
 		}
 	}
@@ -226,21 +230,6 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
 
 	<-ctx.Done()
 	return nil
-}
-
-func waitFor(ctx context.Context, time time.Duration, ready <-chan struct{}) error {
-	ctx, cancel := context.WithTimeout(ctx, time)
-	defer cancel()
-
-	// Wait for the VK / PC close the the ready channel, or time out and return
-	log.G(ctx).Info("Waiting for pod controller / VK to be ready")
-
-	select {
-	case <-ready:
-		return nil
-	case <-ctx.Done():
-		return errors.Wrap(ctx.Err(), "Error while starting up VK")
-	}
 }
 
 func newClient(configPath string) (*kubernetes.Clientset, error) {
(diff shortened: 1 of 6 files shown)


node/podcontroller.go
```diff
@@ -255,10 +279,31 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) error {
...
 }
 
+// WaitReady waits for the pod controller to be ready.
+// As a general rule of thumb, this function will return if `Run` has returned
+// OR the controller is ready.
+//
+// This should return the same error that `Run()` would return, however it may
+// differ especially if the context (either in Run or passed in here) is
+// cancelled.
+func (pc *PodController) WaitReady(ctx context.Context) error {
```
sargun: Should there be a way to retrieve the error after shutdown? Or wait for the controller to shutdown?

node/lifecycle_test.go
```diff
@@ -309,7 +289,7 @@ func wireUpSystem(ctx context.Context, provider PodLifecycleHandler, f testFunct
 
 	// Shutdown the pod controller, and wait for it to exit
 	cancel()
-	return <-sys.retChan
+	return nil
```
sargun: Aren't we missing out on checking that the pod controller had no errors at shutdown? (and shutdown successfully?)

cpuguy83: Right now there's nothing to wait on.
 We can certainly do that... e.g. taking from context.Context we could impement `<-pc.Done()` and `pc.Err()`.
However if we do this we should wait on workers and everything to stop before returning.
WDYT?

sargun: I think it makes sense to. I had a PR that put the workers in an errgroup. Perhaps you can edd the API for now, and then we can actually plumb it in later.

node/podcontroller.go
cpuguy83: As discussed, I added an `Err()` and `Done()` function.

cpuguy83: Do we need this function even with the other new functions?

sargun: I don't think so, and FWIW, I think you need to take `pc.mu` during this function.

node/lifecycle_test.go
```diff
@@ -237,33 +237,20 @@ func TestPodLifecycle(t *testing.T) {
...
 	case <-s.pc.Ready():
-		// This listens for errors, or exits in the future.
-		go func() {
-			podControllerErr := <-podControllerErrChan
-			s.retChan <- podControllerErr
-		}()
-	// If there is an error before things are ready, we need to forward it immediately
-	case podControllerErr := <-podControllerErrChan:
-		s.retChan <- podControllerErr
+	case <-s.pc.Done():
```
sargun: Shouldn't this return s.pc.Err()?

cpuguy83: It does at the end of the function

cmd/virtual-kubelet/internal/commands/root/root.go
```diff
@@ -207,11 +206,15 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
...
-		err = waitFor(ctx, c.StartupTimeout, pc.Ready())
-		if err != nil {
+		ctx, cancel := context.WithTimeout(ctx, c.StartupTimeout)
+		log.G(ctx).Info("Waiting for pod controller / VK to be ready")
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-pc.Ready():
+		}
+		cancel()
```
pires: Shouldn't this be deferred?

sargun: Isn't this the context just for startup, and not the whole lifecycle?

cpuguy83: Right, though it was missing a cancel in the `.Done()` branch.

sargun: if ctx.Done() returns, doesn't that mean it's cancelled / done?

cpuguy83: Generally yes, but not if the parent is the thing that was cancelled.

cpuguy83: And yes, that through me too :)
Linter picked it up.

PR #736, opened 2019-08-14
Wait for the informer to become in sync before starting tests

sargun:
If the informers are starting at the same time as createPods,
then we can get into a situation where the pod seems to get
"lost". Instead, we wait for the informer to get into sync
prior to the createpod event.

This also moves to one informer as a microoptimization in
the tests.


sargun:
successfully ran:
```
go test -c -gcflags="-N -l"  ./node/ 
./node.test -test.count=100 -test.run='TestPodLifecycle/.*/mockProvider' -test.v
```

I can't run with `-race`, until https://github.com/virtual-kubelet/virtual-kubelet/pull/734 is in.

PR #740, opened 2019-08-18
Wait for Workers to exit prior to returning from PodController.Run

sargun:
This changes the behaviour slightly, so rather than immediately exiting on
context cancellation, this calls shutdown, and waits for the current
items to finish being worked on before returning to the user.


sargun:
I'm waiting on #735 to merge. This will need a rebase and some rework, I think.

pires:
Ah, I had gotten it the other way around. Thanks!

PR #743, opened 2019-08-19
Add unit tests for #584

chewong:
This PR adds unit tests for changes made in #584.  It also fixes two small typos in the codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ type PodLifecycleHandler interface {
 ```
 
 There is also an optional interface `PodNotifier` which enables the provider to
-asyncronously notify the virtual-kubelet about pod status changes. If this
+asynchronously notify the virtual-kubelet about pod status changes. If this
 interface is not implemented, virtual-kubelet will periodically check the status
 of all pods.
 
--- a/node/pod_test.go
+++ b/node/pod_test.go
@@ -213,7 +213,7 @@ func TestPodDelete(t *testing.T) {
 	}
 }
 
-func TestPodUpdateStatus(t *testing.T) {
+func TestFetchPodStatusFromProvider(t *testing.T) {
 	startedAt := metav1.NewTime(time.Now())
 	finishedAt := metav1.NewTime(startedAt.Add(time.Second * 10))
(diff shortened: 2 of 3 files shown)


chewong:
I signed it

node/pod.go
```diff
@@ -267,15 +267,28 @@ func (pc *PodController) updatePodStatus(ctx context.Context, pod *corev1.Pod) e
 			pod.Status.Reason = "NotFound"
 			pod.Status.Message = "The pod status was not found and may have been deleted from the provider"
 			for i, c := range pod.Status.ContainerStatuses {
+				var startedAt metav1.Time
+				var fininshedAt metav1.Time
+				if c.State.Waiting != nil {
+					// Skip waiting state
+					continue
```
sargun: Doesn't this result in a behaviour change for waiting containers? Also, what's going on that results in this crash?

chewong: #583 describes the crash in detail. `c.State.Running` is `nil` when one of the containers is in waiting state, causing a crash when VK tries to access `c.State.Running.StartedAt`. What type of behavior change are you expecting from this change?

sargun:
I'm wondering if this function makes sense to keep longer term. Will this continue to exist if we move everything to NotifyPods?

node/pod.go
```diff
@@ -267,15 +267,28 @@ func (pc *PodController) updatePodStatus(ctx context.Context, pod *corev1.Pod) e
...
+				} else if c.State.Terminated != nil {
+					startedAt = c.State.Terminated.StartedAt
+					fininshedAt = c.State.Terminated.FinishedAt
+				} else {
+					startedAt = c.State.Running.StartedAt
+					fininshedAt = metav1.NewTime(time.Now())
```
sargun: I suggest capturing now outside the loop.

sargun:
@chewong I think https://github.com/virtual-kubelet/virtual-kubelet/pull/751 obsoletes this.  Depending on how we proceed, that might make more sense.

sargun:
Can you add a lifecycle test for this?

PR #749, opened 2019-08-22
Notify Pods Based Pod Status Updates [WIP]

sargun:


sargun:
#750 is an alternative to this approach. Let me know which one you prefer.

sargun:
@cpuguy83 If you want, I can split this into two PRs. One, with notifyPods stuffing pod statuses into a map, and the second one around cleaner deletes.

sargun:
I split out some part of this PR here: https://github.com/virtual-kubelet/virtual-kubelet/pull/751. Let's get that merged first.

PR #751, opened 2019-08-25
Have NotifyPods store the pod status in a map

sargun:
We introduce a map that can be used to store the pod status. In this,
we do not need to call GetPodStatus immediately after NotifyPods
is called. Instead, we stash the pod passed via notifypods
as in a map we can access later. In addition to this, for legacy
providers, the logic to merge the pod, and the pod status is
hoisted up to the loop.

It prevents leaks by deleting the entry in the map as soon
as the pod is deleted from k8s.

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -209,15 +209,15 @@ func (pc *PodController) forceDeletePodResource(ctx context.Context, namespace, 
 	return nil
 }
 
-// updatePodStatuses syncs the providers pod status with the kubernetes pod status.
-func (pc *PodController) updatePodStatuses(ctx context.Context, q workqueue.RateLimitingInterface) {
-	ctx, span := trace.StartSpan(ctx, "updatePodStatuses")
+// fetchPodStatusesFromProvider syncs the providers pod status with the kubernetes pod status.
+func (pc *PodController) fetchPodStatusesFromProvider(ctx context.Context, q workqueue.RateLimitingInterface) {
+	ctx, span := trace.StartSpan(ctx, "fetchPodStatusesFromProvider")
 	defer span.End()
 
 	// Update all the pods with the provider status.
 	pods, err := pc.podsLister.List(labels.Everything())
 	if err != nil {
-		err = pkgerrors.Wrap(err, "error getting pod list")
+		err = pkgerrors.Wrap(err, "error getting pod list from kubernetes")
 		span.SetStatus(err)
 		log.G(ctx).WithError(err).Error("Error updating pod statuses")
 		return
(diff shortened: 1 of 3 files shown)


node/lifecycle_test.go
```diff
@@ -23,11 +23,11 @@ import (
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	ktesting "k8s.io/client-go/testing"
```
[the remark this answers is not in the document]
cpuguy83: ?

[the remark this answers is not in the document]
sargun: It should have errored on this being out of order earlier.

node/queue.go
```diff
@@ -146,7 +146,7 @@ func (pc *PodController) providerSyncLoop(ctx context.Context, q workqueue.RateL
 func (pc *PodController) runSyncFromProvider(ctx context.Context, q workqueue.RateLimitingInterface) {
 	if pn, ok := pc.provider.(PodNotifier); ok {
 		pn.NotifyPods(ctx, func(pod *corev1.Pod) {
-			enqueuePodStatusUpdate(ctx, q, pod)
+			pc.enqueuePodStatusUpdate(ctx, q, pod.DeepCopy())
```
cpuguy83: We end up calling `DeepCopy` in eneuquePodStatusUpdate as well.

sargun: I moved to doing the `DeepCopy` before `enqueuePodStatusUpdate`

node/lifecycle_test.go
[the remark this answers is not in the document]
sargun: Fixed in https://github.com/virtual-kubelet/virtual-kubelet/pull/755

node/pod.go
```diff
@@ -226,78 +226,104 @@ func (pc *PodController) updatePodStatuses(ctx context.Context, q workqueue.Rate
 
 	for _, pod := range pods {
 		if !shouldSkipPodStatusUpdate(pod) {
-			enqueuePodStatusUpdate(ctx, q, pod)
+			pc.fetchPodStatusFromProvider(ctx, q, pod)
```
pires: I prefer keeping "enqueue" call here and have pod status from provider fetching happening in a new line just before calling enqueue, effectively meaning two calls in this block.
The reason is that by reading this code, I now have to forcibly read `fetchPodStatusFromProvider` to understand that we may be interacting with the queue.

node/pod.go
```diff
@@ -226,78 +226,104 @@ func (pc *PodController) updatePodStatuses(ctx context.Context, q workqueue.Rate
...
+			podStatus.Phase = corev1.PodFailed
+			podStatus.Reason = "NotFound"
+			podStatus.Message = "The pod status was not found and may have been deleted from the provider"
+			now := metav1.NewTime(time.Now())
+			for i, c := range podStatus.ContainerStatuses {
+				if c.State.Running == nil {
+					continue
+				}
+				podStatus.ContainerStatuses[i].State.Terminated = &corev1.ContainerStateTerminated{
+					ExitCode:    -137,
```
pires: For increased readability, I suggest setting a const for this.

node/pod.go
```diff
@@ -226,78 +226,104 @@ func (pc *PodController) updatePodStatuses(ctx context.Context, q workqueue.Rate
...
+			podStatus = podFromKubernetes.Status.DeepCopy()
+			podStatus.Phase = corev1.PodFailed
+			podStatus.Reason = "NotFound"
+			podStatus.Message = "The pod status was not found and may have been deleted from the provider"
+			now := metav1.NewTime(time.Now())
+			for i, c := range podStatus.ContainerStatuses {
+				if c.State.Running == nil {
+					continue
+				}
+				podStatus.ContainerStatuses[i].State.Terminated = &corev1.ContainerStateTerminated{
```
pires: So if a provider sets a pod's `ContainerStatuses`, VK is overwriting it? I know this hasn't changed behavior in `master` but I'm wondering if we overlooked this in the past.

node/podcontroller.go
```diff
@@ -109,6 +109,17 @@ type PodController struct {
 	resourceManager *manager.ResourceManager
 
 	k8sQ workqueue.RateLimitingInterface
+
+	// From the time of creation, to termination the knownPods map will contain the pods key
+	// (derived from K8's cache library) -> a *knownPod struct.
```
pires: It's "K8s" not "K8's" but I suggest writing Kubernetes as you do in other parts of the PR.

node/pod.go
sargun: I don't want to change this in this PR. I'll do it in a follow-up PR, given this code is a lift-and-shift.

node/pod.go
sargun: Correct. This is a lift-and-shift.

I think @chewong has some work in this area he's doing. 

Long-term, I suggest we do not change this code, because it will / should go away.

pires: Maybe just add a TODO?

node/pod.go
pires: Maybe just add a TODO?

node/pod.go
sargun: Wait, thinking back, should we change this behaviour? If a container is suddenly not found, should we change that? 

Do you want to file an issue discussing your concern, and we can figure out what to do from there?

node/pod.go
sargun: Split.

node/pod.go
sargun: It's on my repo, on the branch `test-pod-going-missing-while-running`.

node/pod.go
pires: Yes, I'll start an issue.

PR #760, opened 2019-09-04
Do not delete pods in a non-graceful manner

sargun:
This moves from forcefully deleting pods to deleting pods in a
graceful manner. It keeps track of the pods that were deleted,
so the provider doesn't get duplicate pod deletions unnecessarily.

It only retries deletion if the pod's graceperiod was reduced.

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	pkgerrors "github.com/pkg/errors"
-	"github.com/virtual-kubelet/virtual-kubelet/errdefs"
 	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"github.com/virtual-kubelet/virtual-kubelet/trace"
 	corev1 "k8s.io/api/core/v1"
@@ -133,61 +132,19 @@ func (pc *PodController) handleProviderError(ctx context.Context, span trace.Spa
 	span.SetStatus(origErr)
 }
 
-func (pc *PodController) deletePod(ctx context.Context, namespace, name string) error {
+func (pc *PodController) deletePod(ctx context.Context, pod *corev1.Pod) error {
 	ctx, span := trace.StartSpan(ctx, "deletePod")
 	defer span.End()
-
-	pod, err := pc.provider.GetPod(ctx, namespace, name)
-	if err != nil {
-		if errdefs.IsNotFound(err) {
-			// The provider is not aware of the pod, but we must still delete the Kubernetes API resource.
-			return pc.forceDeletePodResource(ctx, namespace, name)
-		}
-		return err
-	}
-
-	// NOTE: Some providers return a non-nil error in their GetPod implementation when the pod is not found while some other don't.
-	if pod == nil {
-		// The provider is not aware of the pod, but we must still delete the Kubernetes API resource.
-		return pc.forceDeletePodResource(ctx, namespace, name)
-	}
-
 	ctx = addPodAttributes(ctx, span, pod)
 
-	var delErr error
-	if delErr = pc.provider.DeletePod(ctx, pod.DeepCopy()); delErr != nil && !errdefs.IsNotFound(delErr) {
-		span.SetStatus(delErr)
-		return delErr
+	err := pc.provider.DeletePod(ctx, pod.DeepCopy())
+	if err != nil {
+		span.SetStatus(err)
+		return err
 	}
 
 	log.G(ctx).Debug("Deleted pod from provider")
 
-	if err := pc.forceDeletePodResource(ctx, namespace, name); err != nil {
-		span.SetStatus(err)
-		return err
-	}
-	log.G(ctx).Info("Deleted pod from Kubernetes")
-
-	return nil
-}
-
-func (pc *PodController) forceDeletePodResource(ctx context.Context, namespace, name string) error {
-	ctx, span := trace.StartSpan(ctx, "forceDeletePodResource")
-	defer span.End()
-	ctx = span.WithFields(ctx, log.Fields{
-		"namespace": namespace,
-		"name":      name,
-	})
-
-	var grace int64
-	if err := pc.client.Pods(namespace).Delete(name, &metav1.DeleteOptions{GracePeriodSeconds: &grace}); err != nil {
-		if errors.IsNotFound(err) {
-			log.G(ctx).Debug("Pod does not exist in Kubernetes, nothing to delete")
-			return nil
-		}
-		span.SetStatus(err)
-		return pkgerrors.Wrap(err, "Failed to delete Kubernetes pod")
-	}
 	return nil
 }
 
@@ -279,4 +236,48 @@ func (pc *PodController) podStatusHandler(ctx context.Context, key string) (retE
 	}
 
 	return pc.updatePodStatus(ctx, pod, key)
+}
+
+func (pc *PodController) deletePodHandler(ctx context.Context, key string) error {
+	ctx, span := trace.StartSpan(ctx, "processDeletionReconcilationWorkItem")
+	defer span.End()
+
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	ctx = span.WithFields(ctx, log.Fields{
+		"namespace": namespace,
+		"name":      name,
+	})
+
+	if err != nil {
+		// Log the error as a warning, but do not requeue the key as it is invalid.
+		log.G(ctx).Warn(pkgerrors.Wrapf(err, "invalid resource key: %q", key))
+		span.SetStatus(err)
+		return nil
+	}
+
+	// If the pod has been deleted from API server, we don't need to do anything.
+	k8sPod, err := pc.podsLister.Pods(namespace).Get(name)
+	if errors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		span.SetStatus(err)
+		return err
+	}
+
+	if running(&k8sPod.Status) {
+		log.G(ctx).Error("Force deleting pod in running state")
+	}
+
+	// We don't check with the provider before doing this delete. At this point, even if an outstanding pod status update
+	// was in progress,
+	err = pc.client.Pods(namespace).Delete(name, metav1.NewDeleteOptions(0))
+	if errors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		span.SetStatus(err)
+		return err
+	}
+	return nil
 }
--- a/node/pod.go
+++ b/node/pod.go
@@ -213,8 +213,12 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes 
 	}
 	kPod := obj.(*knownPod)
 	kPod.Lock()
-	podFromProvider := kPod.lastPodStatusReceivedFromProvider
+	podFromProvider := kPod.lastPodStatusReceivedFromProvider.DeepCopy()
 	kPod.Unlock()
+	// We need to do this because the other parts of the pod can be updated elsewhere. Since we're only updating
+	// the pod status, and we should be the sole writers of the pod status, we can blind overwrite it. Therefore
+	// we need to copy the pod and set ResourceVersion to 0.
+	podFromProvider.ResourceVersion = "0"
 
 	if _, err := pc.client.Pods(podFromKubernetes.Namespace).UpdateStatus(podFromProvider); err != nil {
 		span.SetStatus(err)
--- a/node/podcontroller.go
+++ b/node/podcontroller.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	pkgerrors "github.com/pkg/errors"
@@ -51,7 +52,7 @@ type PodLifecycleHandler interface {
 
 	// DeletePod takes a Kubernetes Pod and deletes it from the provider. Once a pod is deleted, the provider is
 	// expected to call the NotifyPods callback with a terminal pod status where all the containers are in a terminal
-	// state, as well as the pod.
+	// state, as well as the pod. DeletePod may be called multiple times for the same pod.
 	DeletePod(ctx context.Context, pod *corev1.Pod) error
 
 	// GetPod retrieves a pod by name from the provider (can be cached).
@@ -101,6 +102,9 @@ type PodController struct {
 
 	k8sQ workqueue.RateLimitingInterface
 
+	// deletionQ is a queue on which pods are reconciled, and we check if pods are in API server after grace period
+	deletionQ workqueue.RateLimitingInterface
+
 	// From the time of creation, to termination the knownPods map will contain the pods key
 	// (derived from Kubernetes' cache library) -> a *knownPod struct.
 	knownPods sync.Map
@@ -127,6 +131,7 @@ type knownPod struct {
 	// should be immutable to avoid having to hold the lock the entire time you're working with them
 	sync.Mutex
 	lastPodStatusReceivedFromProvider *corev1.Pod
+	lastPodUsed                       *corev1.Pod
 }
 
 // PodControllerConfig is used to configure a new PodController.
@@ -190,6 +195,7 @@ func NewPodController(cfg PodControllerConfig) (*PodController, error) {
 		done:            make(chan struct{}),
 		recorder:        cfg.EventRecorder,
 		k8sQ:            workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "syncPodsFromKubernetes"),
+		deletionQ:       workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "deletePodsFromKubernetes"),
 	}
 
 	return pc, nil
@@ -207,7 +213,7 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 
 	defer func() {
 		pc.k8sQ.ShutDown()
-
+		pc.deletionQ.ShutDown()
 		pc.mu.Lock()
 		pc.err = retErr
 		close(pc.done)
@@ -241,13 +247,8 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 		},
 		UpdateFunc: func(oldObj, newObj interface{}) {
 			// Create a copy of the old and new pod objects so we don't mutate the cache.
-			oldPod := oldObj.(*corev1.Pod)
 			newPod := newObj.(*corev1.Pod)
-			// Skip adding this pod's key to the work queue if its .metadata (except .metadata.resourceVersion) and .spec fields haven't changed.
-			// This guarantees that we don't attempt to sync the pod every time its .status field is updated.
-			if podsEffectivelyEqual(oldPod, newPod) {
-				return
-			}
+
 			// At this point we know that something in .metadata or .spec has changed, so we must proceed to sync the pod.
 			if key, err := cache.MetaNamespaceKeyFunc(newPod); err != nil {
 				log.G(ctx).Error(err)
@@ -261,6 +262,8 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 			} else {
 				pc.knownPods.Delete(key)
 				pc.k8sQ.AddRateLimited(key)
+				// If this pod was in the deletion queue, forget about it
+				pc.deletionQ.Forget(key)
 			}
 		},
 	})
@@ -292,6 +295,15 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 		}()
 	}
 
+	for id := 0; id < podSyncWorkers; id++ {
+		wg.Add(1)
+		workerID := strconv.Itoa(id)
+		go func() {
+			defer wg.Done()
+			pc.runDeletionReconcilationWorker(ctx, workerID, pc.deletionQ)
+		}()
+	}
+
 	close(pc.ready)
 
 	log.G(ctx).Info("started workers")
@@ -299,6 +311,7 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 	log.G(ctx).Info("shutting down workers")
 	pc.k8sQ.ShutDown()
 	podStatusQueue.ShutDown()
+	pc.deletionQ.ShutDown()
 
 	wg.Wait()
 	return nil
@@ -350,6 +363,7 @@ func (pc *PodController) syncHandler(ctx context.Context, key string) error {
 
 	// Add the current key as an attribute to the current span.
 	ctx = span.WithField(ctx, "key", key)
+	log.G(ctx).WithField("key", key).Debug("sync handled")
 
 	// Convert the namespace/name string into a distinct namespace and name.
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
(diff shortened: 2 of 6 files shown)


node/pod.go
```diff
@@ -156,40 +156,25 @@ func (pc *PodController) handleProviderError(ctx context.Context, span trace.Spa
...
-	log.G(ctx).Debug("Deleted pod from provider")
-
-	if err := pc.forceDeletePodResource(ctx, namespace, name); err != nil {
+	if err != nil {
 		span.SetStatus(err)
 		return err
 	}
-	log.G(ctx).Info("Deleted pod from Kubernetes")
+
+	log.G(ctx).Debug("Deleted pod from provider")
```
pires: Why are we changing the logging level?

node/pod.go
```diff
@@ -156,40 +156,25 @@ func (pc *PodController) handleProviderError(ctx context.Context, span trace.Spa
...
 
-	var delErr error
-	if delErr = pc.provider.DeletePod(ctx, pod.DeepCopy()); delErr != nil && !errdefs.IsNotFound(delErr) {
-		span.SetStatus(delErr)
-		return delErr
+	err := pc.provider.DeletePod(ctx, pod.DeepCopy())
+	if errdefs.IsNotFound(err) {
+		// TODO: Figure out what to do here. Previously, we used to delete the pod non-gracefully from Kubernetes
+		// The problem is that if a pod was deleted, and then it is no longer in the provider, we end up deleting it
+		// from kubernetes due to no reason whatsoever.
```
pires: This is confusing to me. _if the pod is deleted_ means the pod was marked for deletion in the API, right? After all, `deletePod` is called when an event for deletion is received. So, if it's marked for deletion and it's not found in the provider, why shouldn't we (finally) delete it from the API?

node/podcontroller.go
```diff
@@ -113,6 +113,11 @@ type PodController struct {
 	// From the time of creation, to termination the knownPods map will contain the pods key
 	// (derived from Kubernetes' cache library) -> a *knownPod struct.
 	knownPods sync.Map
+
+	// podForDeletion is also a map of key -> deletablePod. It is used only by the syncHandler to supress
+	// duplicate deletions. It is deleted when the pod is non-gracefully deleted from Kubernetes, in the event
+	// handler.
+	podForDeletion sync.Map
```
pires: Again, I'm confused. The deletion event comes from the API when the pod is marked for deletion.

node/pod.go
sargun: @pires IMHO, that's erroneous behaviour. Imagine this case:

1. The user sets deletion grace period to 60 seconds
2. The VK receives this, and calls delete on provider
3. The provider deletes the pod, and NotifyPod with a terminal status (but the update is delayed)
4. The user cranks deletion timestamp down to 30 seconds
5. The VK receives this update, and calls deletepod on the provider, and gets a pod not found
6. ???

If ??? is delete the pod from API server, we will never get the update in step 3.

node/pod.go
sargun: We aren't. Github seems confused. If you use the patience diff algorithm:

```
-func (pc *PodController) deletePod(ctx context.Context, namespace, name string) error {
+func (pc *PodController) deletePod(ctx context.Context, pod *corev1.Pod) error {
        ctx, span := trace.StartSpan(ctx, "deletePod")
        defer span.End()
-
-       pod, err := pc.provider.GetPod(ctx, namespace, name)
-       if err != nil {
-               if errdefs.IsNotFound(err) {
-                       // The provider is not aware of the pod, but we must still delete the Kubernetes API resource.
-                       return pc.forceDeletePodResource(ctx, namespace, name)
-               }
-               return err
-       }
-
-       // NOTE: Some providers return a non-nil error in their GetPod implementation when the pod is not found while some other don't.
-       if pod == nil {
-               // The provider is not aware of the pod, but we must still delete the Kubernetes API resource.
-               return pc.forceDeletePodResource(ctx, namespace, name)
-       }
-
        ctx = addPodAttributes(ctx, span, pod)
 
-       var delErr error
-       if delErr = pc.provider.DeletePod(ctx, pod.DeepCopy()); delErr != nil && !errdefs.IsNotFound(delErr) {
-               span.SetStatus(delErr)
-               return delErr
+       err := pc.provider.DeletePod(ctx, pod.DeepCopy())
+       if errdefs.IsNotFound(err) {
+               // TODO: Figure out what to do here. Previously, we used to delete the pod non-gracefully from Kubernetes
+               // The problem is that if a pod was deleted, and then it is no longer in the provider, we end up deleting it
+               // from kubernetes due to no reason whatsoever.
+               return err
+       }
+
+       if err != nil {
+               span.SetStatus(err)
+               return err
        }
 
        log.G(ctx).Debug("Deleted pod from provider")
 
-       if err := pc.forceDeletePodResource(ctx, namespace, name); err != nil {
-               span.SetStatus(err)
-               return err
-       }
-       log.G(ctx).Info("Deleted pod from Kubernetes")
-
        return nil
 }
```

node/pod.go
sargun: I kind of wish that Kubernetes had vector clocks. So, you could say, "Does this object have this update?" If it's strictly newer, then do the delete.

node/podcontroller.go
sargun: Perhaps this is a matter of phrasing. I've attempted to rephrase.

node/pod.go
cpuguy83: Another wrench here, whether or not `DeletePod` should even return a `NotFound` is not straight forward.

pires: I think we may have different understandings of grace-period here. The VK, at most, should wait no more than grace-period duration for the provider to announce the pod was in fact deleted, hence the responsibility of deleting it, even without knowing the grace-period, is provider's only.

Maybe [this](https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-terminating-with-grace) helps us being on the same page?

sargun: If the contention is this comment, then I can remove it.

--

`terminationGracePeriodSeconds`
> Optional duration in seconds the pod needs to terminate gracefully. May be decreased in delete request. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period will be used instead. The grace period is the duration in seconds after the processes running in the pod are sent a termination signal and the time when the processes are forcibly halted with a kill signal. Set this value longer than the expected cleanup time for your process. Defaults to 30 seconds.

If the provider gets a DeletePod, nothing in our specification states that the pod will actually be deleted by the time the call returns. 

As stated in the API specification:
> May be decreased in delete request.

If the terminationgraceperiodseconds is cranked down, and the provider has already deleted the pod, this can result in us erroneously deleting the pod in K8s prior to the expiration of the grace period.

pires: OK maybe we're missing each other. `kubectl delete --grace-period` overrides `PodSpec.terminationGracePeriodSeconds`. The VK should enforce the former while the provider should enforce the latter, or so is my understanding.

sargun: Right. But if termination grace period changes, we need to inform the provider (which we do), but it can result in an erroneous action being taken. One of the things we can do is say if it is not found AND it was because this was a duplicate delete, ignore not found.

pires: I don't think we can be sure about that. Say the VK crashes/restarts in between events. It will lose its in-memory map (that you're adding in this PR) and won't understand it's a duplicate delete, right?

sargun: @pires You're right. I believe that behaviour is erroneous. I believe we need a mechanism around that in the future.

cpuguy83:
Ping @ibabou

sargun:
what part of the behavior are you questioning?

sargun:
So, there are two primary behaviour changes:
1. We no longer force-delete pods at deletion of the pod, and instead rely on being notified with a terminal pod status.
2. We suppress duplicate deletes.

(1) is to address three problems. First, it addresses providers who act asynchronously upon callbacks. Second, it allows users to provide terminal pod status updates. Third, it allows for multi-phasic shutdown.

Our use case is that we do a bunch of stuff at shutdown, for example:
1. Log uploading
2. Network shutdown
3. Container cleanup

This means that (1) steps can fail, and (2) different resources free up at different times. 

No matter what, eventually the grace period kicks in, and the pod is deleted from API server. 

---
In regards to suppressing duplicate deletes -- it doesn't violate the existing contract. The only potential situation we may want to enable duplicate deletes is if an updateable field (like labels) gets changed. 

It avoids the behaviour where you have:
1. Something deletes pod
2. We call `DeletePod` on the provider
3. The provider NotifyPods() us of the terminal pod status
4. We call DeletePod again because we updated the pod in the server, and then we get a pod not found back.

node/podcontroller.go
```diff
@@ -342,11 +372,32 @@ func (pc *PodController) syncPodInProvider(ctx context.Context, pod *corev1.Pod)
...
+			// We've tried to delete this pod before.
+			dPod := obj.(deleteablePod)
+			// If our termination grace period is equal to, or longer than the previous pod's termination grace period, we do not
+			// need to retry the delete.
+			if *dPod.pod.Spec.TerminationGracePeriodSeconds <= *pod.Spec.TerminationGracePeriodSeconds {
+				return nil
+			}
+		}
+
+		if err := pc.deletePod(ctx, pod); errdefs.IsNotFound(err) {
```
ibabou: Shouldn't we've a check here to do a Get always and do force removal if not found ?

node/pod.go
```diff
@@ -156,40 +156,25 @@ func (pc *PodController) handleProviderError(ctx context.Context, span trace.Spa
...
-		return pc.forceDeletePodResource(ctx, namespace, name)
-	}
-
 	ctx = addPodAttributes(ctx, span, pod)
 
-	var delErr error
-	if delErr = pc.provider.DeletePod(ctx, pod.DeepCopy()); delErr != nil && !errdefs.IsNotFound(delErr) {
-		span.SetStatus(delErr)
-		return delErr
+	err := pc.provider.DeletePod(ctx, pod.DeepCopy())
```
ibabou: I might not be able to fully capture the change here. but I think the code now don't forcibly delete if Get fails (pod doesn't exist on the provider side anymore) as before. I agree to keep the graceful logic and be able to receive updates. but why not still keep the case of (if Get fails as not found -> forcibly clear from API). however we might still call DeletePod to make sure we did our best, even if GetPod failed.

cpuguy83:
Just to document my understanding of what the Kubelet does:

Here is what I understand about how Kubelet deals with pod deletion:

1. Users performs `kubectl delete`
2. Deletion timestamp (and grace period) are set on the pod
3. Kubelet gets the update notification with the deletion timestamp
4. Kubelet terminates the pod, sends kill signal after grace period if it did not terminate
5. Kubelet updates the API server with the terminal state
6. Kubelet is notified of the terminal state update and deletes the pod from the API server.

@chewong Would love to hear if you have any input on how this should be handled.

pires:
> 4\. Kubelet terminates the pod, sends kill signal after grace period if it did not terminate

IIRC there's two steps here:
- kubelet sends signal to pod and waits grace period
- if pod is not dead after grace period, kubelet force terminates it

sargun:
Should we wait for the pod to disappear from the provider + wait for the containers to all reach a terminal state, or should we _just_ wait for the containers to get to a terminal state prior to deleting the pod?

sargun:
@pires I don't think that really matters, because in our situation, (4) is completely delegated to the provider  / PodLifecycleHandler.

sargun:
@cpuguy83 @pires I added the last behaviour.

pires:
Please, give me the weekend to review.

node/pod.go
```diff
@@ -156,40 +156,25 @@ func (pc *PodController) handleProviderError(ctx context.Context, span trace.Spa
...
+	if errdefs.IsNotFound(err) {
```
cpuguy83: Let's remove this since it's handled by the caller anyway (And actually it seems like it's deleting it).

node/podcontroller.go
```diff
@@ -372,21 +388,39 @@ func (pc *PodController) syncHandler(ctx context.Context, key string) error {
...
+			}
+			if err != nil {
+				err = pkgerrors.Wrapf(err, "failed to delete pod %q in the provider", loggablePodNameFromCoordinates(namespace, name))
+				span.SetStatus(err)
+			}
 			return err
 		}
+		// We tried to delete the pod before. That means we can bail out here, and since this is the last time we are
+		// going to see the pod being updates from Kubernetes, we can remove the pod from the pod deletion map
+		pc.podForDeletion.Delete(key)
```
cpuguy83: What happens to this in the case we don't get a pod status update to trigger this deletion (common example would be a non-async provider)?

sargun: You mean, that if a goes into succeeded, or failed without updating any of its container statuses to terminal statuses? If it's in `Running`, then eventually, the fetchPodStatuses will fetch against it and update the pod status.

sargun: If it goes into suceeded, or failed, we can keep polling the provider. How do providers handle this now?

cpuguy83: Now deletion is handled synchronously.
What I'm worried about is this not getting cleaned up, and likewise for the API server.

The pod status loop (the one that periodically hits sync providers) clean this out?

cpuguy83: We could also remove support for sync providers in core and make sure providers are aware they need to trigger a pod status update after delete has occurred.

node/pod.go
sargun: Done.

node/podcontroller.go
pires: I am in favor for the latter and cut 2.0. We can keep 1.x for bug-fixes.

sargun:
See https://github.com/virtual-kubelet/virtual-kubelet/pull/770. We can merge that PR, and then the corner cases here become a lot less interesting. Whether we call it `v1.2.0`, or `v2.0.0`, I'm not opinionated, but I *strongly* agree that we should deprecate the sync interface. Both from a performance / scalability perspective, and a sanity (fewer approaches). AFAIK, there's no way to implement the full async capability set atop the sync interface without making some (IMHO) unreasonable compromises.

node/podcontroller.go
cpuguy83: I still think we a GC loop to handle edge cases like missed terminal status notifications or other potential failures.
We can always throw big warning messages in the logs to signal that there may be a bug in the provider for anything we do end up having to clean up.

sargun:
I’ll rebase this. Do we need to fix anything else? Or change the behavior, or is this good for an MVP?

The one thing I was thinking was to put this behind a feature flag. Opinions?

pires:
I'm good with feature gates as we can make this opt-in for now and get validation but I concern doing so may add to confusion for provider owners.

Alternatively, @cpuguy83 @sargun can we take like one hour or two and review this thoroughly over a call and decide what to do?

cpuguy83:
We could add a feature gate, but if we automatically clean up deleted pods after the grace period I think there is no need to gate the feature unless there are some other concerns there.

sargun:
Let me rework this PR. I’m going to split it into two different parts. One part, which will suppress duplicate deletes, and a second which will call delete pod on the API server only once we see all the containers have been deleted, or graceperiod has been reached.

sargun:
I know I've suggested this before, but I figure I'll suggest it again. I think we should move to a state machine model. I think that @cpuguy83 brings up a good point around the fact you need synchronizations between the updates originated from Kubernetes, and the updates from API Server. 

A pure line-level machine is not practical, as far as I can tell. For example, if the following happens when the pod is in a running state:
1. Receive a pod status update from Kubernetes with DeletionTimestamp set
2. Call delete on Provider
3. Receive another pod status update from Kubernetes with DeletionTimestamp set
3. ???

If you use a pure line-level machine you would re-trigger the deletion. The line-level machine inherently needs "memory". This memory doesn't have to be unbounded, but if we can enumerate the finite set of states the system can get into, we can hard code every transition in the system, as well as things like timeouts in waiting for transition. 

I think I should remove the duplicate-deletion suppression code from here, and put the ability to avoid deletion of the Pod behind a flag in pod controller config. We can then incrementally get to a generalized solution.

sargun:
There is now a queue that handles deletion of items from the apiserver after graceperiod is expired.

node/pod.go
cpuguy83: In this case should we add the pod to the delete queue?

node/pod.go
```diff
@@ -276,3 +261,37 @@ func (pc *PodController) podStatusHandler(ctx context.Context, key string) (retE
 
 	return pc.updatePodStatus(ctx, pod, key)
 }
+
+func (pc *PodController) deletionReconcilationWorkItemHandler(ctx context.Context, key string) error {
+	ctx, span := trace.StartSpan(ctx, "processDeletionReconcilationWorkItem")
+	defer span.End()
+	// Add the current key as an attribute to the current span.
```
cpuguy83: Comment is no neccessary.

node/pod.go
```diff
@@ -276,3 +261,37 @@ func (pc *PodController) podStatusHandler(ctx context.Context, key string) (retE
...
+	ctx = span.WithField(ctx, "key", key)
+
+	// Convert the namespace/name string into a distinct namespace and name.
```
cpuguy83: Comment is not necessary.

node/pod.go
```diff
@@ -276,3 +261,37 @@ func (pc *PodController) podStatusHandler(ctx context.Context, key string) (retE
...
+	if err != nil && !errdefs.IsNotFound(err) {
+		return err
+	}
+	if err == nil && pod != nil {
+		log.G(ctx).Warn("Container still exists in provider, requeueing")
+		pc.deletionQ.AddRateLimited(key)
+		return nil
+	}
+
+	return pc.client.Pods(namespace).Delete(name, metav1.NewDeleteOptions(0))
```
cpuguy83: Error checking logic above does not look right.
If there is any error then we'll delete the pod but I think we should only be doing this if the error is "not found".

node/pod.go
```diff
@@ -276,3 +261,37 @@ func (pc *PodController) podStatusHandler(ctx context.Context, key string) (retE
...
+
+	// We check if the pod still exists in the provider. If it no longer there, then we ensure that it's
+	// not in the API server.
+	pod, err := pc.provider.GetPod(ctx, namespace, name)
+	if err != nil && !errdefs.IsNotFound(err) {
+		return err
+	}
+	if err == nil && pod != nil {
+		log.G(ctx).Warn("Container still exists in provider, requeueing")
+		pc.deletionQ.AddRateLimited(key)
```
cpuguy83: Can you explain this logic?

node/pod.go
```diff
@@ -276,3 +261,37 @@ func (pc *PodController) podStatusHandler(ctx context.Context, key string) (retE
 
 	return pc.updatePodStatus(ctx, pod, key)
 }
+
+func (pc *PodController) deletionReconcilationWorkItemHandler(ctx context.Context, key string) error {
```
cpuguy83: `deletePodHandler`?

node/pod.go
```diff
@@ -228,6 +210,9 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
 		"old reason": podFromKubernetes.Status.Reason,
 	}).Debug("Updated pod status in kubernetes")
 
+	// If we update the pod status with a terminal pod status, then we need to go ahead and forcefully delete the pod
+	// from api server
```
cpuguy83: Are we supposed to do something here? I don't understand the placement of this comment.

node/pod.go
```diff
@@ -213,8 +191,12 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
 	}
 	kPod := obj.(*knownPod)
 	kPod.Lock()
-	podFromProvider := kPod.lastPodStatusReceivedFromProvider
+	podFromProvider := kPod.lastPodStatusReceivedFromProvider.DeepCopy()
 	kPod.Unlock()
+	// We need to do this because the other parts of the pod can be updated elsewhere. Since we're only updating
+	// the pod status, and we should be the sole writers of the pod status, we can blind overwrite it. Therefore
+	// we need to copy the pod and set ResourceVersion to 0.
+	podFromProvider.ResourceVersion = "0"
```
cpuguy83: I don't understand this and don't see an example of this in the kubelet code.
Does the same issue exist in Kubelet?

node/podcontroller.go
```diff
@@ -503,3 +578,30 @@ func loggablePodName(pod *corev1.Pod) string {
...
+		}
+		return false
+	}
+
+	return cmp.Equal(p1, p2, cmp.FilterPath(filterForResourceVersion, cmp.Ignore()))
+}
+
+// notRunning returns true if every status is terminated or waiting, or the status list
+// is empty.
+func notRunning(podStatus *corev1.PodStatus) bool {
```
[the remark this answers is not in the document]
sargun: I copied and pasted this code from kubernetes.

node/podcontroller.go
```diff
@@ -372,35 +383,64 @@ func (pc *PodController) syncHandler(ctx context.Context, key string) error {
...
 	defer span.End()
 
 	// Add the pod's attributes to the current span.
 	ctx = addPodAttributes(ctx, span, pod)
 
+	// If the pod('s containers) is no longer in a running state then we force-delete the pod from API server
+	// more context is here: https://github.com/virtual-kubelet/virtual-kubelet/pull/760
+	if pod.DeletionTimestamp != nil && notRunning(&pod.Status) {
+		log.G(ctx).Debug("Force deleting pod from API Server as it is no longer running")
+		return pc.forceDeletePodResource(ctx, pod.Namespace, pod.Name)
```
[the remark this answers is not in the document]
sargun: Good point.

node/pod.go
sargun: So, the way the kubelet does it is by fetching the current pod (status) from api server, and then does a client-side merge, before doing a server side patch. Code is attached:

```
// syncPod syncs the given status with the API server. The caller must not hold the lock.
func (m *manager) syncPod(uid types.UID, status versionedPodStatus) {
        if !m.needsUpdate(uid, status) {
                klog.V(1).Infof("Status for pod %q is up-to-date; skipping", uid)
                return
        }

        // TODO: make me easier to express from client code
        pod, err := m.kubeClient.CoreV1().Pods(status.podNamespace).Get(status.podName, metav1.GetOptions{})
        if errors.IsNotFound(err) {
                klog.V(3).Infof("Pod %q (%s) does not exist on the server", status.podName, uid)
                // If the Pod is deleted the status will be cleared in
                // RemoveOrphanedStatuses, so we just ignore the update here.
                return
        }
        if err != nil {
                klog.Warningf("Failed to get status for pod %q: %v", format.PodDesc(status.podName, status.podNamespace, uid), err)
                return
        }

        translatedUID := m.podManager.TranslatePodUID(pod.UID)
        // Type convert original uid just for the purpose of comparison.
        if len(translatedUID) > 0 && translatedUID != kubetypes.ResolvedPodUID(uid) {
                klog.V(2).Infof("Pod %q was deleted and then recreated, skipping status update; old UID %q, new UID %q", format.Pod(pod), uid, translatedUID)
                m.deletePodStatus(uid)
                return
        }

        oldStatus := pod.Status.DeepCopy()
        newPod, patchBytes, err := statusutil.PatchPodStatus(m.kubeClient, pod.Namespace, pod.Name, *oldStatus, mergePodStatus(*oldStatus, status.status))
        klog.V(3).Infof("Patch status for pod %q with %q", format.Pod(pod), patchBytes)
        if err != nil {
                klog.Warningf("Failed to update status for pod %q: %v", format.Pod(pod), err)
                return
        }
        pod = newPod

        klog.V(3).Infof("Status for pod %q updated successfully: (%d, %+v)", format.Pod(pod), status.version, status.status)
        m.apiStatusVersions[kubetypes.MirrorPodUID(pod.UID)] = status.version

        // We don't handle graceful deletion of mirror pods.
        if m.canBeDeleted(pod, status.status) {
                deleteOptions := metav1.NewDeleteOptions(0)
                // Use the pod UID as the precondition for deletion to prevent deleting a newly created pod with the same name and namespace.
                deleteOptions.Preconditions = metav1.NewUIDPreconditions(string(pod.UID))
                err = m.kubeClient.CoreV1().Pods(pod.Namespace).Delete(pod.Name, deleteOptions)
                if err != nil {
                        klog.Warningf("Failed to delete status for pod %q: %v", format.Pod(pod), err)
                        return
                }
                klog.V(3).Infof("Pod %q fully terminated and removed from etcd", format.Pod(pod))
                m.deletePodStatus(uid)
        }
}

```

node/podcontroller.go
```diff
@@ -372,35 +383,64 @@ func (pc *PodController) syncHandler(ctx context.Context, key string) error {
 			span.SetStatus(err)
 			return err
 		}
-		// At this point we know the Pod resource doesn't exist, which most probably means it was deleted.
-		// Hence, we must delete it from the provider if it still exists there.
-		if err := pc.deletePod(ctx, namespace, name); err != nil {
-			err := pkgerrors.Wrapf(err, "failed to delete pod %q in the provider", loggablePodNameFromCoordinates(namespace, name))
+
+		pod, err = pc.provider.GetPod(ctx, namespace, name)
+		if errdefs.IsNotFound(err) || (err == nil && pod == nil) {
```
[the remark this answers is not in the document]
sargun: Queued up for deletion where? In the API server? The only way we get to this code path is if the pod lister did not contain the pod, and therefore we do not need to queue it for deletion in api server.

gabrielhartmann:
The Titus virtual-kubelet implementation works with this PR.

node/pod.go
sargun: If the container is in the provider, we might still get an outstanding update.

node/podcontroller.go
```diff
@@ -372,35 +383,80 @@ func (pc *PodController) syncHandler(ctx context.Context, key string) error {
...
 	defer span.End()
 
 	// Add the pod's attributes to the current span.
 	ctx = addPodAttributes(ctx, span, pod)
 
+	// If the pod('s containers) is no longer in a running state then we force-delete the pod from API server
+	// more context is here: https://github.com/virtual-kubelet/virtual-kubelet/pull/760
+	if pod.DeletionTimestamp != nil && notRunning(&pod.Status) {
+		log.G(ctx).Debug("Force deleting pod from API Server as it is no longer running")
+		pc.deletionQ.Add(key)
```
cpuguy83: Should this return here?

node/pod.go
```diff
@@ -276,3 +238,51 @@ func (pc *PodController) podStatusHandler(ctx context.Context, key string) (retE
...
+	}
+	if err != nil {
+		span.SetStatus(err)
+		return err
+	}
+
+	if running(&k8sPod.Status) {
+		log.G(ctx).Error("Force deleting running pod, provider probably didn't originate container status update")
+		// Although a pod with deletiontimestamp == nil should never end up here, it's worth making the check
+		if k8sPod.DeletionTimestamp != nil && time.Now().After(k8sPod.DeletionTimestamp.Time) {
```
cpuguy83: Should we be returning if we are not after the grace period?

node/pod.go
```diff
@@ -276,3 +238,51 @@ func (pc *PodController) podStatusHandler(ctx context.Context, key string) (retE
...
+			log.G(ctx).Warn("Force deleting pod after grace period")
```
cpuguy83: Info?

node/pod.go
sargun: I don't think so. We can delete *before* the grace period time. There's no guarantee that the pod will stick around *at least* as long as the grace period.

node/pod.go
sargun: I think this is an error / warn, as it shouldn't happen. It only comes up if providers are misbehaving.

node/pod.go
cpuguy83: ping @pires

sargun: The “0” thing is a “regression”. In 1.0, we used to do it here:
https://github.com/virtual-kubelet/virtual-kubelet/blob/release-1.0/node/pod.go#L114

PR #761, opened 2019-09-05
Cache Downloaded Go Modules

sargun:
This caches the downloaded go modules. It invalidates them based a hash of the go.mod, and go.sum. The test step showed a reduction from 1:30 -> 1:00, and the e2e tests from 8:30 to 5 minutes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,9 @@ jobs:
     working_directory: /go/src/github.com/virtual-kubelet/virtual-kubelet
     steps:
       - checkout
+      - restore_cache:
+          keys:
+            - validate-{{ checksum "go.mod" }}-{{ checksum "go.sum" }}
       - run:
           name: go vet
           command: V=1 CI=1 make vet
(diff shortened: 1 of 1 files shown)


cpuguy83:
What's the reason to invalidate at all?

sargun:
@cpuguy83 Because otherwise if you update go.mod, it'll never pickup the new modules and add them to cache.

cpuguy83:
I don't understand, isn't that go's responsibility?

sargun:
From a Circle-CI caching perspective, it wont make a tarball of the new modules. What happens is that at the end of the build, it hashes go.sum and go.mod, and stores the dependencies based on that. At startup, it downloads dependencies based on go.sum, and go.mod.

If we didn't invalidate them, then if you updated go.mod, it would not update the tarball, since it's first-write-wins.

pires:
Updating `go.mod` should never happen in the course of a a CI run, right?

sargun:
@pires If I update my deps, go.mod will be updated, resulting in different modules being downloaded, and hence requires cache invalidation.

sargun:
@pires i.e. between two CI runs.

sargun:
The reason you need invalidation is because it's first-write-wins, not last-write-wins. If you update your deps, Circle-CI wont know and will continue to give you a stale cache.

pires:
I was assuming we have one new machine per CI run. Is this not the case then? 🤔

sargun:
@pires We do get a blank machine per CI run. What caching does is that it takes everything downloaded from the previous CI run and copies it over. This copies it over ONLY if the hash of go.sum, and go.mod match.

PR #770, opened 2019-09-24
Remove sync providers

sargun:
This removes the legacy sync provider interface. All new providers
are expected to implement the async NotifyPods interface.

The legacy sync provider interface creates complexities around
how the deletion flow works, and the mixed sync and async APIs
block us from evolving functionality.

This collapses in the NotifyPods interface into the PodLifecycleHandler
interface.


sargun:
This PR changes the interface for PodLifeCycleHandler (Provider), to be async. This means a (breaking) interface change, and a change in behaviour. The primary motivator for this work is graceful deletes (https://github.com/virtual-kubelet/virtual-kubelet/pull/760). On the other side, supporting async invocations only also mean that the API footprint we need to design for is cutdown.

I want to get a read on the provider maintainers (https://github.com/virtual-kubelet/community/blob/master/OWNERS.md), and farm for dissent. I think we should try to move to have this change in sooner than later (As the earlier we change the interface, the fewer providers in the future we'll have to deal with changing).

### Azure
@chewong 
### Fargate
@ofiliz 
### Alibaba Cloud
@resouer 
### Huawei
@fisherxu 
### Azure Batch
@lawrencegripper 
### Hashicorp Nomad
@anubhavmishra

pires:
I think it's fine for providers to not change for as long as they can use a tagged release, ie `1.1.0`.

sargun:
This has been open for ~6 days now. I suggest we wait until tomorrow, when it's been open for one week. @cpuguy83 @pires, how long do you think we need to farm for dissent?

lawrencegripper:
Hi,

Sorry, had a discussion a while back on email around maintaining the AzureBatch provider the TLDR is I’m unable to dedicate any time to it and suggested marking it as depreciated.

chewong:
cc @ibabou

sargun:
@cpuguy83 @pires I don't see any dissent from the known providers, and according to Github search, the v1 API is not consumed. I propose we merge this, and subsequently release a new version of VK.

cpuguy83:
Let's merge this, if there is a late-comer with objections please let us know.
We can always revert or take other steps to resolve later.

Thanks!

PR #780, opened 2019-10-11
Fix pod status updates for when pod is updated outside of VK

sargun:
Pods can be updated outside of VK. Right now, if this happens, pod
status updates are dropped because the resourceversion from the
provider will mismatch with what's on the server, breaking
pod status updates.

Since we're the only ones writing to the pod status, we
can do a blind overwrite.


sargun:
I have pulled this into PR https://github.com/virtual-kubelet/virtual-kubelet/pull/760.

PR #785, opened 2019-10-22
Add flags to tune GC behaviour

sargun:
This adds the `DelayGC`, and `DisableGC` flags in order to change the behaviour of the VK to customize deletion behaviour in your own implementation of K8s controllers.

### DIFF
--- a/node/podcontroller.go
+++ b/node/podcontroller.go
@@ -124,6 +124,8 @@ type PodController struct {
 	// This is used since `pc.Run()` is typically called in a goroutine and managing
 	// this can be non-trivial for callers.
 	err error
+
+	podDeletionPolicy PodDeletionPolicy
 }
 
 type knownPod struct {
@@ -155,6 +157,8 @@ type PodControllerConfig struct {
 	ConfigMapInformer corev1informers.ConfigMapInformer
 	SecretInformer    corev1informers.SecretInformer
 	ServiceInformer   corev1informers.ServiceInformer
+
+	PodDeletionPolicy PodDeletionPolicy
 }
 
 func NewPodController(cfg PodControllerConfig) (*PodController, error) {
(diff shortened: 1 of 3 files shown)


gabrielhartmann:
@sargun These configuration flags are useful but not easily set by providers.  The[ root command generates the podcontroller config](https://github.com/sargun/virtual-kubelet/blob/disable-gc/cmd/virtual-kubelet/internal/commands/root/root.go#L180-L188) without consulting outside sources such as a provider.

gabrielhartmann:
Currently the only way to get config into the root command is [via the `opt.Opts` argument](https://github.com/sargun/virtual-kubelet/blob/disable-gc/cmd/virtual-kubelet/internal/commands/root/root.go#L45).

This options argument is not consulted when [constructing the PodControllerConfig object](https://github.com/sargun/virtual-kubelet/blob/disable-gc/cmd/virtual-kubelet/internal/commands/root/root.go#L180-L188).

If the options you propose are to be consumed in a way that's consistent with the current configuration mechanism you'd want to mirror the options in the `opt.Opts` argument, and then consume them when constructing the `PodControllerconfig` object.

cpuguy83:
I'm not a huge fan of bool configurations, particularly when they conflict.
What if we pass in a policy value, e.g.

```go
type PodDeletionPolicy int

const (
	PodDeletionPolicyDefault PodDeletionPolicy = 0
	PodDeletionPolicyNever
	PodDeletionPolicyDelayingFn
)
```

For `PodDeletionPolicyDelayingFn` we'd have in the `PodControllerConfig` an option to expose how to handle that:

```go
type PodControllerConfig {
	...
	PodDeletionPolicy PodDeletionPolicy
	PodDeletionPolicyDelayingFn func(*corev1.Pod) (time.Duration)
}
```

sargun:
What about something like:

```
type PodDeletionPolicy interface {
  deletePod() bool
  delayDeletePod(pod *v1.Pod) time.Duration
}
```

```
func NewPodNeverDeletionPolicy() PodDeletionPolicy {
...
}

func CustomPodDeletionPolicy(delayingFunc func(*v1.Pod) time.Duration) PodDeletionPolicy {
...
}

func DelayUntilGracePeriodPodDeletionPolicy() PodDeletionPolicy {
}
```

I know you're not a fan of hidden interfaces / structs, but in this case, it looks good if we want "one" field to configure it.

cpuguy83:
The downsides of that is every polciy must implement all functions.
It also makes it harder to extend and maintain in the long run, and difficult to inject custom functionality.

node/podcontroller.go
```diff
@@ -155,6 +157,8 @@ type PodControllerConfig struct {
 	ConfigMapInformer corev1informers.ConfigMapInformer
 	SecretInformer    corev1informers.SecretInformer
 	ServiceInformer   corev1informers.ServiceInformer
+
+	PodDeletionPolicy PodDeletionPolicy
```
cpuguy83: I'm not sure how this is supposed to be used, because for each policy we end up having to check values anyway and we aren't using the interface. This seems to be more trying to make an enum in go using an interface.
If we need an enum, which I think we do, I think it's best use the standard way of doing this in go of using simple values.

sargun: If we have an enum, then we have two values. The enum + the delaying func. This prevents someone from specifying a delaying func, but it not working because they didn't set the enum correctly.

cpuguy83: I think that's fine, also not everything requires. The enum tells us the strategy so we know how to proceed, and in some cases of the enum that means we need to look at another field. Even then I don't think we always need to do that.
For common cases we can have an enum value for that. The extra field would be for injecting custom behavior.

sargun: One last thing. RE: Enum, what if you want to control the delete policy independently from the delay policy. For example, if you want to delay *some* pods, and delete *other* pods. Does it make sense to take the above interface and expose it, rather than having an enum value, and two knobs for controlling delete and delay?

Or is this just too wacky of a use case.

sargun: If we don't want to make it so that every user needs to implement both functions, then we can add some helper functions that make it so they need to implement one or the other. Since the deletion policy includes both of these, making them fully customizable seems like a "better" option. Otherwise, I think we need to expose these directly.

cpuguy83: My issue here is it's fundamentally not an interface.

sargun: Why not always have it be a function, and we can define our own default function which noops.

cpuguy83: How do you handle a case like "never delete"?
I was thinking we'd end up with a long list of return values to handle different cases.

sargun: If the function's signature is:
```go
deletionPolicyCallback(*v1.Pod) (shouldDelete bool, delayDelete time.Duration)
```

If you want to never delete, you just return `false` for the first value. This would get checked before processing it in the deletionQ. If you return `true`, and a non-zero deletion value, that delays deletion. If you return `true`, and a zero, or negative duration, that triggers immediate deletion.

cpuguy83: My worry is this can't possibly handle all cases just the common ones and I can see needing to scrap this in the future.

We should also avoid requiring changes in the core in order to support an edge case.

PR #786, opened 2019-10-23
Re-add support for sync providers

cpuguy83:
This brings back support for sync providers by wrapping them in a
provider that handles async notifications.

This way we don't break the existing API.

### DIFF
--- a/node/lifecycle_test.go
+++ b/node/lifecycle_test.go
@@ -107,7 +107,7 @@ func TestPodLifecycle(t *testing.T) {
 	}
 
 	// isPodDeletedGracefullyFunc is a condition func that waits until the pod is in a terminal state, which is the VK's
-	// action when the pod is deleted from the provider
+	// action when the pod is deleted from the provider.
 	isPodDeletedGracefullyFunc := func(ctx context.Context, watcher watch.Interface) error {
 		_, watchErr := watchutils.UntilWithoutRetry(ctx, watcher, func(ev watch.Event) (bool, error) {
 			log.G(ctx).WithField("event", ev).Info("got event")
@@ -130,7 +130,7 @@ func TestPodLifecycle(t *testing.T) {
 		t.Run(env, func(t *testing.T) {
 
 			// createStartDeleteScenario tests the basic flow of creating a pod, waiting for it to start, and deleting
-			// it gracefully
+			// it gracefully.
 			t.Run("createStartDeleteScenario", func(t *testing.T) {
 				assert.NilError(t, wireUpSystem(ctx, h(), func(ctx context.Context, s *system) {
 					testCreateStartDeleteScenario(ctx, t, s, isPodDeletedGracefullyFunc, true)
@@ -138,7 +138,7 @@ func TestPodLifecycle(t *testing.T) {
 			})
 
 			// createStartDeleteScenarioWithDeletionErrorNotFound tests the flow if the pod was not found in the provider
-			// for some reason
+			// for some reason.
 			t.Run("createStartDeleteScenarioWithDeletionErrorNotFound", func(t *testing.T) {
 				mp := h()
 				mp.setErrorOnDelete(errdefs.NotFound("not found"))
@@ -148,7 +148,7 @@ func TestPodLifecycle(t *testing.T) {
 			})
 
 			// createStartDeleteScenarioWithDeletionRandomError tests the flow if the pod was unable to be deleted in the
-			// provider
+			// provider.
 			t.Run("createStartDeleteScenarioWithDeletionRandomError", func(t *testing.T) {
 				mp := h()
 				deletionFunc := func(ctx context.Context, watcher watch.Interface) error {
@@ -186,7 +186,7 @@ func TestPodLifecycle(t *testing.T) {
 				})
 			})
 
-			// failedPodScenario ensures that the VK ignores failed pods that were failed prior to the PC starting up
+			// failedPodScenario ensures that the VK ignores failed pods that were failed prior to the pod controller starting up.
 			t.Run("failedPodScenario", func(t *testing.T) {
 				t.Run("mockProvider", func(t *testing.T) {
 					mp := newMockProvider()
@@ -196,7 +196,7 @@ func TestPodLifecycle(t *testing.T) {
 				})
 			})
 
-			// succeededPodScenario ensures that the VK ignores succeeded pods that were succeeded prior to the PC starting up.
+			// succeededPodScenario ensures that the VK ignores succeeded pods that were succeeded prior to the pod controller starting up.
 			t.Run("succeededPodScenario", func(t *testing.T) {
 				t.Run("mockProvider", func(t *testing.T) {
 					mp := newMockProvider()
@@ -207,7 +207,7 @@ func TestPodLifecycle(t *testing.T) {
 			})
 
 			// updatePodWhileRunningScenario updates a pod while the VK is running to ensure the update is propagated
-			// to the provider
+			// to the provider.
 			t.Run("updatePodWhileRunningScenario", func(t *testing.T) {
 				t.Run("mockProvider", func(t *testing.T) {
 					mp := newMockProvider()
--- a/node/lifecycle_test.go
+++ b/node/lifecycle_test.go
@@ -117,11 +117,14 @@ func TestPodLifecycle(t *testing.T) {
 		return watchErr
 	}
 
-	envs := map[string]func() testingProvider{
-		"Async": func() testingProvider {
+	envs := map[string]func(*testing.T) testingProvider{
+		"Async": func(t *testing.T) testingProvider {
 			return newMockProvider()
 		},
-		"Sync": func() testingProvider {
+		"Sync": func(t *testing.T) testingProvider {
+			if testing.Short() {
+				t.Skip()
+			}
 			return newSyncMockProvider()
 		},
 	}
@@ -132,7 +135,7 @@ func TestPodLifecycle(t *testing.T) {
 			// createStartDeleteScenario tests the basic flow of creating a pod, waiting for it to start, and deleting
 			// it gracefully.
 			t.Run("createStartDeleteScenario", func(t *testing.T) {
-				assert.NilError(t, wireUpSystem(ctx, h(), func(ctx context.Context, s *system) {
+				assert.NilError(t, wireUpSystem(ctx, h(t), func(ctx context.Context, s *system) {
 					testCreateStartDeleteScenario(ctx, t, s, isPodDeletedGracefullyFunc, true)
 				}))
 			})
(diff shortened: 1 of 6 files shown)


node/lifecycle_test.go
```diff
@@ -106,106 +106,118 @@ func TestPodLifecycle(t *testing.T) {
...
+				t.Run("mockProvider", func(t *testing.T) {
+					mp := newMockProvider()
+					assert.NilError(t, wireUpSystem(ctx, mp, func(ctx context.Context, s *system) {
+						testSucceededPodScenario(ctx, t, s)
+					}))
+				})
+			})
+
+			// updatePodWhileRunningScenario updates a pod while the VK is running to ensure the update is propagated
+			// to the provider
```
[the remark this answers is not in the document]
pires: nit: sometimes sentences end with a dot, sometimes not.

node/lifecycle_test.go
```diff
@@ -106,106 +106,118 @@ func TestPodLifecycle(t *testing.T) {
...
-	t.Run("testDanglingPodScenarioWithDeletionTimestamp", func(t *testing.T) {
-		t.Run("mockProvider", func(t *testing.T) {
-			mp := newMockProvider()
-			assert.NilError(t, wireUpSystem(ctx, mp, func(ctx context.Context, s *system) {
-				testDanglingPodScenarioWithDeletionTimestamp(ctx, t, s, mp)
-			}))
-		})
-	})
+			// createStartDeleteScenarioWithDeletionRandomError tests the flow if the pod was unable to be deleted in the
+			// provider
```
pires:
```suggestion
			// provider.
```

node/lifecycle_test.go
```diff
@@ -106,106 +106,118 @@ func TestPodLifecycle(t *testing.T) {
...
-	t.Run("danglingPodScenario", func(t *testing.T) {
-		t.Run("mockProvider", func(t *testing.T) {
-			mp := newMockProvider()
-			assert.NilError(t, wireUpSystem(ctx, mp, func(ctx context.Context, s *system) {
-				testDanglingPodScenario(ctx, t, s, mp)
-			}))
-		})
-	})
+			// createStartDeleteScenarioWithDeletionErrorNotFound tests the flow if the pod was not found in the provider
+			// for some reason
```
pires:
```suggestion
			// for some reason.
```

node/lifecycle_test.go
```diff
@@ -106,106 +106,118 @@ func TestPodLifecycle(t *testing.T) {
...
-		assert.NilError(t, wireUpSystem(ctx, mp, func(ctx context.Context, s *system) {
-			testCreateStartDeleteScenario(ctx, t, s, deletionFunc, false)
-			pods, err := s.client.CoreV1().Pods(testNamespace).List(metav1.ListOptions{})
-			assert.NilError(t, err)
-			assert.Assert(t, is.Len(pods.Items, 1))
-			assert.Assert(t, pods.Items[0].DeletionTimestamp != nil)
-		}))
-	})
+			// createStartDeleteScenario tests the basic flow of creating a pod, waiting for it to start, and deleting
+			// it gracefully
```
pires:
```suggestion
			// it gracefully.
```

node/lifecycle_test.go
```diff
@@ -106,106 +106,118 @@ func TestPodLifecycle(t *testing.T) {
 		return watchErr
 	}
 
-	// isPodDeletedGracefullyFunc is a condition func that waits until the pod is in a  terminal state, which is the VK's
+	// isPodDeletedGracefullyFunc is a condition func that waits until the pod is in a terminal state, which is the VK's
 	// action when the pod is deleted from the provider
```
pires:
```suggestion
	// action when the pod is deleted from the provider.
```

node/podcontroller.go
```diff
@@ -157,6 +160,7 @@ type PodControllerConfig struct {
 	ServiceInformer   corev1informers.ServiceInformer
 }
 
+// NewPodController creates a new pod controller from the provided config
```
pires:
```suggestion
// NewPodController creates a new pod controller with the provided config.
```

node/sync.go
```diff
@@ -0,0 +1,206 @@
...
+	podStatusReasonNotFound          = "NotFound"
+	podStatusMessageNotFound         = "The pod status was not found and may have been deleted from the provider"
+	containerStatusReasonNotFound    = "NotFound"
+	containerStatusMessageNotFound   = "Container was not found and was likely deleted"
+	containerStatusExitCodeNotFound  = -137
+	statusTerminatedReason           = "Terminated"
+	containerStatusTerminatedMessage = "Container was terminated. The exit code may not reflect the real exit code"
+)
+
+// syncProvderWrapper wraps a PodLifecycleHandler to give it async-like pod status notification behavior
```
pires:
```suggestion
// syncProviderWrapper wraps a PodLifecycleHandler to give it async-like pod status notification behavior.
```

node/sync.go
```diff
@@ -0,0 +1,206 @@
...
+type syncProviderWrapper struct {
+	PodLifecycleHandler
+	h func(*corev1.Pod)
+	l corev1listers.PodLister
+
+	// deletedPods makes sure we don't set the "NotFOund" status
```
pires:
```suggestion
	// deletedPods makes sure we don't set the "NotFound" status
```

node/sync.go
```diff
@@ -0,0 +1,206 @@
...
+	// for pods which have been requested to be deleted.
+	// This is needed for our loop which just grabs pod statues every 5 seconds
```
pires:
```suggestion
	// This is needed for our loop which just grabs pod statuses every 5 seconds.
```

node/sync.go
```diff
@@ -0,0 +1,206 @@
...
+	p.h = f
+}
+
+func (p *syncProviderWrapper) _deletePodKey(ctx context.Context, key string) {
+	log.G(ctx).WithField("key", key).Debug("Cleaning up pod from deletion cache")
+	p.deletedPods.Delete(key)
+}
+
+func (p *syncProviderWrapper) DeletePod(ctx context.Context, pod *corev1.Pod) error {
+	log.G(ctx).Debug("syncProviderWrappped.DeletePod")
```
pires:
```suggestion
	log.G(ctx).Debug("syncProviderWrappper.DeletePod")
```

node/sync.go
```diff
@@ -0,0 +1,206 @@
...
+			log.G(ctx).WithFields(map[string]interface{}{
+				"name":      pod.Name,
+				"namespace": pod.Namespace,
+			}).WithError(err).Error("Could not fetch pod status")
+		}
+	}
+}
+
+func (p *syncProviderWrapper) updatePodStatus(ctx context.Context, podFromKubernetes *corev1.Pod) error {
+	ctx, span := trace.StartSpan(ctx, "syncProviderWrapper.getPodStatus")
```
pires:
```suggestion
	ctx, span := trace.StartSpan(ctx, "syncProviderWrapper.updatePodStatus")
```

node/sync.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return err
+	}
+
+	if _, exists := p.deletedPods.Load(key); exists {
+		log.G(ctx).Debug("pod is in known deleted state, ignoring")
+		return nil
+	}
+
+	// Only change the status when the pod was already up
```
pires:
```suggestion
	// Only change the status when the pod was already up.
```

node/sync.go
```diff
@@ -0,0 +1,206 @@
...
+	log.G(ctx).Debug("syncProviderWrappped.DeletePod")
+	key, err := cache.MetaNamespaceKeyFunc(pod)
+	if err != nil {
+		return err
+	}
+
+	p.deletedPods.Store(key, pod)
+	if err := p.PodLifecycleHandler.DeletePod(ctx, pod.DeepCopy()); err != nil {
+		log.G(ctx).WithField("key", key).WithError(err).Debug("Removed key from deleted pods cache")
+		p.deletedPods.Delete(key)
```
pires: So what happens when there's no error? Can you please add a comment describing the intended behavior?

cpuguy83: Added a comment.

node/sync.go
```diff
@@ -0,0 +1,209 @@
...
+		return err
+	}
+
+	p.deletedPods.Store(key, pod)
+	if err := p.PodLifecycleHandler.DeletePod(ctx, pod.DeepCopy()); err != nil {
+		log.G(ctx).WithField("key", key).WithError(err).Debug("Removed key from deleted pods cache")
+		// We aren't going to actually delete the pod from the provider since there is an error so delete it from our cache,
+		// otherwise we could end up leaking pods in our deletion cache.
+		// Delete will be retried by the pod controller.
+		p.deletedPods.Delete(key)
```
sargun: Won't `_deletePodKey()` always get called, thus preventing the leak, even if we don't do our delete here?

node/sync.go
```diff
@@ -0,0 +1,209 @@
...
+			FinishedAt: now,
+		}
+		if cs.State.Running != nil {
+			updated.Status.ContainerStatuses[i].State.Terminated.StartedAt = cs.State.Running.StartedAt
+		}
+	}
+	updated.Status.Reason = statusTerminatedReason
+
+	p.h(updated)
+	log.G(ctx).Error("Notified pod terminal pod status")
```
sargun: Is this an error?

node/sync.go
```diff
@@ -0,0 +1,209 @@
...
+	pods, err := p.l.List(labels.Everything())
+	if err != nil {
+		err = errors.Wrap(err, "error getting pod list from kubernetes")
+		span.SetStatus(err)
+		log.G(ctx).WithError(err).Error("Error updating pod statuses")
+		return
+	}
+	ctx = span.WithField(ctx, "nPods", int64(len(pods)))
+
+	for _, pod := range pods {
```
sargun: Would it make sense to do this on a workqueue?

node/sync.go
```diff
@@ -0,0 +1,209 @@
...
+	containerStatusMessageNotFound   = "Container was not found and was likely deleted"
+	containerStatusExitCodeNotFound  = -137
+	statusTerminatedReason           = "Terminated"
+	containerStatusTerminatedMessage = "Container was terminated. The exit code may not reflect the real exit code"
+)
+
+// syncProviderWrapper wraps a PodLifecycleHandler to give it async-like pod status notification behavior.
+type syncProviderWrapper struct {
+	PodLifecycleHandler
+	h func(*corev1.Pod)
```
sargun: Any reason not to name this `podNotifier`? or `notify`?

node/sync.go
cpuguy83: It is only deleted after the pod has been deleted from the API server.

node/sync.go
cpuguy83: We essentially have the same flow here as before.
What I'm worried about with a work queue is that the number of workers pod sync workers may not be ideal for fetching pod status on a short interval like this and it could end up getting throttled.
We could expose a tune-able specifically for this, but then we are kind of blessing this sync provider implementation.... maybe that's ok, though?

node/lifecycle_test.go
```diff
@@ -106,106 +106,118 @@ func TestPodLifecycle(t *testing.T) {
...
-			assert.NilError(t, wireUpSystem(ctx, newMockProvider(), func(ctx context.Context, s *system) {
-				testCreateStartDeleteScenario(ctx, t, s, isPodDeletedGracefullyFunc, true)
-			}))
-		})
-	})
+	envs := map[string]func() testingProvider{
+		"Async": func() testingProvider {
+			return newMockProvider()
+		},
+		"Sync": func() testingProvider {
```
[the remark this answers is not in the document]
cpuguy83: Done.

PR #787, opened 2019-10-31
Delete pods which are transitioned to a terminal state

sargun:
If the provider generates a pod status update in a terminal state,
(i.e. the pod finishes because the command it was running finished, as
it was a run-to-completion pod), then we need to delete it (gracefully)
from API server.

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -189,6 +189,13 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes 
 		"old reason": podFromKubernetes.Status.Reason,
 	}).Debug("Updated pod status in kubernetes")
 
+	if !running(&podFromProvider.Status) && (podFromProvider.Status.Phase == corev1.PodFailed || podFromProvider.Status.Phase == corev1.PodSucceeded) {
+		log.G(ctx).Debug("Deleting terminal pod from API server")
+		if err := pc.client.Pods(podFromKubernetes.Namespace).Delete(podFromProvider.Name, nil); err != nil {
+			span.SetStatus(err)
+			return pkgerrors.Wrap(err, "error while deleting pod in kubernetes")
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


node/pod.go
```diff
@@ -189,6 +189,13 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
 		"old reason": podFromKubernetes.Status.Reason,
 	}).Debug("Updated pod status in kubernetes")
 
+	if !running(&podFromProvider.Status) && (podFromProvider.Status.Phase == corev1.PodFailed || podFromProvider.Status.Phase == corev1.PodSucceeded) {
```
gabrielhartmann: This condition is unclear.

gabrielhartmann:
Why is the new logic [different than here](https://github.com/virtual-kubelet/virtual-kubelet/blob/master/node/podcontroller.go#L422).

node/pod.go
```diff
@@ -189,6 +189,13 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
...
+		log.G(ctx).Debug("Deleting terminal pod from API server")
+		if err := pc.client.Pods(podFromKubernetes.Namespace).Delete(podFromProvider.Name, nil); err != nil {
```
cpuguy83: Should this be added to the deletion queue?

cpuguy83:
Going to experiment a bit with k8s to see how this is dealt with in kubelet.
Do you have a code pointer for this behavior?

node/pod.go
sargun: No. It'll set deletion timestamp, with the default grace period. When the pod is updated on api server, we'll get the api update back from apiserver, and make a callback to deletepod, and add it to the deletion queue.

sargun:
it's in `kubelet_pods.go`, and `kubelet.go`. AFAICT, the kubelet waits until the resources are "freed" before triggering a delete. We don't really know what "freed" means in this case, so I'm not sure what to do.

PR #789, opened 2019-11-04
After handling status update, reset update timer with correct duration

### LINKED ISSUE #788: Node goes to NotReady status a short time after using NotifyNodeStatus callback
After causing the NotifyNodeStatus callback to be used, the node goes into NotReady status after ~30 seconds and then recovers soon after.

The reason: after updating the node status the periodic update timer is reset, but it is always set to the statusInterval duration (1 minute) even when the periodic update is using Ping, which should be called every 10 seconds.

https://github.com/virtual-kubelet/virtual-kubelet/blob/ba940a9739fe0c661f9edfa3192ad84a412e23a3/node/node.go#L279

Since Ping stops being called for 1 minute the virtual kubelet stops posting status in this time, and kubernetes marks the node as NotReady.

tghartland:
If the ping timer is being used, it should be reset with the ping update
interval. If the status update interval is used then Ping stops being
called for long enough to cause kubernetes to mark the node as NotReady.

Fixes #788

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -244,7 +244,12 @@ func (n *NodeController) controlLoop(ctx context.Context) error {
 
 	statusTimer := time.NewTimer(n.statusInterval)
 	defer statusTimer.Stop()
+	timerResetDuration := n.statusInterval
 	if n.disableLease {
+		// when resetting the timer after processing a status update, reset it to the ping interval
+		// (since it will be the ping timer as n.disableLease == true)
+		timerResetDuration = n.pingInterval
+
 		// hack to make sure this channel always blocks since we won't be using it
 		if !statusTimer.Stop() {
 			<-statusTimer.C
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -259,10 +259,13 @@ func (n *NodeController) controlLoop(ctx context.Context) error {
 			return nil
 		case updated := <-n.chStatusUpdate:
 			var t *time.Timer
+			var resetDuration time.Duration
```
cpuguy83: Maybe we can set this before we start the loop?

tghartland: I've made this change in the new commit, please take a look.

Personally I think it was more obvious why the reset duration had to be changed when it was bundled with this
https://github.com/virtual-kubelet/virtual-kubelet/blob/ba940a9739fe0c661f9edfa3192ad84a412e23a3/node/node.go#L262-L266
logic.

Maybe selecting the active timer could be moved out of the loop as well?

tghartland:
@cpuguy83 I've pushed a new commit with a test that shows this.

Without the fix (ping timer is reset to status timer interval):
```
--- FAIL: TestPingAfterStatusUpdate (0.20s)
    node_test.go:377: assertion failed: expression is false: testP.maxPingInterval < maxAllowedInterval: maximum time between node pings (63.134795ms) was greater than the maximum expected interval (25ms)
time="2019-11-11T14:06:04+01:00" level=debug msg="got node from api server"
FAIL
```

With the fix:
```
--- PASS: TestPingAfterStatusUpdate (0.20s)
PASS
```

PR #792, opened 2019-11-13
bump golang to version 1.13.4

ducksecops:


sargun:
Do you know if k8s is tested with 1.13.4?

sargun:
@cpuguy83 Even though this is out of date, I'm inclined to rebase this, push to 1.13.9, and pull. What do you think?

sargun:
This also breaks updates.

cpuguy83:
Yes we should probably bump since 1.12 is no longer supported, but we may also need to update client k8s libs.

sargun:
@cpuguy83 v1.16?

sargun:
Scratch that. Kubernetes is still on Go 1.12. I do not suggest we upgrade past the version they are on.

PR #793, opened 2019-11-14
[Sync Provider] Fix panic on not found pod status

cpuguy83:


node/sync.go
```diff
@@ -202,6 +204,11 @@ func (p *syncProviderWrapper) updatePodStatus(ctx context.Context, podFromKubern
 		log.G(ctx).Debug("Setting pod not found on pod status")
 	}
 
+	if podStatus == nil {
+		log.G(ctx).WithError(statusErr).Debug("Skipping pod status update due to missing pod status")
```
sargun: This seems like a higher-level message than debug?

cpuguy83: Sure, I updated this to actually return the error.

node/sync.go
```diff
@@ -176,32 +179,35 @@ func (p *syncProviderWrapper) updatePodStatus(ctx context.Context, podFromKubern
 		return nil
 	}
 
+	if podFromKubernetes.Status.Phase != corev1.PodRunning && time.Since(podFromKubernetes.ObjectMeta.CreationTimestamp.Time) <= time.Minute {
+		span.SetStatus(err)
```
sargun: Shouldn't this be `span.SetStatus(statusErr)`, because `err` gets rewritten by `cache.MetaNamespaceKeyFunc`?

sargun:
It might be worth adding "Don't return nil pod status" to the docs.

node/sync.go
cpuguy83: :( Yes. Thanks

cpuguy83:
Yeah. In this case at least it's an error that's returned with the nil status I just messed up here.

PR #794, opened 2019-11-15
Cherry picks 1.2.1

cpuguy83:


cpuguy83:
This is the same as master, so merging.

PR #801, opened 2019-12-20
Handle providers which do not implement UpdatePod

sargun:
This addresses the issue where certain (optional) APIs are unimplemented by the provider.

### DIFF
--- a/errdefs/unsupported.go
+++ b/errdefs/unsupported.go
@@ -1,0 +1,65 @@
+package errdefs
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ErrUnsupported is an error interface which denotes whether the operation failed due
+// to the operation being unsupported. It will not succeed on subsequent retries.
+type ErrUnsupported interface {
+	Unsupported() bool
+	error
+}
+
+type unsupportedError struct {
+	error
+}
+
+func (e *unsupportedError) Unsupported() bool {
+	return true
+}
+
+func (e *unsupportedError) Cause() error {
+	return e.error
+}
+
+// AsUnsupported wraps the passed in error to make it of type ErrUnsupported
+//
+// Callers should make sure the passed in error has exactly the error message
+// it wants as this function does not decorate the message.
+func AsUnsupported(err error) error {
+	if err == nil {
+		return nil
+	}
+	return &unsupportedError{err}
+}
+
+// Unsupported makes an ErrUnsupported from the provided error message
+func Unsupported(msg string) error {
+	return &unsupportedError{errors.New(msg)}
+}
+
+// Unsupportedf makes an ErrUnsupported from the provided error format and args
+func Unsupportedf(format string, args ...interface{}) error {
+	return &unsupportedError{fmt.Errorf(format, args...)}
+}
+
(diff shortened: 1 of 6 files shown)


node/pod.go
```diff
@@ -70,7 +71,7 @@ func (pc *PodController) createOrUpdatePod(ctx context.Context, pod *corev1.Pod)
 	if podFromProvider, _ := pc.provider.GetPod(ctx, pod.Namespace, pod.Name); podFromProvider != nil {
 		if !podsEqual(podFromProvider, podForProvider) {
 			log.G(ctx).Debugf("Pod %s exists, updating pod in provider", podFromProvider.Name)
-			if origErr := pc.provider.UpdatePod(ctx, podForProvider); origErr != nil {
+			if origErr := pc.provider.UpdatePod(ctx, podForProvider); origErr != nil && !errdefs.IsUnsupported(origErr) {
```
cpuguy83: Can we make sure not to log that the pod was updated in the unsupported case?

We may also want to have the queue handler deal with unsupported by just not re-queuing instead of returning no error. WDYT?

sargun: I'm worried if we do it in the queue handler for all queues, then it'll be unable to differentiate isunsupported from things which are allowed to be unsupported versus those which are not.

Maybe we should do something like K8s, and have the ability to signal how retryable errors are? https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L619

cpuguy83: When should an `Unsuported` error be ok to re-queue?

sargun: Never, but no provider should ever have something like CreatePod unsupported.

cpuguy83: No working provider, but it will be obvious that things aren't working by looking at the logs (And the fact that nothing works :) ).

cpuguy83: I still think we should return the error here.
The provider didn't perform the action and callers need to know that information.

node/lifecycle_test.go
```diff
@@ -247,19 +283,51 @@ func wireUpSystem(ctx context.Context, provider PodLifecycleHandler, f testFunct
 
 	// Create the fake client.
 	client := fake.NewSimpleClientset()
+	tracker := client.Tracker()
```
sargun: It turns out our reactor update / checking logic was incorrect.

cpuguy83: Can you explain?

sargun: 1. We never checked if the ResourceVersion was set correctly on update (check before set)
2. Although we incremented the ResourceVersion, we never actually did anything with that, since we returned "false" to handled.

errdefs/unsupported.go
```diff
@@ -0,0 +1,65 @@
+package errdefs
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ErrUnsupported is an error interface which denotes whether the operation failed due
+// to the operation being unsupported. It will not succeed on subsequent retries.
+type ErrUnsupported interface {
```
cpuguy83, line 10: Pedantic request... WDYT about renaming this to `NotImplemented`?

Otherwise this LGTM.

Issue #791, opened 2019-11-11
kubectl exec api conn idleTimeout 30s too short.


lauly:
https://github.com/virtual-kubelet/virtual-kubelet/blob/master/node/api/exec.go
This code file ###idleTimeout := time.Second * 30### is too short,Can I config the idleTimeout param?

cpuguy83:
If the Kubelet API has a param for this we could add support for it.
Otherwise, the design of this package is such that you can bring your own handlers.

cwdsuzhou:
I have the same issue

/help

PR #803, opened 2020-01-13
Make exec timeout configurable

cwdsuzhou:
Make exec timeout configurable

See 
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/types.go#L165
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L2296

kubelet also allows to change default steamIdleTimeout, so I think it's to reasonable to make it configurable

Fix: https://github.com/virtual-kubelet/virtual-kubelet/issues/791

### DIFF
--- a/node/api/exec.go
+++ b/node/api/exec.go
@@ -123,7 +123,7 @@ func HandleContainerExec(h ContainerExecHandlerFunc, opts ...ContainerExecHandle
 			streamOpts,
 			cfg.StreamIdleTimeout,
 			cfg.StreamCreationTimeout,
-			supportedStreamProtocols.
+			supportedStreamProtocols,
 		)
 
 		return nil
--- a/node/api/exec.go
+++ b/node/api/exec.go
@@ -53,10 +53,10 @@ type TermSize struct {
 // Note that this handler currently depends on gorrilla/mux to get url parts as variables.
 // TODO(@cpuguy83): don't force gorilla/mux on consumers of this function
 // ContainerExecHandlerConfig is used to pass options to options to the container exec handler.
-type ContainerExecHandlerConfig {
+type ContainerExecHandlerConfig struct {
 	// StreamIdleTimeout is the maximum time a streaming connection
 	// can be idle before the connection is automatically closed.
-	StreamCreationTimeout time.Duration
+	StreamIdleTimeout time.Duration
 	// StreamCreationTimeout is the maximum time for streaming connection
 	StreamCreationTimeout time.Duration
 }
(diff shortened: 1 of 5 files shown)


cwdsuzhou:
@cpuguy83 PTAL

cpuguy83:
Thanks!

One thing I'd like is if this can be changed so we aren't breaking callers of these functions.
Maybe we can have it accept something like a functional options `... func(*Config)`

node/api/server.go
```diff
@@ -39,7 +39,7 @@ type PodHandlerConfig struct {
 }
 
 // PodHandler creates an http handler for interacting with pods/containers.
-func PodHandler(p PodHandlerConfig, debug bool) http.Handler {
+func PodHandler(p PodHandlerConfig, debug bool, cfg ...StreamTimeOutConfig) http.Handler {
```
cpuguy83: Sorry I was thinking something like functional options so `...func(*Config)` And then you can provide a list of functions to set the config... e.g. `WithStreamIdleTimeout(dur)`.

But I think we could probably just add a stream config to the `PodHandlerConfig`

cwdsuzhou: Updated, thanks

node/api/exec.go
```diff
@@ -52,7 +52,8 @@ type TermSize struct {
 // HandleContainerExec makes an http handler func from a Provider which execs a command in a pod's container
 // Note that this handler currently depends on gorrilla/mux to get url parts as variables.
 // TODO(@cpuguy83): don't force gorilla/mux on consumers of this function
-func HandleContainerExec(h ContainerExecHandlerFunc) http.HandlerFunc {
+func HandleContainerExec(h ContainerExecHandlerFunc, streamIdleTimeout, streamCreationTimeout time.Duration) http.
```
cpuguy83: So this is where we need to provide some options, but we want to do it in a non-breaking way.

So this function can be updated to look like this:


```suggestion
// ContainerExecHandlerConfig is used to pass options to options to the container exec handler.
type ContainerExecHandlerConfig {
	// StreamIdleTimeout is the maximum time a streaming connection
	// can be idle before the connection is automatically closed.
	StreamCreationTimeout time.Duration
	// StreamCreationTimeout is the maximum time for streaming connection
	StreamCreationTimeout time.Duration
}

// ContainerExecHandlerOption configures a ContainerExecHandlerConfig
// It is used as functional options passed to `HandleContainerExec`
type ContainerExecHandlerOption func(*ContainerExecHandlerConfig)

// WithExecStreamIdleTimeout sets the idle timeout for a container exec stream
func WithExecStreamIdleTimeout(dur time.Duration) ContainerExecHandlerOption {
	return func(cfg *ContainerExecHandlerConfig) {
		cfg.StreamIdleTimeout = dur
	}
}

// HandleContainerExec makes an http handler func from a Provider which execs a command in a pod's container
// Note that this handler currently depends on gorrilla/mux to get url parts as variables.
// TODO(@cpuguy83): don't force gorilla/mux on consumers of this function
func HandleContainerExec(h ContainerExecHandlerFunc, opts ..ContainerExecHandlerOption) HandlerFunc {
    if h == nil {
    	return NotImplemented
    }
    
    var cfg ContainerExecHandlerConfig
    for _, o := range opts {
    	o(&cfg)
    }
```

When things like `PodHandler` have a timeout set it can call `HandleContainerExec` with functions to set these options.

cpuguy83: This would be in addition to what you already have in this PR.

cwdsuzhou:
Updated

PR #806, opened 2020-01-20
Add support for v1.PodLogOptions

ldx:
Check if "follow" is requested for container logs, and pass it to the
provider so tailing logs can be supported.

### DIFF
--- a/node/api/logs.go
+++ b/node/api/logs.go
@@ -16,8 +16,10 @@ package api
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cpuguy83:
From what I can tell in the kubelet, there is no supported query param for `follow`.
https://github.com/kubernetes/kubernetes/blob/15c3f1b119ff4b4073c86df05c819a9672b80d68/pkg/kubelet/server/server.go#L542-L583

ldx:
There is: https://github.com/kubernetes/kubernetes/blob/15c3f1b119ff4b4073c86df05c819a9672b80d68/staging/src/k8s.io/api/core/v1/types.go#L4808

If you use `kubectl logs -f`, `follow=true` will be present in the API request query.

cpuguy83:
Right so we need to decode the log options from the request.

ldx:
> Right so we need to decode the log options from the request.

Yes, the same way as with `tailLines`.

Thinking a bit more about this: we could even add support for all fields in `v1.PodLogOptions` here in one go. Let me know, and I can refactor this PR to do that.

node/api/logs.go
```diff
@@ -37,6 +37,7 @@ type ContainerLogOpts struct {
 	Since      time.Duration
 	LimitBytes int
 	Timestamps bool
+	Follow     bool
```
cwdsuzhou: we  can not just add Follow option, since the current implementation actually do not support it

ldx: @cwdsuzhou you mean providers don't support it? Ours definitely can once this field is in the struct.

cpuguy83:
Yes let's support v1.PodLogOptions and convert that to our log config.

Thank you!

ldx:
@cpuguy83 I refactored this to include all fields from v1.PodLogOptions.

node/api/logs.go
cwdsuzhou: Maybe something wrong with mysql implementation. But when I change some logic of `vk`, it works.

ldx:
@cpuguy83 ping - anything else I can do?

node/api/logs.go
```diff
@@ -33,10 +35,71 @@ type ContainerLogsHandlerFunc func(ctx context.Context, namespace, podName, cont
...
+}
+
+func parseLogOptions(q url.Values) (opts ContainerLogOpts, err error) {
+	if tailLines := q.Get("tailLines"); tailLines != "" {
+		opts.Tail, err = strconv.Atoi(tailLines)
+		if err != nil {
+			return opts, errdefs.AsInvalidInput(errors.Wrap(err, "could not parse \"tailLines\""))
+		}
+		if opts.Tail < 0 {
+			return opts, errdefs.InvalidInput(fmt.Sprintf("\"tailLines\" is %d", opts.Tail))
```
cpuguy83, line 54: Can you use `InvalidInputf` instead of `fmt.Sprintf`?

cpuguy83: There are a few of these.

ldx: Sure, let me fix this :+1:

ldx: @cpuguy83 done.

Issue #807, opened 2020-02-04
NodeConditions are Overwritten During Node Patch


gabrielhartmann:
---

### Environment summary

Provider (e.g. ACI, AWS Fargate): `N/A`

Version (e.g. 0.1, 0.2-beta): `v1.1.1-0.20191101162338-ba940a9739fe`

K8s Master Info (e.g. AKS, ACS, Bare Metal, EKS): Bare Metal

Install Method (e.g. Helm Chart, ): Custom

### Issue Details

Node Conditions set on a Node Status object are overwritten by Virtual Kubelet node patches.  The patch mechanism isn't working, and instead is being treated like a forced update.  Using the `node-problem-detector` provides an easy way to show this behavior.  Here are example logs from a custom node problem definition.

```
time="2020-02-04T00:21:23Z" level=info msg="got node from api server" node.Status.Conditions="[{Ready True 2020-02-04 00:21:13 +0000 UTC 2020-02-04 00:21:13 +0000 UTC KubeletReady kubelet is ready.}]"
time="2020-02-04T00:21:23Z" level=info msg="updated node status in api server" node.Status.Conditions="[{Ready True 2020-02-04 00:21:23 +0000 UTC 2020-02-04 00:21:13 +0000 UTC KubeletReady kubelet is ready.}]" node.resourceVersion=14836248
time="2020-02-04T00:21:33Z" level=info msg="got node from api server" node.Status.Conditions="[{Ready True 2020-02-04 00:21:23 +0000 UTC 2020-02-04 00:21:13 +0000 UTC KubeletReady kubelet is ready.}]"
time="2020-02-04T00:21:33Z" level=info msg="updated node status in api server" node.Status.Conditions="[{Ready True 2020-02-04 00:21:33 +0000 UTC 2020-02-04 00:21:13 +0000 UTC KubeletReady kubelet is ready.}]" node.resourceVersion=14836255
time="2020-02-04T00:21:43Z" level=info msg="got node from api server" node.Status.Conditions="[{Ready True 2020-02-04 00:21:33 +0000 UTC 2020-02-04 00:21:13 +0000 UTC KubeletReady kubelet is ready.}]"
time="2020-02-04T00:21:43Z" level=info msg="updated node status in api server" node.Status.Conditions="[{Ready True 2020-02-04 00:21:43 +0000 UTC 2020-02-04 00:21:13 +0000 UTC KubeletReady kubelet is ready.}]" node.resourceVersion=14836262
[...]

PR #808, opened 2020-02-04
Patch NodeStatus conserving NodeConditions

gabrielhartmann:
Resolves Issue #807
Borrows technique from node-problem-detector:
https://github.com/kubernetes/node-problem-detector/blob/a442e71190e7d460c892474336c38a412b332be1/pkg/exporters/k8sexporter/problemclient/problem_client.go#L126-L133

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -28,7 +28,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/kubernetes/typed/coordination/v1beta1"
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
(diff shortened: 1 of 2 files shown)


gabrielhartmann:
I don't believe this change regresses the test introduced here: https://github.com/virtual-kubelet/virtual-kubelet/pull/557/files

node/node.go
```diff
@@ -408,11 +407,12 @@ var emptyGetOptions = metav1.GetOptions{}
 
 // patchNodeStatus patches node status.
 // Copied from github.com/kubernetes/kubernetes/pkg/util/node
```
sargun: Can you borrow the latest code from kubernetes upstream? https://github.com/kubernetes/kubernetes/blob/master/pkg/util/node/node.go#L175-L206

sargun: Or: https://github.com/kubernetes/kubernetes/blob/master/pkg/util/node/node.go#L256-L268

sargun:
Can you verify node conditions are preserved?

gabrielhartmann:
In my experiments this works to a degree.  `NodeCondition` objects are no longer entirely removed by status updates from VK.  However the result of the patch consistently modify out of band NodeConditions.  Particularly their `Status`.

gabrielhartmann:
If the test added here is run on `master` the patchBytes generated are:
```
{
	"status": {
		"$setElementOrder/conditions": [{
			"type": "Ready"
		}],
		"conditions": [{
			"$patch": "delete",
			"type": "CustomProblem"
		}]
	}
}
```

pires:
I couldn't go through all the code in enough detail to revive my memory of how we implemented the node controller, but IIRC we store the `Node` in the node controller and do not react to node update events, hence the node controller wouldn't know about changes made by others.

PR #817, opened 2020-03-09
add 'GET' method to pod exec handler, so it can support websocket

guoliangshuai:
I found a BUG in virtual-kubelet:
users trying to do pod exec with WebSocket will end with 405, while pod exec with SPDY will be ok.
The reason is:

WebSocket usually starts handshake with a http request of 'GET'
pod exec handler in virtual-kubelet only register 'POST' method
PS：code of pod exec handler in virtual-kubelet actually suport both WebSocket and SPDY


guoliangshuai:
I signed it

Issue #820, opened 2020-03-16
VK do not support `logs --since` and `logs --limitBytes`


cwdsuzhou:
VK do not support `logs --since` and `logs --limitBytes`

PR #821, opened 2020-03-16
Bugfix: since and limitBytes unsupported

cwdsuzhou:
Fix https://github.com/virtual-kubelet/virtual-kubelet/issues/820

### DIFF
--- a/node/api/logs.go
+++ b/node/api/logs.go
@@ -16,15 +16,20 @@ package api
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
-	"strconv"
+	"net/url"
 	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
 	"github.com/virtual-kubelet/virtual-kubelet/errdefs"
 	"github.com/virtual-kubelet/virtual-kubelet/log"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/pkg/api/legacyscheme"
+	"k8s.io/kubernetes/pkg/apis/core/v1/validation"
 )
 
 // ContainerLogsHandlerFunc is used in place of backend implementations for getting container logs
@@ -37,6 +42,10 @@ type ContainerLogOpts struct {
 	Since      time.Duration
 	LimitBytes int
 	Timestamps bool
+}
+
+func init() {
+	legacyscheme.Scheme.AddKnownTypes(v1.SchemeGroupVersion, &v1.PodLogOptions{})
 }
 
 // HandleContainerLogs creates an http handler function from a provider to serve logs from a pod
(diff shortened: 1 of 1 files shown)


cwdsuzhou:
ping @cpuguy83  PTAL

cpuguy83:
This should go into github.com/virtual-kubelet/node-cli
The cmd in this repo is only used for e2e tests.

Thanks!

cpuguy83:
Argh... wrong tab....

cpuguy83:
Can we use what kubelet is doing?

```go
	query := request.Request.URL.Query()
	// backwards compatibility for the "tail" query parameter
	if tail := request.QueryParameter("tail"); len(tail) > 0 {
		query["tailLines"] = []string{tail}
		// "all" is the same as omitting tail
		if tail == "all" {
			delete(query, "tailLines")
		}
	}
	// container logs on the kubelet are locked to the v1 API version of PodLogOptions
	logOptions := &v1.PodLogOptions{}
	if err := legacyscheme.ParameterCodec.DecodeParameters(query, v1.SchemeGroupVersion, logOptions); err != nil {
		response.WriteError(http.StatusBadRequest, fmt.Errorf(`{"message": "Unable to decode query."}`))
		return
	}
	logOptions.TypeMeta = metav1.TypeMeta{}
	if errs := validation.ValidatePodLogOptions(logOptions); len(errs) > 0 {
		response.WriteError(http.StatusUnprocessableEntity, fmt.Errorf(`{"message": "Invalid request."}`))
		return
	}
```

cpuguy83:
I should clarify, I don't want to pull in a bunch of k8s packages do accomplish this (b/c this tends to be an insane amount of code), but I do want to make sure we are decoding everything we should be and with proper API versioning.

cwdsuzhou:
> I should clarify, I don't want to pull in a bunch of k8s packages do accomplish this (b/c this tends to be an insane amount of code), but I do want to make sure we are decoding everything we should be and with proper API versioning.

Got it! But we have exposed `ContainerLogOpts` to  users, so if we want to support `PodLogOptions`, we have to adapt to it.

cwdsuzhou:
done @cpuguy83

node/api/logs.go
```diff
@@ -39,6 +44,10 @@ type ContainerLogOpts struct {
 	Timestamps bool
 }
 
+func init() {
+	legacyscheme.Scheme.AddKnownTypes(v1.SchemeGroupVersion, &v1.PodLogOptions{})
```
cpuguy83, line 48: Let's not use `func init()`.

Also is there a way to not use a singleton like this?

cpuguy83:
This PR seems similar to #806
I do think I'd prefer to manually parse the request over pulling in more from k8s/api.

cwdsuzhou:
> This PR seems similar to #806
> I do think I'd prefer to manually parse the request over pulling in more from k8s/api.

Seems similar, but I would like use the  existing codes to keep vk and kubelet have the same behaviors. If kubelet's behavior updates, vk can automatically update.

cpuguy83:
I think I'd prefer #806 just because I don't want to pull in a bunch of packages from k8s.
In fact I'd like to reduce what we can that's not in the core types or client packages.

Issue #822, opened 2020-03-16
Add flags to expose qps


cwdsuzhou:
VK as the entrance of a cluster, may run many pods on it. Default QPS setting of k8s client is obviously not enough for large-cluster, we need expose the qps and burst, e.g. the case in our cluster.

![image](https://user-images.githubusercontent.com/12230714/76755931-b4536480-67bf-11ea-99fb-f36f77f0cd89.png)

PR #824, opened 2020-03-18
Check pods status deep equal before update

cwdsuzhou:
Check pods status deep equal before update. If we do not check it, we would update pod status frequently and reach the qps limit quickly.

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -213,6 +213,10 @@ func (pc *PodController) enqueuePodStatusUpdate(ctx context.Context, q workqueue
 		if obj, ok := pc.knownPods.Load(key); ok {
 			kpod := obj.(*knownPod)
 			kpod.Lock()
+			if cmp.Equal(kpod.lastPodStatusReceivedFromProvider, pod) {
+				kpod.Unlock()
+				return
+			}
 			kpod.lastPodStatusReceivedFromProvider = pod
 			kpod.Unlock()
 			q.AddRateLimited(key)


node/pod.go
```diff
@@ -185,6 +186,9 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
 	kPod.Lock()
 	podFromProvider := kPod.lastPodStatusReceivedFromProvider.DeepCopy()
 	kPod.Unlock()
+	if reflect.DeepEqual(podFromKubernetes.Status, podFromProvider.Status) {
```
sargun: Any reason not to use cmp here?

cwdsuzhou: Thanks for review. 

Good idea, updated

cwdsuzhou:
ping @cpuguy83 @sargun , is this PR ready to be merged

node/pod.go
```diff
@@ -185,6 +185,9 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
 	kPod.Lock()
 	podFromProvider := kPod.lastPodStatusReceivedFromProvider.DeepCopy()
 	kPod.Unlock()
+	if cmp.Equal(podFromKubernetes.Status, podFromProvider.Status) {
```
sargun, line 188: I think we keep track of the last pod we received from provider. Can we do the deduping there? That we look at the last pod from we received from the provider, and if it wasn't the same, then we dedupe? Otherwise, this mechanism can lead to "false updates" and "missed updates" due to lag.

Also, I think this behaviour should be behind a config, because right now, firing away an update is a way to force a pod status update from VK -- which a provider may want to do if something interferes with the write to pod status.

cwdsuzhou: IMO, `lastPodStatusReceivedFromProvider` seems not like a good design. As you mentioned, that may cause `false updates` and `false updates`.  Why not we change it to a `map`, not just a variable.

cwdsuzhou: https://github.com/virtual-kubelet/virtual-kubelet/blob/master/node/pod.go#L180
We have checked here, so it would not directly lead to "false updates" or "missed updates". This commit just add one more check before updating status to K8S.

@cpuguy83  wdyt?

sargun: I think that without some level of checking here, you can get into a situation where you can miss updates. 

For example, if you have a pod which has a start, and a container fails, and it's restarted? You could fail to send the successful start because you haven't received the update from API server indicating it's failed?

cwdsuzhou: If restart, pod running time would change. Other messages in `status.conditions` would also change. What's more, if pods' status has not changed, why we need update the status to apiserver.

As you said, `start- fail - restart`, update status or not would not change the result.

sargun: Are the timestamps generated client side, or server side? I think client-side?

I think that although Kubernetes is a line-level system, people could be depending on this behaviour. 

I think it makes much more sense to compare it to the last pod we "sent" in the NotifyPods callback if we really want to dedupe. That way the behaviour is not breaking.

cwdsuzhou: I still think we should compare it here. If status has not changed, updating it to api-server is meaningless. And this issue has leaded to the rate limit of client-go https://github.com/virtual-kubelet/virtual-kubelet/issues/822

sargun: What's your reasoning against suppressing duplicate pod status updates in the NotifyPods callback?

cwdsuzhou: No, I am not against this. I think we'd add check both in the `NotifyPods` and `UpdatePodStatus`

sargun: Do you want to add it to NotifyPods? And then we can hash this one out later?

cwdsuzhou: > Do you want to add it to NotifyPods? And then we can hash this one out later?

Sure, done

node/pod.go
```diff
@@ -213,7 +216,9 @@ func (pc *PodController) enqueuePodStatusUpdate(ctx context.Context, q workqueue
 		if obj, ok := pc.knownPods.Load(key); ok {
 			kpod := obj.(*knownPod)
 			kpod.Lock()
-			kpod.lastPodStatusReceivedFromProvider = pod
+			if !cmp.Equal(kpod.lastPodStatusReceivedFromProvider, pod) {
```
sargun, line 219: If cmp.equal, then we don't need to call q.AddRateLimited below.

node/pod.go
sargun: Do you want to do the pod update dedupe in a different PR, or drop this dedupe behaviour in this PR to get it through?

sargun: In regards to:
> IMO, lastPodStatusReceivedFromProvider seems not like a good design. As you mentioned, that may cause false updates and false updates. Why not we change it to a map, not just a variable.

I think that there are two reasons behind this design (and I will not be one to defend it as being perfect):

1. We want thee ability to perform critical code operations on the k8s pod, and the provider pod under the same lock.
2. From a performance perspective, having to hold a *global* lock while doing these (relatively) simple operations gets slow quick.

node/pod.go
cwdsuzhou: Yes, thanks

node/pod.go
cwdsuzhou: > Do you want to do the pod update dedupe in a different PR, or drop this dedupe behaviour in this PR to get it through?

Thanks, I will move `pod update dedupe` to a different PR

node/pod.go
cwdsuzhou: Done, PTAL

node/pod.go
```diff
@@ -213,8 +213,11 @@ func (pc *PodController) enqueuePodStatusUpdate(ctx context.Context, q workqueue
 		if obj, ok := pc.knownPods.Load(key); ok {
 			kpod := obj.(*knownPod)
 			kpod.Lock()
+			defer kpod.Unlock()
+			if cmp.Equal(kpod.lastPodStatusReceivedFromProvider, pod) {
+				return
+			}
 			kpod.lastPodStatusReceivedFromProvider = pod
-			kpod.Unlock()
 			q.AddRateLimited(key)
```
sargun, line 221: Dumb question, do you know if this function blocks?

node/pod.go
```diff
@@ -213,8 +213,11 @@ func (pc *PodController) enqueuePodStatusUpdate(ctx context.Context, q workqueue
 		if obj, ok := pc.knownPods.Load(key); ok {
 			kpod := obj.(*knownPod)
 			kpod.Lock()
+			defer kpod.Unlock()
+			if cmp.Equal(kpod.lastPodStatusReceivedFromProvider, pod) {
```
sargun, line 217: Do we need to compare the entire pod, or just the pod status?

sargun:
Thank you for working through this. I appreciate the contribution.

node/pod.go
cwdsuzhou: I think it may be safer to compare the entire pod, which would not change the original behavior.

node/pod.go
cwdsuzhou: I check the implementation. This `q` is base on a `chan` with size `1000`. So if not full, would not block.

cwdsuzhou:
> Thank you for working through this. I appreciate the contribution.

Thank, I would like open another PR to add check in `UpdatePodStatus`

node/pod.go
cwdsuzhou: > In regards to:
> 
> > IMO, lastPodStatusReceivedFromProvider seems not like a good design. As you mentioned, that may cause false updates and false updates. Why not we change it to a map, not just a variable.
> 
> I think that there are two reasons behind this design (and I will not be one to defend it as being perfect):
> 
> 1. We want thee ability to perform critical code operations on the k8s pod, and the provider pod under the same lock.
> 2. From a performance perspective, having to hold a _global_ lock while doing these (relatively) simple operations gets slow quick.

Do we have a plan to change this logic, e.g. a map `map[pod.UID]pod` to make it possible to process sync pod status in parallel.

cwdsuzhou:
@sargun @cpuguy83 could you help merge this PR?  thanks

sargun:
I’ll wait a day in case @cpuguy83 has any issues. Is there a way to write a test for this by any chance?

node/pod.go
```diff
@@ -213,8 +213,11 @@ func (pc *PodController) enqueuePodStatusUpdate(ctx context.Context, q workqueue
 		if obj, ok := pc.knownPods.Load(key); ok {
 			kpod := obj.(*knownPod)
 			kpod.Lock()
+			defer kpod.Unlock()
```
cpuguy83, line 216: This defer could keep `kpod` locked if `AddRateLimited` is blocked.
I would prefer if we play this safer and just unblock immediately after `cmp.Equal`

cwdsuzhou: Seems safer, done!

cwdsuzhou:
> I’ll wait a day in case @cpuguy83 has any issues. Is there a way to write a test for this by any chance?

I go through the test codes, it seems hard now. I prefer to open another PR to add tests for this

cwdsuzhou:
@sargun @cpuguy83  can you take a look at this PR 
https://github.com/virtual-kubelet/virtual-kubelet/pull/830

PR #825, opened 2020-03-20
Add /pods HTTP endpoint

sargun:


node/api/server.go
```diff
@@ -36,7 +36,8 @@ type ServeMux interface {
 type PodHandlerConfig struct {
 	RunInContainer        ContainerExecHandlerFunc
 	GetContainerLogs      ContainerLogsHandlerFunc
-	GetPods               PodListerFunc
+	GetPodsFromProvider   PodListerFunc
```
cpuguy83: I don't think we should change the name of this field. Maybe add a doc string?

sargun:
Still working on this PR. I'll ask for re-review once things are ready.

sargun:
@cpuguy83 I made it so we just re-use the podlister that the pod controller has.

PR #827, opened 2020-03-27
Refactor environment variable population to library functions

justnoise:
I'm working through how to create a library for modifying environment variables. This relates to https://github.com/virtual-kubelet/virtual-kubelet/pull/650 and makes it possible to defer downward API resolution (or any env var resolution) to the provider.  If this isn't the correct place or protocol to discuss such changes, let me know and I'll holler elsewhere.

My thinking, to allow the greatest amount of control and reuse for clients, is to address this in 2 parts:

1. Refactor populating environment variables into reusable components functions. These are available for providers to use (package `/env`).
2. Make these functions pluggable by the user in the PodController. I like the idea of using appliers.

This draft PR addresses Part 1.  I'd love some feedback on it before writing code and tests.

Each of the following functions go through all initContainers and containers in a pod and replace environment variable value references and perform expansions.  Each function iterates through all containers in a pod which allows each function to cache resources returned from the lister (similar to kubelet_pods.go).  I'm not sure how much caching is actually needed since the underlying listers/informers should cache those structures as well (I think?).  I'd love to hear opinions or corrections if I've misunderstood client-go.  To resolve all environment variables, you'll need to take up to 7 passes through the env var list to resolve all environment variables.  I find some solace in the hope that iterating through contiguous items in a processor cache is blazingly fast.

The kubelet is pretty specific about the order of operations to resolve environment variables and we are transposing those operations.  If the following functions are applied in the order they're listed here the result should mirror the end result of the kubelet's implementation.

```go
// Iterate through all containers EnvFrom and Env, insert resolved
// EnvFrom items at the start of the env.  Replace
// Env.ConfigMapKeyRefs with the resolved value.
env.ResolveConfigMapRefs(ctx, pod, configMapLister) error

// Iterate through all containers EnvFrom and Env, insert resolved
// EnvFrom items at the start of the env.  Replace Env.SecretKeyRefs
// with the resolved value.
env.ResolveSecretRefs(ctx, pod, secretLister) error

// Iterate through all containers Env, replace all FieldRef values
// with their values
env.ResolveFieldRefs(pod) error

// Iterate through all containers Env, replace all ResourceFieldRef
// values with the values from the node if it exists or from the
// container if that is referenced (follow the same behavior as
// k8s.io/pkg/kubelet/kubelet_pods.go).
env.ResolveResourceRefs(*pod, *node) error

// Resolves service env vars and inserts them to the *front* of the
// env var list if a variable by the same key doesn't already exist in
// the list of env vars (see:
// https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_pods.go#L749-L754).
// We insert onto the front of the list since service env vars are
// needed for resolving env var expansions.
env.InsertServiceEnvVars(ctx, pod, serviceLister) error

// go through all env vars and resolve any $(VAR_NAME) expansions
// using only the values listed before the env var.
env.ResolveEnvVarExpansions(pod) error

// remove duplicate env vars in each container by only keeping the
// latest keys
env.Uniqify(*pod)
```

For part 1, `populateEnvironmentVariables()` would be updated to call those 7 functions in order.


cpuguy83:
I'd like to not do `Applier`s just because it's introducing more things into the runtime that don't really need to be there, at least not for applying downward API.

In the runtime we can make downward API optional. We can still apply it by default but not make it a requirement as it is today.

We can also provide a single go-to default iterator for applying the downward API to a pod in the "correct" order.

But otherwise this proposal LGTM.

justnoise:
> We can also provide a single go-to default iterator for applying the downward API to a pod in the "correct" order.

Can you briefly outline what an iterator might look like?  Is this more like a callback or interface from the user that's passed into the PodController to resolve (or postpone processing) downward API env vars?  Maybe a method the provider can implement?

cpuguy83:
@justnoise Sorry I mean a a wrapper function that does everything in a specific order.

```go
func ResolveEnv(context.Context, pod *v1.Pod, config ResolveConfig) {
   ResolveConfigMapRefs(ctx, pod, config.ConfigMapLister)
   ResolveSecretRefs(ctx, pod, config.SecretsLister)
   // ... 
}
```

We would call the same function from within the runtime for backwards compatibility.

cpuguy83:
Closing since it seems like #829 replaces this one.

Issue #649, opened 2019-06-04
Wasteful envvar population on podspec update check. 


yashdesai93:
Currently we populate env vars every time we want to check is an existing pod's podspec has been updated. 
here: https://github.com/virtual-kubelet/virtual-kubelet/blob/master/vkubelet/pod.go#L46 

This is wasteful, since we don't allow env-vars to change.
It is also incorrect behavior, since configMaps/secrets might have been updated/deleted
since the first `CreatePod` call, and it can cause update of other podspec fields to fail. 
Even in the case of optional envs, we will create false events. 

one scenario:
create pod on k8s master with configMap as envvar
vkubelet: populate env var from configMap
provider.CreatePod (pod)
pod is running
delete configmap
update pod's spec.activeDeadlineSeconds
vkubelet: populate env var fails, abort update.

A possible simple fix would be to call `populateEnvironmentVariables` only during creation
and instead of updating the pod spec itself, return a map of envvars, which is passed to the provider in the `CreatePod` call.

cpuguy83:
I think it's fine to change this to only populate on create.
However we should make sure the call to `UpdatePod` in the provider copies the existing pod env to the updated pod object.

jipperinbham:
For anyone else that comes across this issue, we were seeing an issue where every pod was getting `UpdatePod` called when restarting the virtual-kubelet due to [podutils.PopulateEnvironmentVariables](https://github.com/virtual-kubelet/virtual-kubelet/blob/d41d02609f8ed3f606da0ddfc36b074554826436/internal/podutils/env.go#L71) getting called before performing the `podEqual` computation.

We've [fixed it internally in our fork](https://github.com/superfly/virtual-kubelet/pull/1) and would be happy to upstream the changes.

PR #829, opened 2020-04-09
Create functions to resolve env vars, allow users to override default resolution

justnoise:
Here's a shot at reworking resolving environment variables with the hope of allowing providers to customize how environment variables are resolved.

The original idea was spelled out in this draft: https://github.com/virtual-kubelet/virtual-kubelet/pull/827

Since this changes the order in which environment variables are resolved, I pulled in k8s kubelet tests (kube_env_test.go) to be confident our results match the kubelet's implementation.

The upside of this change is that it creates a library of components for resolving environment variables and allows providers greater control.  A real world example is that our provider can now populate `status.PodIP` fieldRefs.  The downside is that this implementation is less performant than the existing implementation from the kubelet since it iterates through a container's Env multiple times.

Alternative implementations/changes:
1. I'm curious if passing in envResolver as a function to the pod controller is too restrictive? Would an interface allow the user an easier path to customize env var resolution? As it is, the user can still create a closure to access variables that aren't present in `env.ResolverConfig`
2. Should a new directory (`lib/` or `pkg/`) be created to hold library building blocks that the runtime is constructed out of?


cpuguy83:
WDYT about if we put this into a `podutils` package in the root of the repo?
Probably would be nice to make the existing error constants in the node package as aliases to the new package.

cwdsuzhou:
I would like this PR to be merged. I have met the same issue

cwdsuzhou:
@justnoise wdyt about @cpuguy83 's suggestions.

justnoise:
> @justnoise wdyt about @cpuguy83 's suggestions.

Thanks for letting me know you're also interested in this change!  I think they're good suggestions and I'll try to find time to move  things around this week.

cwdsuzhou:
> > @justnoise wdyt about @cpuguy83 's suggestions.
> 
> Thanks for letting me know you're also interested in this change! I think they're good suggestions and I'll try to find time to move things around this week.

Yeap, actually, we would not like to conevert field from configmap or secret to env.  So we prefer we could define the `func`

node/podcontroller.go
```diff
@@ -158,6 +163,8 @@ type PodControllerConfig struct {
 	ConfigMapInformer corev1informers.ConfigMapInformer
 	SecretInformer    corev1informers.SecretInformer
 	ServiceInformer   corev1informers.ServiceInformer
+
+	EnvResolver podutils.ResolverFunc
```
cpuguy83, line 167: Since we are already resolving envs, I think we don't need to make it configurable, and instead make the non-pod informers optional... e.g. if a configmap informer is not provided, then we don't resolve config maps.... or maybe it's all or nothing... provide all these informers and we'll resolve env vars, otherwise we won't.

podutils/env.go
```diff
@@ -0,0 +1,598 @@
...
+			if len(invalidKeys) > 0 {
+				sort.Strings(invalidKeys)
+				recorder.Eventf(pod, corev1.EventTypeWarning, ReasonInvalidEnvironmentVariableNames, "Keys [%s] from the EnvFrom configMap %s/%s were skipped since they are considered invalid environment variable names.", strings.Join(invalidKeys, ", "), pod.Namespace, name)
+			}
+		}
+	}
+
+	for i := range container.Env {
+		if container.Env[i].ValueFrom != nil && container.Env[i].ValueFrom.ConfigMapKeyRef != nil {
+			envVar := container.Env[i]
```
cwdsuzhou, line 204: I recommend using a code-style like this:
```
if container.Env[i].ValueFrom != nil {
    continue
}

if container.Env[i].ValueFrom.ConfigMapKeyRef != nil {
    continue
}
```

node/podcontroller.go
cwdsuzhou: > Since we are already resolving envs, I think we don't need to make it configurable, and instead make the non-pod informers optional... e.g. if a configmap informer is not provided, then we don't resolve config maps.... or maybe it's all or nothing... provide all these informers and we'll resolve env vars, otherwise we won't.

Agree with @cpuguy83 , actually we just want to the same behavior with `kubelet`. Since the current solution can satisfy the requirement, I think that is enough.


@justnoise WDYT?

cpuguy83:
Ping, is this still being worked on?

cwdsuzhou:
I  am also waiting for this PR

justnoise:
Yes. I just returned from being out of town/vacation this week.  I'm working through other work priorities but should be able to get to this today.

podutils/env.go
justnoise: I like it since this gets rid of a level of indentation. However I prefer the coupling of the nested checks into a single statement using a short circuit `||`. That shows that the checks for `nil` are related, reducing a bit of the cognitive overhead as well as reducing line count:

```go
if container.Env[i].ValueFrom == nil ||
    container.Env[i].ValueFrom.ConfigMapKeyRef == nil {
    continue
}
```

cwdsuzhou:
@cpuguy83 could you help take a look at this PR

podutils/env.go
```diff
@@ -0,0 +1,610 @@
...
+	// ReasonFailedToReadMandatorySecret is the reason used in events emitted when a mandatory secret could not be read.
+	ReasonFailedToReadMandatorySecret = "FailedToReadMandatorySecret"
+
+	// ReasonInvalidEnvironmentVariableNames is the reason used in events emitted when a configmap/secret referenced in a ".spec.containers[*].envFrom" field contains invalid environment variable names.
+	ReasonInvalidEnvironmentVariableNames = "InvalidEnvironmentVariableNames"
+)
+
+var masterServices = sets.NewString("kubernetes")
+
+type ResolverConfig struct {
```
cpuguy83, line 70: Can you add a doc string for this?

podutils/env.go
```diff
@@ -0,0 +1,610 @@
...
+	ConfigMapLister corev1listers.ConfigMapLister
+	SecretLister    corev1listers.SecretLister
+	ServiceLister   corev1listers.ServiceLister
+}
+
+func (rc ResolverConfig) CanResolveEnvVars() bool {
```
cpuguy83, line 76: Needs doc string

node/env.go
```diff
@@ -1,476 +0,0 @@
...
-	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
-	fieldpath "k8s.io/kubernetes/pkg/fieldpath"
-	"k8s.io/kubernetes/pkg/kubelet/envvars"
-	"k8s.io/kubernetes/third_party/forked/golang/expansion"
-
-	"github.com/virtual-kubelet/virtual-kubelet/internal/manager"
-	"github.com/virtual-kubelet/virtual-kubelet/log"
-)
-
-const (
```
cpuguy83, line 39: Unfortunately we need to keep these constants for API compat.
They can just be aliases to the new package, though.

justnoise: Gotcha. You said this before but I misunderstood the purpose. Thanks!

podutils/env.go
justnoise: With a new eye on API compatibility and minimizing exported structs and methods, I've decided to do away with the `ResolverConfig`.  I simply moved the fields in `ResolverConfig` into the `PodController`.  I think that's a lot cleaner overall.

podutils/env.go
justnoise: This has been moved into the `PodController`.

cwdsuzhou:
@justnoise are you still work on this?

cpuguy83 (approved):
LGTM

justnoise:
I tested this out with our (Kip) provider and feel pretty good about the changes. At some point in the future, it might make sense to consider implementing https://github.com/virtual-kubelet/virtual-kubelet/issues/649 in order to prevent spurious updates to env vars and save a bit of work.

Looks like @sargun was recently added as a reviewer.  Welcome to the party.  Happy to make any necessary or helpful changes.

node/podcontroller.go
```diff
@@ -194,32 +198,18 @@ func NewPodController(cfg PodControllerConfig) (*PodController, error) {
 	if cfg.PodInformer == nil {
 		return nil, errdefs.InvalidInput("missing pod informer")
 	}
-	if cfg.ConfigMapInformer == nil {
```
sargun, line 197: Any reason to switch these to being optional?

node/pod.go
```diff
@@ -68,9 +69,12 @@ func (pc *PodController) createOrUpdatePod(ctx context.Context, pod *corev1.Pod)
 
 	// We do this so we don't mutate the pod from the informer cache
 	pod = pod.DeepCopy()
-	if err := populateEnvironmentVariables(ctx, pod, pc.resourceManager, pc.recorder); err != nil {
-		span.SetStatus(err)
-		return err
+	if pc.canResolveEnvVars() {
+		err := podutils.PopulateEnvironmentVariables(ctx, pod, pc.recorder, pc.configMapLister, pc.secretLister, pc.serviceLister)
```
sargun, line 73: Maybe this is dumb question, but how do we know that these are consistent / up to date?

node/podcontroller.go
```diff
@@ -228,6 +218,11 @@ func NewPodController(cfg PodControllerConfig) (*PodController, error) {
 		podStatusQ:         workqueue.NewNamedRateLimitingQueue(cfg.RateLimiter, "syncPodStatusFromProvider"),
 		podEventFilterFunc: cfg.PodEventFilterFunc,
 	}
+	if cfg.ConfigMapInformer != nil && cfg.SecretInformer != nil && cfg.ServiceInformer != nil {
+		pc.configMapLister = cfg.ConfigMapInformer.Lister()
```
sargun, line 222: My suggestion is that you make these encapsulated in some structure like listers, and then you do something like:

```go

func (l *listers) initialized() bool {
   return l != nil
}
```

sargun (commented):
Wow. This is a lot of code. Can we do the restructuring of the listers in one PR maybe? My other question, is how is consistency handled? I see that there is proper error handling, in that it returns not found, but should we check for caches being up to date?

node/podcontroller.go
cpuguy83: Gives the provider the option to handle this themselves rather than having it embedded in the runtime.

cpuguy83:
ping @justnoise

justnoise:
Howdy y'all.  I think this PR should be closed because I no longer work at Elotl or on a VK provider.  I was hoping to push this PR through during my last weeks of employment but the features this change enables are not a priority.  I think it's a large enough change that there should be someone engaged in the project who needs the change and can maintain the change.

justnoise:
I should also add a big thank you for taking the time to read through and comment on this PR. It was a big change and the feedback not only helped make the code better but also taught me a few things in the process.

cwdsuzhou:
@sargun @cpuguy83 I would like  open  a new PR to do this? 

WDYT?

PR #830, opened 2020-04-21
dedup in updatePodStatus

cwdsuzhou:
dedup in updatePodStatus as mentioned in https://github.com/virtual-kubelet/virtual-kubelet/pull/824#discussion_r411047984

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -185,6 +185,9 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes 
 	kPod.Lock()
 	podFromProvider := kPod.lastPodStatusReceivedFromProvider.DeepCopy()
 	kPod.Unlock()
+	if cmp.Equal(podFromKubernetes.Status, podFromProvider.Status) {
+		return nil
+	}
 	// We need to do this because the other parts of the pod can be updated elsewhere. Since we're only updating
 	// the pod status, and we should be the sole writers of the pod status, we can blind overwrite it. Therefore
 	// we need to copy the pod and set ResourceVersion to 0.


cwdsuzhou:
@sargun @cpuguy83  PTAL  thanks

node/pod.go
```diff
@@ -185,6 +185,9 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
 	kPod.Lock()
 	podFromProvider := kPod.lastPodStatusReceivedFromProvider.DeepCopy()
 	kPod.Unlock()
+	if cmp.Equal(podFromKubernetes.Status, podFromProvider.Status) {
```
sargun, line 188: Stupid question, with the UpdatePod callback suppression, where would this get picked up?

cwdsuzhou: Sorry, I do nor get it. But `UpdatePod` would only update pod from `api-server` to `provider`.  This commit only change the logic in func `updatePodStatus`.

sargun:
This still technically changes behavior. I’m inclined to put it behind an option, or add further safety measures (compare the pod against the last update sent to api server).

Again, although this doesn’t break the Kubernetes “spec”, it does change behaviour, and can thus result in unexpected results.

cwdsuzhou:
Since, this is a func called updatePodStatus. I think it is reasonable to update status here rather than spec. If we think this is an improvement, it’s better add option. But IMO this is a bugfix. In a large cluster, we found It would  easily reach the rate-limit of client-go.

node/pod.go
sargun: I'm curious, what's the case where the provider would send duplicate updates?

cwdsuzhou:
ping @cpuguy83 @sargun for review

cwdsuzhou:
@sargun  wdyt, please let me know.
In our large cluster, `updatePodStatus` has been the bottleneck. After we applied the commit, we found we it would have better performance.

cwdsuzhou:
@cpuguy83 could you help merge this PR?

cpuguy83:
It seems like @sargun has some unanswered questions.

sargun:
Sorry, I was leaving my comment yesterday. My question is that even with the previous change of deduping at update / notify time, are you still seeing duplicate / unnecessary updates? If so, it sounds like there is buggy behaviour elsewhere. I'm curious as to what that behaviour is

cwdsuzhou:
> Sorry, I was leaving my comment yesterday. My question is that even with the previous change of deduping at update / notify time, are you still seeing duplicate / unnecessary updates? If so, it sounds like there is buggy behaviour elsewhere. I'm curious as to what that behaviour is

Yes, this issue still exists. It would happen when we `restart` virtual-kubelet. @sargun

node/pod.go
cwdsuzhou: restart virtual-kubelet

sargun:
This only happens in the sync pod provider case right? Any reason you don't convert your provider to async? The sync API is deprecated.

cwdsuzhou:
> This only happens in the sync pod provider case right? Any reason you don't convert your provider to async? The sync API is deprecated.



> This only happens in the sync pod provider case right? Any reason you don't convert your provider to async? The sync API is deprecated.

Good idea. However, due to some reason, provider may do not support async pod status. I think it is also reasonable. This PR just add check before update status, would not harm the original logic.

cwdsuzhou:
No matter we use sync or async, we have to call `updatePodStatus`, so if we do not add a check this issue may still exist

sargun:
So, there are three cases:
1. Sync providers
2. Async providers
3. All providers at startup

I'm going to dig in a little bit next week on the async and sync provider cases. I understand the startup case. I think that for the sync case, we can put this logic in syncProviderWrapper. 

In the async case, I'm still not groking how this can happen even on startup, because the AddPod function that called on startup should hit this case:
https://github.com/virtual-kubelet/virtual-kubelet/blob/master/node/pod.go#L76-L88

And the update should only happen at startup if the pod retrieved from the provider was not equal to the pod from apiserver. Even in that case, all it will do is generate an update.

sargun:
I think this would be safer if there was a temp-dirty bit that was set when we do this, and then the next time we receive a pod status update from K8s, we check the dirty bit, and put it back on the sync from provider queue (and clear the dirty bit).

Most of the time it'll just result in a noop.

cwdsuzhou:
IMO, pod spec update should enqueue K8SQ, pod status update should enqueue podStatusQueue. I think we should not concern `spec change`  in func `podStatusUpdate`.
If status did not change, it is also meaningless to update  it to K8S

sargun:
Please rebase.

cwdsuzhou:
> Please rebase.

@sargun done, PTAL

sargun:
@cpuguy83 Do you want to take another look, or good to merge?

sargun:
Let's give it another day or two, in case @cpuguy83 has any feedback. In the mean time, can you rebase?

PR #833, opened 2020-06-03
Fix stream timeout defaults

cpuguy83:
This was an unintentional breaking change in
0bdf7423037e0ccefb794c8b5f570137c65ce864

A timeout of 0 doesn't make any sense, so use the old values of 30s as a
default.

### DIFF
--- a/node/api/exec.go
+++ b/node/api/exec.go
@@ -91,6 +91,14 @@ func HandleContainerExec(h ContainerExecHandlerFunc, opts ...ContainerExecHandle
 	for _, o := range opts {
 		o(&cfg)
 	}
+
+	if cfg.StreamIdleTimeout == 0 {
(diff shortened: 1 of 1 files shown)


cpuguy83:
@cwdsuzhou

cwdsuzhou:
LGTM

ldx:
@cpuguy83 the defaults in the kubelet are 4 hours for the idle timeout and 30 seconds for the create timeout:

    DefaultStreamIdleTimeout     = 4 * time.Hour
    DefaultStreamCreationTimeout = 30 * time.Second

Right now VK closes exec connections after 30s of inactivity.

cpuguy83:
That's fine, we can change the defaults. Just that the default inadvertently got changed to 0.

PR #835, opened 2020-06-06
Avoid enqueue when status of k8s pods change

cwdsuzhou:


cwdsuzhou:
@cpuguy83 PTAL

We should ignore podStatus change, it not. too many useless enqueue here.

node/podcontroller.go
```diff
@@ -270,13 +270,19 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
...
 			newPod := newObj.(*corev1.Pod)
 
 			// At this point we know that something in .metadata or .spec has changed, so we must proceed to sync the pod.
 			if key, err := cache.MetaNamespaceKeyFunc(newPod); err != nil {
 				log.G(ctx).Error(err)
 			} else {
-				pc.k8sQ.AddRateLimited(key)
+				// ignore ResourceVersion change
+				oldPod.ResourceVersion = newPod.ResourceVersion
+				if !cmp.Equal(oldPod.Spec, newPod.Spec) ||
```
sargun, line 282: If you're trying to achieve update suppression, you can use: https://github.com/virtual-kubelet/virtual-kubelet/blob/a67cfab42bc77a441636801d906efceba0683da7/node/pod.go#L103

Also, you need to check deletion timestamp and phase change IIRC

cwdsuzhou: No, wo should  Ignore status update here. But we need check `spec`, `deleteTime`, `Annotation`, `Labels` and so on. So I  exclude pod status comparison

node/podcontroller.go
```diff
@@ -270,13 +270,19 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
...
+			oldPod := oldObj.(*corev1.Pod)
 			newPod := newObj.(*corev1.Pod)
 
 			// At this point we know that something in .metadata or .spec has changed, so we must proceed to sync the pod.
 			if key, err := cache.MetaNamespaceKeyFunc(newPod); err != nil {
 				log.G(ctx).Error(err)
 			} else {
-				pc.k8sQ.AddRateLimited(key)
+				// ignore ResourceVersion change
+				oldPod.ResourceVersion = newPod.ResourceVersion
```
sargun, line 281: aren't these cache references? So you're mutating the cache here?

node/podcontroller.go
sargun: Please add a more targeted equal method that tests the super set of the attributes of podequal. Something like:
```
podNeedsWork(old, new *v1.pod) bool {
 if !podsEqual(...) { return true };
 if ... return true
 return false
}
```

I'm curious, where is the perf concern?

sargun:
Somehow my review got split. None the less, I think this is a fine idea.

node/podcontroller.go
cwdsuzhou: Done, thanks, PTAL thanks

cwdsuzhou:
@sargun PR updated, thanks

node/pod.go
```diff
@@ -119,6 +119,17 @@ func podsEqual(pod1, pod2 *corev1.Pod) bool {
 
 }
 
+// podShouldEnqueue checks if two pods equal according according to podsEqual func and DeleteTimeStamp
+func podShouldEnqueue(oldPod, newPod *corev1.Pod) bool {
+	if !podsEqual(oldPod, newPod) {
+		return true
+	}
+	if !oldPod.DeletionTimestamp.Equal(newPod.DeletionTimestamp) {
```
sargun, line 127: In the earlier patch, you checked all of object meta. Any reason for this change?

cwdsuzhou: I think Labels, Annotations, DeleteTimeStamp update should enqueue, the earlier patch check object meta for easier comparison. But I found podsEqual has check labels and annotations , so I thin this would be easier and effective.

sargun: I agree. Why didn't you add labels and annotations here? Honestly, we should probably make the update pod callback on label and annotation change. -- that's work for a different time.

sargun:
One last question, and then we're good to go.

node/pod.go
cwdsuzhou: @sargun `podsEqual` has already checked  labels and annotations.

sargun: We should also check to make sure the deletion graceperiod is equal. and then I think it'll be good to go.

cwdsuzhou: See https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L220

Actually, it would only set when DeletionTimestamp is set, and if it changes, DeletionTimestamp must have been changed. So I think it is not necessary!

sargun: I think it can change, let's take this example:

Timestamp | Actor | Action
00:00:00    | User  | Delete pod A, with Grace period of 30s
00:00:00   | K8s.   | Sets Deletion timestamp to 00:00:30, and graceperiod to 30s
00:00:15    | User  | Delete pod A, with Grace period of 15s
00:00:15   | K8s.   | Sets Deletion timestamp to 00:00:30, and graceperiod to 15s

In this, the Deletion Timestamp can change without the graceperiod changing.

cwdsuzhou: Yes, this is a case, but it would not affect the final delete time.

cwdsuzhou: If you think it is necessary,  I would like update this PR

sargun: It doesn't hurt to add this check, IMHO.

cwdsuzhou: PR updated, PTAL

thanks

shsjshentao:
I think it should not ignore podStatus change when enqueue, if i set graceful deletion and delete pod, then all containers has exicted in advance, it should kill now instead of waiting graceful deletion.

cwdsuzhou:
> I think it should not ignore podStatus change when enqueue, if i set graceful deletion and delete pod, then all containers has dead in advance, it should kill immediately instead of waiting graceful deletion.

You can take a look at these 2 PRs , may solve your issues.

https://github.com/virtual-kubelet/virtual-kubelet/pull/902
https://github.com/virtual-kubelet/virtual-kubelet/pull/874

shsjshentao:
> > I think it should not ignore podStatus change when enqueue, if i set graceful deletion and delete pod, then all containers has dead in advance, it should kill immediately instead of waiting graceful deletion.
> 
> You can take a look at these 2 PRs , may solve your issues.
> 
> #902 #874

No, these 2 PRs cannot solve this problem.

PR #838, opened 2020-06-11
Introduce three-way patch for proper handling of updates

sargun:
As described in the patch itself, there is a case that if a node is updated out of
band (e.g. node-problem-detector (https://github.com/kubernetes/node-problem-detector)),
we will overwrite the patch in our typicaly two-way strategic patch for node status
updates.

The reason why the standard kubelet can do this is because the flow goes:
apiserver->kubelet: Fetch current node
kubelet->kubelet: Update apiserver's snapshot with local state changes
kubelet->apiserver: patch

We don't have this luxury, as we rely on providers making a callback into us
in order to get the most recent pod status. They do not have a way
to do that merge operation themselves, and a two-way merge doesn't
give us enough metadata.

In order to work around this, we perform a three-way merge on behalf of
the user. We do this by stashing the contents of the last update inside
of it. We then fetch that status back, and use that for the future
update itself.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -17,7 +17,6 @@ package node
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"time"
 
 	pkgerrors "github.com/pkg/errors"
@@ -31,6 +30,14 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/kubernetes/typed/coordination/v1beta1"
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/retry"
+)
+
+const (
+	// Annotation with the JSON-serialized last applied node conditions. Based on kube ctl apply. Used to calculate
+	// the three-way patch
+	virtualKubeletLastNodeAppliedNodeStatus = "virtual-kubelet.io/last-applied-node-status"
+	virtualKubeletLastNodeAppliedObjectMeta = "virtual-kubelet.io/last-applied-object-meta"
 )
 
 // NodeProvider is the interface used for registering a node and updating its
@@ -287,6 +294,13 @@ func (n *NodeController) controlLoop(ctx context.Context) error {
 			}
 
 			n.n.Status = updated.Status
+			n.n.ObjectMeta = metav1.ObjectMeta{
+				Annotations: updated.Annotations,
+				Labels:      updated.Labels,
+				Name:        n.n.ObjectMeta.Name,
+				Namespace:   n.n.ObjectMeta.Namespace,
+				UID:         n.n.ObjectMeta.UID,
+			}
 			if err := n.updateStatus(ctx, false); err != nil {
 				log.G(ctx).WithError(err).Error("Error handling node status update")
 			}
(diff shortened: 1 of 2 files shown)


sargun:
This is a pretty messy change. Feedback appreciated.

cpuguy83:
It seems like it'd be simpler to have the node provider fetch the node status?

sargun:
@cpuguy83 That would can't be done without an API-breaking change. Specifically because of what we do around 0-out resource version. The read / modify / write has to be done in a "txn" (preservation of resource version).

I'm fine with an API-breaking change, but how do you see that looking?

Also, how do we balance the weight of requiring all providers to implement this, versus doing it ourselves?

sargun:
I have some ideas around "new" APIs to do this. Is it worth investigating them further, or hopping on a VC to chat about this?

cpuguy83:
> That would can't be done without an API-breaking change

I'm not sure I follow here.
I'm saying that the provider can, if they want to not have the existing behavior, fetch the latest version itself, then apply changes on top, then notify vk of the change.

sargun:
Not with the current code. We squash the node because we set resource version to 0, so we can overwrite any node changes that happened in the interim period.

cpuguy83:
Right. So it comes down to a potential for a race in that case.

sargun:
In practice, this happens enough where it breaks stuff. :(

sargun:
@cpuguy83 *nudge*

sargun:
@cpuguy83 I've simplified the code and the upgrade scenario a little bit.

sargun:
This fixes: https://github.com/virtual-kubelet/virtual-kubelet/issues/807

node/node.go
```diff
@@ -406,42 +420,95 @@ func updateNodeLease(ctx context.Context, leases v1beta1.LeaseInterface, lease *
...
+		   |                                                 | --------------------------------\  |
+		   |                                                 |-| Node conditions are [A, B, C] |  |
+		   |                                                 | |-------------------------------|  |
+		   | ------------------------------------------\     |                                    |
+		   |-| Updates internal node conditions to [A] |     |                                    |
+		   | |-----------------------------------------|     |                                    |
+		   |                                                 |                                    |
+		   | Patch delete: B, upsert A                       |                                    |
+		   |------------------------------------------------>|                                    |
+		   |                                                 |                                    |
```
pires, line 451: I think a few lines are missing here clearly showing the problem is that node conditions are `[A]` instead of `[A, C]`, right?

sargun:
@pires Updated the doc comment / cleanup the commit a little bit, and addressed how updates / initial transition / pre-created nodes should work. PTAL.

node/node.go
```diff
@@ -406,42 +420,117 @@ func updateNodeLease(ctx context.Context, leases v1beta1.LeaseInterface, lease *
...
+		return nil, pkgerrors.Wrap(err, "Cannot marshal node status from provider")
 	}
-
-	// Reset spec to make sure only patch for Status or ObjectMeta is generated.
-	// Note that we don't reset ObjectMeta here, because:
-	// 1. This aligns with Nodes().UpdateStatus().
-	// 2. Some component does use this to update node annotations.
-	newNode.Spec = oldNode.Spec
-	newData, err := json.Marshal(newNode)
+	newNode.Annotations[virtualKubeletLastNodeAppliedNodeStatus] = string(virtualKubeletLastNodeAppliedNodeStatusBytes)
```
cpuguy83, line 514: Should these be base64 encoded?

sargun: Nope. JSON is fine to put here. In fact, this is exactly how kubectl does it. Base64 is just adding overhead and a readability headache.

cpuguy83:
@pires LGTY?

sargun:
I think @pires only issue was with the comments. @cpuguy83 Are the comments clear to you?

sargun:
thanks!

Issue #836, opened 2020-06-06
NotifyPods may not work when vk starts up


cwdsuzhou:
---

When vk starts up, updatw pod status may failed. There is a race between we get pod from KnowPods and and to Known pods.

https://github.com/virtual-kubelet/virtual-kubelet/blob/master/node/podcontroller.go#L246
notifyPods add pods the statusQueue

https://github.com/virtual-kubelet/virtual-kubelet/blob/master/node/podcontroller.go#L267
here we add to KnownPods

https://github.com/virtual-kubelet/virtual-kubelet/blob/master/node/pod.go#L213
when pods dequeue, it may does not exist in KnownPods

I would like send a PR to fix it

### Environment summary

Provider (e.g. ACI, AWS Fargate)

Version (e.g. 0.1, 0.2-beta)

K8s Master Info (e.g. AKS, ACS, Bare Metal, EKS)

Install Method (e.g. Helm Chart, )

sargun:
Looks right. Thanks for spotting this. This should just be a matter of moving:

```
	// Wait for the caches to be synced *before* starting to do work.
	if ok := cache.WaitForCacheSync(ctx.Done(), pc.podsInformer.Informer().HasSynced); !ok {
		return pkgerrors.New("failed to wait for caches to sync")
	}
	log.G(ctx).Info("Pod cache in-sync")
```

above 

```
provider.NotifyPods(ctx, func(pod *corev1.Pod) {
		pc.enqueuePodStatusUpdate(ctx, podStatusQueue, pod.DeepCopy())
	})
```

cwdsuzhou:
> Looks right. Thanks for spotting this. This should just be a matter of moving:
> 
> ```
> 	// Wait for the caches to be synced *before* starting to do work.
> 	if ok := cache.WaitForCacheSync(ctx.Done(), pc.podsInformer.Informer().HasSynced); !ok {
> 		return pkgerrors.New("failed to wait for caches to sync")
> 	}
> 	log.G(ctx).Info("Pod cache in-sync")
> ```
> 
> above
> 
> ```
> provider.NotifyPods(ctx, func(pod *corev1.Pod) {
> 		pc.enqueuePodStatusUpdate(ctx, podStatusQueue, pod.DeepCopy())
> 	})
> ```

Yes, I think so

sargun:
Let me know when you submit a PR.

PR #839, opened 2020-06-12
Bugfix: cache may be not ready when NotifyPods is called

cwdsuzhou:
Fix https://github.com/virtual-kubelet/virtual-kubelet/issues/836


cwdsuzhou:
@sargun  PTAL thanks

sargun:
Small concern I have: This may break existing providers as they will get events that can result in pod status updates before the notify is set. How do we deal with that for async providers?

sargun:
Yeah, this blows up the provider we use in production.

sargun:
The easy hack seems to be....add another queue. the right solution seems to be that we should make the API change that the provider needs to delay / queue the changes on their side.

sargun:
@cwdsuzhou Any suggestions, RE: the comment above?

cpuguy83:
Not sure if I exactly understand the problem described, but would it make sense to check if pods are in sync before returning early from the knownPods check?

sargun:
@cpuguy83 I think that'd work.

cwdsuzhou:
> Small concern I have: This may break existing providers as they will get events that can result in pod status updates before the notify is set. How do we deal with that for async providers?

I think this case may exist, do you have a better solution.

sargun:
Personally, I would really like a traditional queue in front of enqueuePodStatusUpdate. We can then in enqueuePodStatusUpdate check if the pod is not in known pods, and then check if it's in the pod lister. If it is not in known pods, and in the pod lister, then defer, and try again later.

@cpuguy83 opposed?

Also, I believe using cache..waitinsync... isn't sufficient here, because waitinsync doesn't actually check if all of the callbacks have finished being processed IIRC.

cpuguy83:
I'd prefer a simple as possible.
Can we get around this by re-queuing?

sargun:
@cpuguy83 With a traditional queue, yes. It'd require direct usage of the workqueue object, as opposed to our wrapping.

We should also check the podlister in enqueue pod status, and if the pod is not in known pods, but in pod lister, then it may be eligible for requeuing.

cwdsuzhou:
I think we can add flag to mark if cache has synced.
If we can not get it from  the known pod and the flag is false, re-enqueue it, else keep the current logic.

WDYT? @sargun @cpuguy83

sargun:
1. Caches in sync doesn't actually mean knownpods has finished populating. The callbacks are executed asynchronously to updating the cache.  We can fix this problem by checking the podlister AND the knownpods map
2. re-enqueue it where?

cwdsuzhou:
> 1. Caches in sync doesn't actually mean knownpods has finished populating. The callbacks are executed asynchronously to updating the cache.  We can fix this problem by checking the podlister AND the knownpods map
> 2. re-enqueue it where?

1. Yes, check pod lister may be a good idea.
2. enqueuePodStatusUpdate

sargun:
Do you want to try your hand at the patch?

cwdsuzhou:
> Do you want to try your hand at the patch?

Sure, but I would like know what do @cpuguy83 think before I update the PR

cpuguy83:
It seems fine. Just to be clear, would this be adding a new queue?

cwdsuzhou:
> It seems fine. Just to be clear, would this be adding a new queue?

Ｉprefer not. Just add it back to statusQueue

cwdsuzhou:
@sargun  @cpuguy83  PTAL thanks

node/podcontroller.go
```diff
@@ -256,6 +253,10 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 	}
 	log.G(ctx).Info("Pod cache in-sync")
 
+	provider.NotifyPods(ctx, func(pod *corev1.Pod) {
```
cpuguy83, line 256: Did we need to put this back to the original position?

node/pod.go
```diff
@@ -247,7 +252,17 @@ func (pc *PodController) enqueuePodStatusUpdate(ctx context.Context, q workqueue
...
 			q.AddRateLimited(key)
+			return
+		}
+		// check the pod from lister
+		_, err := pc.podsLister.Pods(pod.Namespace).Get(pod.Name)
+		if err != nil {
+			return
 		}
+		obj := &knownPod{lastPodStatusReceivedFromProvider: pod}
+		pc.knownPods.Store(key, obj)
```
sargun, line 263: This exposes a race if the pod is deleted from API server. If pods.get succeeds, and then the pod is deleted (and we get the deleted callback), we can re-store the pod in known pods, and leak. I think the only "right way" is to add another queue and re-queue.

node/pod.go
```diff
@@ -247,7 +252,17 @@ func (pc *PodController) enqueuePodStatusUpdate(ctx context.Context, q workqueue
 			kpod.lastPodStatusReceivedFromProvider = pod
 			kpod.Unlock()
 			q.AddRateLimited(key)
+			return
+		}
+		// check the pod from lister
+		_, err := pc.podsLister.Pods(pod.Namespace).Get(pod.Name)
+		if err != nil {
```
sargun, line 259: This should be attenuated to not found.

sargun: Err. This reads cache right? So, the only valid error should be not found or return.

cwdsuzhou: If the error is not `Not Found` what we should do?

node/pod.go
cwdsuzhou: Normally, we would receive an update event before we delete it because we need set the deleteTimeStamp.

node/pod.go
sargun: Good question. In the multi-queue model, I suggest at a minimum, we log it.

sargun:
I think the easier approach is just add another queue....

node/pod.go
cwdsuzhou: IMO， just log it

node/pod.go
```diff
@@ -209,6 +210,10 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
 	kPod.Lock()
 	podFromProvider := kPod.lastPodStatusReceivedFromProvider.DeepCopy()
 	kPod.Unlock()
+	if podFromProvider == nil {
+		// if podFromProvider it nil means pod has not synced, we should re-enqueue it
+		return fmt.Errorf("pod %v has not been synced to kown pod", key)
```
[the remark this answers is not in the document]
cwdsuzhou: fixed

node/podcontroller.go
cwdsuzhou: IMO, should not. If we put it back, the issue would still exist.

cwdsuzhou:
> I think the easier approach is just add another queue....

I would not like add another queue, current seems enough, what do you think @cpuguy83

sargun:
I don't see a fix for the pointed out race condition.

cwdsuzhou:
> I don't see a fix for the pointed out race condition.

Sorry, I can not found a perfect solution to the race condition through it has very low possibility.  

But I think we check whether all pods have been added to known pods before when starting the func `NotifyPods`, so we do not need to check lister again.

sargun:
I think the race-free solution is to add another queue that will hold these updates until we're done synching, or to change the contract, and say that notifypods may block at startup while we wait for cache sync.

Upon further thought, I'm not sure we need to wait for all callbacks to complete, as I'm not sure the provider should be sending us updates for pods that do not yet exist from the perspective of VK.

cwdsuzhou:
> > I don't see a fix for the pointed out race condition.
> 
> Sorry, I can not found a perfect solution to the race condition through it has very low possibility.
> 
> But I think we check whether all pods have been added to known pods before when starting the func `NotifyPods`, so we do not need to check lister again.

PR updated according to this @sargun

sargun:
I'm going to try to take another look at this this weekend. Sorry for the wait.

cwdsuzhou:
> I'm going to try to take another look at this this weekend. Sorry for the wait.

Please let me know if you have some suggestions!

cwdsuzhou:
@cpuguy83 @sargun  PTAL, I think  race has been solved

node/podcontroller.go
```diff
@@ -263,6 +262,21 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 	}
 	log.G(ctx).Info("Pod cache in-sync")
 
+	pods, err := pc.podsLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list pod error: %v", err)
+	}
+	var podCount int32
+	started := false
```
sargun, line 270: At a minimum this will trigger a race condition.

cwdsuzhou: Yes, but actually would not influent the  final results.

I  have  added another commit based on this one. But I still think that would make the `add func` heavier.

WDYT? @sargun @cpuguy83

sargun:
I don't mean to be a debbie-downer (yet again), but I think that there is yet another race condition this exposes where if the pod lister lists N pods, and then one of the pods is terminated, started wont be set to one until *yet another* pod is added to the node.

cpuguy83:
Closing in favor of https://github.com/virtual-kubelet/virtual-kubelet/pull/851

Thanks!

cwdsuzhou:
Thanks @cpuguy83 @sargun

Issue #840, opened 2020-06-18
DefaultControllerRateLimiter may cause slowly update


cwdsuzhou:
---

### Environment summary

Provider (e.g. ACI, AWS Fargate)

Version (e.g. 0.1, 0.2-beta)

K8s Master Info (e.g. AKS, ACS, Bare Metal, EKS)

Install Method (e.g. Helm Chart, )

### Issue Details

DefaultControllerRateLimiter may cause slowly update.

When we delete large number of pods, we found some pods cannot be deleted in time. We would like send a PR to expose the qps limit of the default rate limiter.

PR #834, opened 2020-06-04
Upgrade Kubernetes libraries to 1.17

sargun:
This updates the Kubernetes libraries to 1.17.6. I also added a doc that explains what the process is for updating these deps in the future.

PR #841, opened 2020-06-21
Add parameters to support change work queue

cwdsuzhou:
Fix virtual-kubelet/virtual-kubelet#840

### DIFF
--- a/node/podcontroller.go
+++ b/node/podcontroller.go
@@ -27,7 +27,6 @@ import (
 	"github.com/virtual-kubelet/virtual-kubelet/internal/manager"
 	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"github.com/virtual-kubelet/virtual-kubelet/trace"
-	"golang.org/x/time/rate"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	corev1informers "k8s.io/client-go/informers/core/v1"
@@ -109,6 +108,8 @@ type PodController struct {
 	// deletionQ is a queue on which pods are reconciled, and we check if pods are in API server after grace period
 	deletionQ workqueue.RateLimitingInterface
 
+	podStatusQ workqueue.RateLimitingInterface
+
 	// From the time of creation, to termination the knownPods map will contain the pods key
 	// (derived from Kubernetes' cache library) -> a *knownPod struct.
 	knownPods sync.Map
@@ -128,8 +129,6 @@ type PodController struct {
 	// This is used since `pc.Run()` is typically called in a goroutine and managing
 	// this can be non-trivial for callers.
 	err error
-	// qps is the default qps limit when retry
-	qps int
 }
 
 type knownPod struct {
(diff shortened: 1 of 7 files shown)


cwdsuzhou:
tests seem unhappy since https://github.com/virtual-kubelet/virtual-kubelet/pull/834 merged

node/podcontroller.go
```diff
@@ -198,8 +217,9 @@ func NewPodController(cfg PodControllerConfig) (*PodController, error) {
 		ready:           make(chan struct{}),
 		done:            make(chan struct{}),
 		recorder:        cfg.EventRecorder,
-		k8sQ:            workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "syncPodsFromKubernetes"),
-		deletionQ:       workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "deletePodsFromKubernetes"),
+		k8sQ:            workqueue.NewNamedRateLimitingQueue(controllerRateLimiter(cfg.WorkQueueRetryQPS), "syncPodsFromKubernetes"),
```
sargun, line 220: Were the previous defaults 10?

sargun:
tests should be fixed by https://github.com/virtual-kubelet/virtual-kubelet/pull/842 -- the problem is that K8s just release 1.17.7. This broke the validation script.

node/podcontroller.go
cwdsuzhou: Yeap, it is defined 10 in the original client-go

sargun:
btw, what do you recommend?

node/podcontroller.go
cwdsuzhou: > btw, what do you recommend?

In our environment, we have changed it to 100 now to accelerate deleting pods.

cwdsuzhou:
Before this works, we should wait for the PR https://github.com/virtual-kubelet/node-cli/pull/22 in node-cli to be merged

sargun:
@cwdsuzhou I'll hold off on the merge. But, this to me looks like it should work, and fallback to 10.

cpuguy83:
Currently we do not expose anything about the queue, or even that we use a queue to the package API. This changes that.

cpuguy83:
I wonder if it would be better to expose the queue to the config instead of the queue config.
The downside here is we are stuck with the `workqueue.RatelimitingInterface` in that case... not that it's particularly bad, just need to be careful about what we expose.

cwdsuzhou:
> @cwdsuzhou I'll hold off on the merge. But, this to me looks like it should work, and fallback to 10.

yes,  this should work

cwdsuzhou:
> I wonder if it would be better to expose the queue to the config instead of the queue config.
> The downside here is we are stuck with the `workqueue.RatelimitingInterface` in that case... not that it's particularly bad, just need to be careful about what we expose.

This is better.

cwdsuzhou:
@cpuguy83  @sargun 

PR updated, PTAL thanks

cwdsuzhou:
> @cpuguy83 @sargun
> 
> PR updated, PTAL thanks

I will close https://github.com/virtual-kubelet/node-cli/pull/22 and re-open a new one when  this PR has been merged.

sargun:
@cpuguy83 Are you okay with this, or do you want to take a different approach of queue / queue configuration / queue configuration policy?

PR #842, opened 2020-06-23
Fix kubernetes version dependency checking script

sargun:
We had locked to version v1.17.6 when this script was released. At the time,
it was also the current stable release, and what was presented by the
github API. It turns out the Github API does not present all tags. This
changes it to fetch the annotated tag from the upstream repo.


sargun:
@cwdsuzhou Let me know if you have feedback.

cwdsuzhou:
LGTM

thanks

PR #849, opened 2020-07-07
Add serializing queue behind enqueue pods

sargun:
This adds a serializing queue behind enqueue pods, and does
deduping to prevent the race condition where a provider
notifies us as to a pod's existence prior to our cache being warm.

Benchmarking :(:
Before:
```
BenchmarkCreatePods-16    	   10000	    565819 ns/op
```

After:
```
BenchmarkCreatePods-16    	   10000	    742474 ns/op
```

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -233,22 +233,30 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes 
 
 // enqueuePodStatusUpdate updates our pod status map, and marks the pod as dirty in the workqueue. The pod must be DeepCopy'd
 // prior to enqueuePodStatusUpdate.
-func (pc *PodController) enqueuePodStatusUpdate(ctx context.Context, q workqueue.RateLimitingInterface, pod *corev1.Pod) {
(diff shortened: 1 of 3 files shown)


sargun:
@cwdsuzhou This is a solution to your https://github.com/virtual-kubelet/virtual-kubelet/pull/839 PTAL.

cwdsuzhou:
> @cwdsuzhou This is a solution to your #839 PTAL.

I have updated the PR 1 week ago,  I think  race condition does not  exist now.

sargun:
@cpuguy83 Do you have a preference between the two approaches?

cpuguy83:
This one seems heavy for a problem that only seems to occur during startup.

re: https://github.com/virtual-kubelet/virtual-kubelet/pull/839#issuecomment-651250511
>  or to change the contract, and say that notifypods may block at startup while we wait for cache sync.

We do not have any such contract.
The interface says that `NotifyPods` must not block (actually it says "will not", but this should be "must not").
Perhaps the intention there isn't clear, though.
I always intended this to be such that the behavior of the passed in callback is undefined.
Really it could already block since the queue is rate limited.

PR #851, opened 2020-07-14
Solve the notification on startup pod status notification race condition

sargun:
This solves the race condition as described in
https://github.com/virtual-kubelet/virtual-kubelet/issues/836.

It does this by checking two conditions when the possible race condition
is detected.

If we receive a pod notification from the provider, and it is not
in our known pods list:
1. Is our cache in-sync?
2. Is it known to our pod lister?

The first case can happen because of the order we start the
provider and sync our caches. The second case can happen because
even if the cache returns synced, it does not mean all of the call
backs on the informer have quiesced.

This slightly changes the behaviour of notifyPods to that it
can block (especially at startup). We can solve this later
by using something like a fair (ticket?) lock.

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -16,6 +16,8 @@ package node
 
 import (
 	"context"
+	"math/rand"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	pkgerrors "github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


sargun:
Some people may find for loops preferable to `goto`. I can refactor that.

node/podcontroller.go
```diff
@@ -83,7 +83,8 @@ type PodNotifier interface {
 	// fashion. The provided pod's PodStatus should be up to date when
 	// this function is called.
 	//
-	// NotifyPods will not block callers.
+	// NotifyPods may block callers. The NotifyPods function may be called
```
cpuguy83, line 86:
```
NotifyPods must not block the caller since it should only register a callback. The callback passed into `NotifyPods` may block when called.
```

node/pod.go
```diff
@@ -234,21 +236,47 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
...
+		if errors.IsNotFound(err) {
+			return
+		}
+		if err != nil {
+			l.WithError(err).Error("Received error from pod lister while trying to see if pod exists")
+			return
 		}
+		// err is nil, and therefore the pod exists in k8s, but does not exist in our known pods map. Sleep
+		// and retry for somewhere between 1 and 1000 microseconds.
+		time.Sleep(time.Microsecond * time.Duration(rand.Intn(1000)))
```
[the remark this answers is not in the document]
cpuguy83: Perhaps a debug message here?

node/pod.go
```diff
@@ -234,21 +236,47 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
...
+	var ok bool
+retry:
+	obj, ok = pc.knownPods.Load(key)
+	if !ok {
+		// Blind wait for sync. If we haven't synced yet, that's okay too.
+		cache.WaitForCacheSync(ctx.Done(), pc.podsInformer.Informer().HasSynced)
+
+		_, err = pc.podsLister.Pods(pod.Namespace).Get(pod.Name)
+		if errors.IsNotFound(err) {
+			return
```
cpuguy83, line 259:
```suggestion
			l.WithField("pod", key).WithError(err).Debug("Pod not found in pod lister, not enqueuing pod status update")
			return
```

node/pod.go
```diff
@@ -234,21 +236,47 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
...
+		}
+		if err != nil {
+			l.WithError(err).Error("Received error from pod lister while trying to see if pod exists")
```
cpuguy83, line 262: Could there be some transient error that we should try again for?

sargun:
@cpuguy83 Pushed another with fixes to the suggestions. Also, used the Kubernetes waiting library instead of manually implementing the loop. I think a 151 millisecond wait is plenty. If you think it's good, I'll squash.

node/pod.go
```diff
@@ -234,21 +237,66 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
...
+	// 151 milliseconds is arbitrarily chosen here as a small prime number.
+	err = wait.PollImmediateUntil(151*time.Millisecond, func() (bool, error) {
+		var ok bool
+		obj, ok = pc.knownPods.Load(key)
+		if ok {
+			return true, nil
 		}
+
+		// Blind sync. Ignore return code.
+		cache.WaitForCacheSync(ctx.Done(), pc.podsInformer.Informer().HasSynced)
```
cwdsuzhou, line 267: Would this cost too much. In  addition, we  have synced Pod in `Run()`?

cwdsuzhou: That is to say every time we enqueue a pod, we have to run this ?

sargun: No, only if we didn't find it in known pods. If it was in known pods, it immediately returns.

cwdsuzhou: When the vk  starts up,  how can you avoid race since we have  called `cache.WaitForCacheSync` in `Run()`

sargun: It's not a "race" then -- if the pod isn't found, then this will block (and retry).

node/pod.go
```diff
@@ -234,21 +237,66 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
...
+		// err is nil, and therefore the pod exists in k8s, but does not exist in our known pods map. This likely means
+		// that we're in some kind of startup synchronization issue where the provider knows about a pod (as it performs
+		// recover, that we do not yet know about).
+		return false, nil
+	}, ctx.Done())
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			err = fmt.Errorf("Pod %q not found in pod lister: %w", key, err)
+			log.G(ctx).WithError(err).Warn("Not enqueuing pod status update")
```
cpuguy83, line 283: This seems like a perfectly acceptable scenario, where the pod has been deleted from the API server but either haven't been notified or the provider hasn't been notified.

I feel like `Debug` is good here rather than `Warn`

node/pod.go
```diff
@@ -234,21 +237,66 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
...
+		} else {
+			log.G(ctx).WithError(err).Warn("Not enqueuing pod status update due to error from pod lister")
```
cpuguy83, line 285: This feels like we should retry.
What other kinds of error could we receive here?

Since we don't have a means of telling the provider that there was an error here, it seems dangerous to not retry.

sargun: How many times do we retry, how long do we wait -- how many times do we retry?

sargun: Once we fix this up, I'll go ahead and update the patch. I think coming up with a good retry policy is hard. This could only really happen at startup, and if knownPods is not up to date. It wont make any existing user's experience worse.

cpuguy83: There should be a max retry constant already defined. I think it'd make sense to use that at least.

sargun:
@cpuguy83 Upon digging through podlister code, it doesn't look like there are any transient errors that apply to us. The non-transient errors are around things like bad key name for example. Given that, I don't think it makes sense to add retries.

I did cap the retry / poll loop with a timeout though.

cpuguy83:
Oh your right, we are using a local cache.
That makes sense.

cpuguy83:
I'm on mobile currently.
I'll give this another review... probably tomorrow... and merge unless anyone has an objection.

sargun:
No worries. I think this was less of a code patch and more of a code spelunking and commenting exercise.

node/pod.go
cpuguy83: Just noting for posterity on this thread:
We use a local informer cache, so there is no network call here and not likely to be transient errors.

PR #853, opened 2020-07-16
Use /stats/summary for metrics handler

ldx:
To allow users to attach both the pod and metrics handlers to the same mux, use "/stats/summary" as the pattern in AttachPodMetricsRoutes. Otherwise, if mux.Handle() is called again for the same pattern, it will panic, and this is what happens in node-cli due to https://github.com/virtual-kubelet/node-cli/pull/27 calling both AttachPodRoutes() AttachPodMetricsRoutes() with the same mux.

### DIFF
--- a/node/api/server.go
+++ b/node/api/server.go
@@ -40,6 +40,7 @@ type PodHandlerConfig struct {
 	GetPods PodListerFunc
 	// GetPodsFromKubernetes is meant to enumerate the pods that the node is meant to be running
 	GetPodsFromKubernetes PodListerFunc
+	GetStatsSummary       PodStatsSummaryHandlerFunc
 	StreamIdleTimeout     time.Duration
 	StreamCreationTimeout time.Duration
 }
@@ -64,6 +65,13 @@ func PodHandler(p PodHandlerConfig, debug bool) http.Handler {
 			WithExecStreamIdleTimeout(p.StreamIdleTimeout),
 		),
 	).Methods("POST", "GET")
+
+	if p.GetStatsSummary != nil {
+		f := HandlePodStatsSummary(p.GetStatsSummary)
+		r.HandleFunc("/stats/summary", f).Methods("GET")
+		r.HandleFunc("/stats/summary/", f).Methods("GET")
+	}
+
 	r.NotFoundHandler = http.HandlerFunc(NotFound)
 	return r
 }


cpuguy83:
How about we add the metrics routes to the PodHandler?

ldx:
> How about we add the metrics routes to the PodHandler?

Sure, that's actually where it belongs. We'll have to thread the changes through node-cli, though.

Let me refactor this to move the stats route into PodHandlerConfig as a first step.

ldx:
@cpuguy83 can you take another look now?

node/api/server.go
```diff
@@ -64,27 +65,13 @@ func PodHandler(p PodHandlerConfig, debug bool) http.Handler {
...
 	).Methods("POST", "GET")
-	r.NotFoundHandler = http.HandlerFunc(NotFound)
-	return r
-}
 
-// PodStatsSummaryHandler creates an http handler for serving pod metrics.
-//
-// If the passed in handler func is nil this will create handlers which only
-//  serves http.StatusNotImplemented
-func PodStatsSummaryHandler(f PodStatsSummaryHandlerFunc) http.Handler {
```
cpuguy83, line 75: Let's not delete existing functions.
We can keep the old one for attaching to other

node/api/server.go
```diff
@@ -111,7 +98,7 @@ type PodMetricsConfig struct {
 // Callers should take care to namespace the serve mux as they see fit, however
 // these routes get called by the Kubernetes API server.
 func AttachPodMetricsRoutes(p PodMetricsConfig, mux ServeMux) {
-	mux.Handle("/", InstrumentHandler(HandlePodStatsSummary(p.GetStatsSummary)))
+	mux.Handle("/stats/summary", InstrumentHandler(HandlePodStatsSummary(p.GetStatsSummary)))
```
cpuguy83, line 101: This should remain as is.

node/api/server.go
```diff
@@ -64,6 +65,13 @@ func PodHandler(p PodHandlerConfig, debug bool) http.Handler {
 			WithExecStreamIdleTimeout(p.StreamIdleTimeout),
 		),
 	).Methods("POST", "GET")
+
+	if p.GetStatsSummary != nil {
```
cpuguy83, line 69: Just a nit, `HandlePotStatsSummary` will return a `NotImplemented` handler if this is nil anyway.

PR #858, opened 2020-07-17
feat: systemd notify

LeoLiuYan:


sargun:
We have this in our higher level code that wraps VK. Once VK calls GetPods on us, we send systemd the ready message.

Issue #655, opened 2019-06-05
Introduce a management agent


cpuguy83:
Right now there is a lot of setup for consumers of the core.
It would be nice to introduce some sort of agent for setting up and managing the node and pod controller.

This could basically take a `kubernetes.Client` and node config, then handle the initialization... basically much of what is in `cmd/virtual-kubelet/commands/root/root.go` right now.

For performance the agent could allow callers to optionally pass in things like pre-created informers.

pires:
I like this idea! We should aim to remove any setup that isn't strictly managed by the VK pieces.

cpuguy83:
I don't think we'll need to break anything.
This would be a new object that consumers can use if they want to.
This would end up being a 1.X release.

pires:
Oh, if it's opt-in, OK then.

cpuguy83:
This was added in node/nodeutil

PR #859, opened 2020-07-22
Add controller manager to remove boiler plate code

cpuguy83:
This moves pod and node controller setup to a manager object.
It can also be extended to support configuring the http service.

Related to #655

### DIFF
--- a/cmd/virtual-kubelet/internal/commands/root/root.go
+++ b/cmd/virtual-kubelet/internal/commands/root/root.go
@@ -16,6 +16,8 @@ package root
 
 import (
 	"context"
+	"crypto/tls"
+	"fmt"
 	"os"
 	"path"
 
@@ -26,17 +28,14 @@ import (
 	"github.com/virtual-kubelet/virtual-kubelet/internal/manager"
 	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"github.com/virtual-kubelet/virtual-kubelet/node"
+	"github.com/virtual-kubelet/virtual-kubelet/node/api"
+	"github.com/virtual-kubelet/virtual-kubelet/node/nodeutil"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	kubeinformers "k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/kubernetes/typed/coordination/v1beta1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/record"
 )
 
@@ -79,7 +78,7 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
 		}
 	}
 
-	client, err := newClient(c.KubeConfigPath)
+	client, err := nodeutil.ClientsetFromEnv(c.KubeConfigPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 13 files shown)


cpuguy83:
@sargun WDYT?

sargun:
I'll take a look at this soon / later -- EOW.

cmd/virtual-kubelet/internal/commands/root/root.go
```diff
@@ -142,119 +131,129 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
...
+	c := corev1.NodeCondition{
+		Type:               "Ready",
+		Status:             corev1.ConditionTrue,
+		Reason:             "KubeletReady",
+		Message:            "Kubelet is ready",
+		LastHeartbeatTime:  metav1.Now(),
+		LastTransitionTime: metav1.Now(),
+	}
+	n.Status.Conditions = append(n.Status.Conditions, c)
 }
```
ldx, line 259: Not related to this change, but the code duplication between node-cli and cmd/virtual-kubelet/internal makes me wonder if node-cli really should be a separate repo.

cpuguy83: The reason for the split is API guarantees.
`node-cli` does not have a stable API, but we wanted to have a way to help people prototype providers, remove some boiler plate, etc.

This PR is really an attempt to bring some of that work into a stable API.

cpuguy83:
Refactoring this a little bit to remove duplication in configs between mnagagers and other controllers.

node/nodeutil/node.go
```diff
@@ -0,0 +1,34 @@
+package nodeutil
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SetNodeReady sets the node ready condition to true
```
sargun, line 8: Did we already have something else like this before? Where was it?

cpuguy83: We did not.

cpuguy83: So I added this so the test suite can determine if the node is ready vs waiting on pod being able to read from /stats/summary (which isn't technically accurate) and thought it would be a good utility. You can see it in use in the cmd code.

cmd/virtual-kubelet/internal/commands/root/root.go
```diff
@@ -142,17 +135,43 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
...
+		pmh = pm.GetStatsSummary
+	}
+
+	tlsConfig, err := loadTLSConfig(os.Getenv("APISERVER_CERT_LOCATION"), os.Getenv("APISERVER_KEY_LOCATION"))
+	if err != nil {
+		return err
+	}
+	l, err := tls.Listen("tcp", fmt.Sprintf(":%d", c.ListenPort), tlsConfig)
+	if err != nil {
+		return errors.Wrap(err, "error setting up listener")
```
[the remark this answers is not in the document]
cpuguy83: We can indeed.

cmd/virtual-kubelet/internal/commands/root/root.go
```diff
@@ -142,17 +135,43 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
...
+	var pmh api.PodStatsSummaryHandlerFunc
+	if pm, ok := p.(provider.PodMetricsProvider); ok {
+		pmh = pm.GetStatsSummary
+	}
+
+	tlsConfig, err := loadTLSConfig(os.Getenv("APISERVER_CERT_LOCATION"), os.Getenv("APISERVER_KEY_LOCATION"))
+	if err != nil {
+		return err
+	}
+	l, err := tls.Listen("tcp", fmt.Sprintf(":%d", c.ListenPort), tlsConfig)
```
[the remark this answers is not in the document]
cpuguy83: Yes, we should, though this code is just for the mock provider for e2e.

pires:
@cpuguy83 are you going to pick this up any time soon?

cpuguy83:
We've got all this into node/nodeutil now, so no need for this.

PR #860, opened 2020-07-24
Add contributing guidelines

cpuguy83:
It came up on the monthly call that there is no strong guidelines for
this project on contribution, so this adds some.

Ideally contributors can fall back on this as a palce to get started
when submitting or reviewing changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,6 +77,70 @@ git remote add upstream git@github.com:virtual-kubelet/virtual-kubelet.git
 ```
 3. Submit a pull request.
 
+
+## Submission and Review guidelines
+
+We welcome and appreciate everyone to submit and review changes. Here are some guidelines to follow for help ensure
+a successful contribution experience.
+
+Please note these are general guidelines, and while they are a good starting point, they are not specifically rules.
+If you have a question about something, feel free to ask:
+
+- [#virtual-kubelet](https://kubernetes.slack.com/archives/C8YU1QP8W) on Kubernetes Slack
+- [virtualkubelet-dev@lists.cncf.io](mailto:virtualkubelet-dev@lists.cncf.io)
+- GitHub Issues
+
+#### Don't make breaking API changes.
+
+Since Virtual Kubelet has reached 1.0 it is a major goal of the project to keep a stable API.
+Breaking changes must only be considered if a 2.0 release is on the table, which should only come with thoughtful
+consideration of the projects users as well as maintenance burden.
+
+Also note that behavior changes in the runtime can have cascading effects that cause unintended failures. Behavior
+changes should come well documented and with ample consideration for downstream effects. If possible, they should be
+opt-in.
+
+#### Public APIs
+
+Public API's should be extendable and flexible without requiring breaking changes.
+While we can always add a new function (`Foo2()`), a new type, etc, doing so makes it harder for people to update to
+the new behavior.
+
+Build API interfaces that do not need to be changed to adopt new or improved functionality. Opinions on how a particular
+thing should work should be encoded by the user rather than implicit in the runtime. Defaults are fine, but defaults
+should be overridable.
+
+The smaller the surface area of an API, the easier it is to do more interesting things with it.
+
+#### Building blocks
+
+Don't overload functionality. If something is complicated to setup we can provide helpers or wrappers to do that, but
+don't require users to do things a certain way because this tends to diminish the usefulness, especially as it relates
+to runtimes.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -77,6 +77,57 @@ git remote add upstream git@github.com:virtual-kubelet/virtual-kubelet.git
...
+
+#### Use context.Context
+
+Probably if it is a public/exported API, it should take a `context.Context`. Even if it doesn't need one today, it may
+need it tomorrow, and then we have a breaking API change.
+
+We use `context.Context` for storing loggers, tracing spans, and cancellation all across the project. Better safe
+than sorry: add a `context.Context`.
+
+#### Errors
```
sargun, line 123: Do we intend to move to go1.13 style errors there? Now that some folks converted us to Go 1.13, I would be happy to spend the time to move us, but it might be worth saying what our plan is going forward.

CONTRIBUTING.md
```diff
@@ -77,6 +77,57 @@ git remote add upstream git@github.com:virtual-kubelet/virtual-kubelet.git
...
+We also do not want the maintenance burden of every users individual edge cases.
+
+#### Use context.Context
+
+Probably if it is a public/exported API, it should take a `context.Context`. Even if it doesn't need one today, it may
+need it tomorrow, and then we have a breaking API change.
+
+We use `context.Context` for storing loggers, tracing spans, and cancellation all across the project. Better safe
+than sorry: add a `context.Context`.
+
```
sargun, line 122: Please add a section around testing.

CONTRIBUTING.md
```diff
@@ -77,6 +77,57 @@ git remote add upstream git@github.com:virtual-kubelet/virtual-kubelet.git
...
+We welcome and appreciate everyone to submit and review changes. Here are some guidelines to follow for help ensure
+a successful contribution experience.
+
+Please note these are general guidelines, and while they are a good starting point, they are not specifically rules.
+If you have a question about something, feel free to ask.
+
+#### Don't make breaking API changes.
+
+Since Virtual Kubelet has reached 1.0 it is a major goal of the project to keep a stable API.
```
sargun, line 91: I feel like this is a nicer way of saying don't fuck it up. I think one thing we learned from the events change for example, is that simple changes, even though API compatible can result in massive cascading failures due to change in behaviour of the system.

I think that saying be careful about changing the behaviour of the system, and if you're worried it might impact the system, make it clear how it'll change the behaviour of the system, and consider adding a knob to change the behaviour.

CONTRIBUTING.md
```diff
@@ -77,6 +77,57 @@ git remote add upstream git@github.com:virtual-kubelet/virtual-kubelet.git
...
 3. Submit a pull request.
 
+
+## Submission and Review guidelines
+
+We welcome and appreciate everyone to submit and review changes. Here are some guidelines to follow for help ensure
+a successful contribution experience.
+
+Please note these are general guidelines, and while they are a good starting point, they are not specifically rules.
+If you have a question about something, feel free to ask.
```
sargun, line 87: Ask where? ;-)

I would say #virtual-kubelet on Kubernetes Slack? I know we have a CNCF mailing list somewhere too..

CONTRIBUTING.md
cpuguy83: We can. I guess some concrete error in errdefs?

CONTRIBUTING.md
cpuguy83: Yep, I will add more detail on that.

Do we need to fix the events change? 😄

CONTRIBUTING.md
cpuguy83: Added

CONTRIBUTING.md
cpuguy83: Added more detail here.

CONTRIBUTING.md
cpuguy83: Added

CONTRIBUTING.md
sargun: We ended up tactically fixing it by hacking around it with a custom event recorder. I believe in the PR I asked if we could put it behind a flag, and there was no flag. :(
https://github.com/virtual-kubelet/virtual-kubelet/pull/797#issuecomment-572682283

CONTRIBUTING.md
sargun: I would just say use errdefs now, and add concrete errors there. And then we can file an issue to say move to go1.13 errors elsewhere.

sargun:
This is great!

PR #863, opened 2020-07-28
Move node pinging to its own goroutine

sargun:
This moves the job of pinging the node provider into its own
goroutine. If it takes a long time, it shouldn't slow down
leases, and vice-versa.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -28,6 +28,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/typed/coordination/v1beta1"
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/util/retry"
@@ -70,12 +71,19 @@ type NodeProvider interface { //nolint:golint
 // Note: When if there are multiple NodeControllerOpts which apply against the same
 // underlying options, the last NodeControllerOpt will win.
 func NewNodeController(p NodeProvider, node *corev1.Node, nodes v1.NodeInterface, opts ...NodeControllerOpt) (*NodeController, error) {
-	n := &NodeController{p: p, n: node, nodes: nodes, chReady: make(chan struct{})}
+	n := &NodeController{
+		p:       p,
+		n:       node,
+		nodes:   nodes,
+		chReady: make(chan struct{}),
+	}
 	for _, o := range opts {
 		if err := o(n); err != nil {
 			return nil, pkgerrors.Wrap(err, "error applying node option")
 		}
 	}
+	n.nodePingController = newNodePingController(n.p, n.pingInterval, n.pingTimeout)
+
 	return n, nil
 }
 
@@ -104,7 +112,17 @@ func WithNodeEnableLeaseV1Beta1(client v1beta1.LeaseInterface, baseLease *coord.
 	}
 }
 
-// WithNodePingInterval sets the interval for checking node status
+// WithNodePingTimeout limits the amount of time that the virtual kubelet will wait for the node provider to
+// respond to the ping callback. If it does not return within this time, it will be considered an error
+// condition
+func WithNodePingTimeout(timeout time.Duration) NodeControllerOpt {
+	return func(n *NodeController) error {
+		n.pingTimeout = &timeout
+		return nil
+	}
+}
+
+// WithNodePingInterval sets the interval between checking for node statuses via Ping()
 // If node leases are not supported (or not enabled), this is the frequency
 // with which the node status will be updated in Kubernetes.
 func WithNodePingInterval(d time.Duration) NodeControllerOpt {
@@ -164,6 +182,9 @@ type NodeController struct { // nolint: golint
 	nodeStatusUpdateErrorHandler ErrorHandler
 
 	chReady chan struct{}
+
+	nodePingController *nodePingController
+	pingTimeout        *time.Duration
 }
 
 // The default intervals used for lease and status updates.
@@ -270,6 +291,10 @@ func (n *NodeController) controlLoop(ctx context.Context) error {
 
 	close(n.chReady)
 
+	group := &wait.Group{}
+	group.StartWithContext(ctx, n.nodePingController.run)
+	defer group.Wait()
+
 	loop := func() bool {
 		ctx, span := trace.StartSpan(ctx, "node.controlLoop.loop")
 		defer span.End()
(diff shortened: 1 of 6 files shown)


node/node.go
```diff
@@ -251,6 +275,39 @@ func (n *NodeController) Ready() <-chan struct{} {
...
 }
 
+// pingLoop does one thing. It pings the provider. It does this until context is cancelled.
+func (n *NodeController) pingLoop(ctx context.Context) {
+	realCheckFunc := func(ctx context.Context) {
+		ctx, span := trace.StartSpan(ctx, "node.pingNode")
+		defer span.End()
+		err := n.p.Ping(ctx)
+		n.lastPingErrorLock.Lock()
+		n.lastPingError = err
```
sargun, line 285: I figure errors best be left uncopied. Hopefully no one is doing anything creative here.

cpuguy83:
The intention of the ping is specifically that it is light weight.
If ping can respond successfully then we keep the node active, otherwise we wait the next interval and try again.

https://github.com/virtual-kubelet/virtual-kubelet/blob/5a39c167a67dfc636325f2bc07aac96b5f5157da/node/node.go#L49-L51

cpuguy83:
Having a timeout on ping that is configurable by the caller does sound reasonable.

sargun:
We basically do a "systemd is-active $FOO" as our ping. 99.999% of the. time this works great. Pings, 99.999% of the time take <10ms (this is great). We also rely on pings to recheck our node conditions (which is asynchronous). If for whatever reason, apiserver has gotten slow, this whole process kind of falls apart, hence above PR.

sargun:
@cpuguy83 What do you think of ^?

node/node.go
```diff
@@ -251,6 +275,39 @@ func (n *NodeController) Ready() <-chan struct{} {
...
+		n.lastPingErrorLock.Unlock()
+		span.SetStatus(err)
+	}
+
+	// We should have a context that we cancel when we pass to the node, this wrapper just determines whether
+	// or not it will be one with a timeout
+	wrappedCheckFunc := func(ctx context.Context) {
+		if n.pingTimeout != nil {
+			ctx, cancel := context.WithTimeout(ctx, *n.pingTimeout)
```
cpuguy83, line 294: I'm not sure about the timeout actually.
Because there is nothing we can do in this loop to ensure that the ping func actually cancels on timeout.
Meanwhile the node provider can define its own timeout.

node/node.go
```diff
@@ -251,6 +275,39 @@ func (n *NodeController) Ready() <-chan struct{} {
...
+			defer cancel()
+			realCheckFunc(ctx)
+		} else {
+			ctx, cancel := context.WithCancel(ctx)
```
cpuguy83, line 298: Any particular reason for cancellation?

node/node.go
```diff
@@ -164,6 +182,12 @@ type NodeController struct { // nolint: golint
 	nodeStatusUpdateErrorHandler ErrorHandler
 
 	chReady chan struct{}
+
+	// What was the last error we got from the node upon ping
+	lastPingErrorLock  sync.RWMutex
```
cpuguy83, line 187: There's only one reader and one writer for this, I think?
Seems like they are mutually exclusive anyway and can use a normal `sync.Mutex`

node/node.go
sargun: We can leave it out for now.

node/node_ping_controller.go
```diff
@@ -0,0 +1,104 @@
...
+			return context.WithTimeout(ctx2, *npc.pingTimeout)
+		}
+	}
+
+	checkFunc := func(ctx context.Context) {
+		ctx, cancel := mkContextFunc(ctx)
+		defer cancel()
+		ctx, span := trace.StartSpan(ctx, "node.pingLoop")
+		defer span.End()
+		doChan := sf.DoChan(key, func() (interface{}, error) {
```
cpuguy83, line 63: So the problem here is we still can't cancel the goroutine.
Maybe let's add a doc to the `Ping` interface to say that implementations must cancel and return the context error when the context is cancelled.

I think I'd rather have bad implementations block here than leaking these goroutines.

sargun: I totally agree with you abut leaking goroutines. In this implementation we will never have more than 1 ping call in-flight, the singeflight prohibits that.

All the cancel does is that it asks the Node implementation nicely to try to cancel the current singleflight.

sargun: tl;dr: This will not leak more than one goroutine "worst case scenario"

sargun: To think this through -- we cannot pre-empt the call to `Ping()`, _but_ we can make sure the previous call to ping returns before invoking another call to `Ping()`. The code isn't fully plumbed here, but my rough plan was:

1. We take a timestamp prior to pinging the node
2. When the ping is completed we record the timestamp when we started the ping
3. We give the user the option to base the lease renewal timestamp not on the node ping response timestamp.

node/node_test.go
```diff
@@ -622,6 +623,50 @@ func TestManualConditionsPreserved(t *testing.T) {
...
+
+func TestNodePingSingleInflight(t *testing.T) {
+	const pingTimeout = 100 * time.Millisecond
+	c := testclient.NewSimpleClientset()
+	testP := &testNodeProviderPing{}
+
+	var calls, finished int64
+	testP.customPingFunction = func(context.Context) error {
+		atomic.AddInt64(&calls, 1)
+		time.Sleep(2 * time.Second)
```
cpuguy83, line 635: Could we use some synchronization here instead of sleeps?

It seems like you want this to timeout.
So it could wait for `<-ctx.Done()`.
And/or it could wait for a singal after the first result returns.

node/node_test.go
```diff
@@ -622,6 +623,50 @@ func TestManualConditionsPreserved(t *testing.T) {
...
+	assert.NilError(t, err)
+	timeTakenToCompleteFirstPing := time.Since(start)
+	assert.Assert(t, timeTakenToCompleteFirstPing < pingTimeout*5, "Time taken to complete first ping: %v", timeTakenToCompleteFirstPing)
+
+	assert.Assert(t, cmp.Error(firstPing.error, context.DeadlineExceeded.Error()))
+	assert.Assert(t, is.Equal(int64(1), atomic.LoadInt64(&calls)))
+	assert.Assert(t, is.Equal(int64(0), atomic.LoadInt64(&finished)))
+
+	// Wait until the first sleep finishes
+	for atomic.LoadInt64(&finished) == 0 {
```
cpuguy83, line 662: Just a test, but this is going to cause a tight loop and eat up CPU.
Should be better on go1.14, but really problematic <=go1.13.

Maybe a cond variable or a a channel to synchronize this.

node/node_test.go
sargun: I'll clean this up.

node/node_test.go
sargun: We have waitable int, in mock_test.go. Mind if I just use that?

cpuguy83: As long as it works I'm happy 😸

PR #866, opened 2020-07-28
Support custom filter for pod event handlers

cpuguy83:
This allows users who have a shared informer that is *not* filtering on
node name to supply a filter for event handlers to ensure events do not
fire for pods not scheduled to the node.

### DIFF
--- a/node/podcontroller.go
+++ b/node/podcontroller.go
@@ -88,6 +88,12 @@ type PodNotifier interface {
 	NotifyPods(context.Context, func(*corev1.Pod))
 }
 
+// PodEventFilterFunc is used to filter pod events received from Kubernetes.
+//
+// Filters that return true means the event handler will be run
+// Filters that return false means the filter will *not* be run.
+type PodEventFilterFunc func(context.Context, *corev1.Pod) bool
+
 // PodController is the controller implementation for Pod resources.
 type PodController struct {
 	provider PodLifecycleHandler
@@ -114,6 +120,8 @@ type PodController struct {
 	// From the time of creation, to termination the knownPods map will contain the pods key
 	// (derived from Kubernetes' cache library) -> a *knownPod struct.
 	knownPods sync.Map
+
+	podEventFilterFunc PodEventFilterFunc
 
 	// ready is a channel which will be closed once the pod controller is fully up and running.
 	// this channel will never be closed if there is an error on startup.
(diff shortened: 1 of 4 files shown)


node/podcontroller.go
```diff
@@ -300,7 +313,16 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 				pc.deletionQ.Forget(key)
 			}
 		},
-	})
+	}
+
+	if pc.eventFilterFunc != nil {
+		eventHandler = cache.FilteringResourceEventHandler{
```
sargun, line 319: One thing that's neat about how this works in K8s, is if the filter function changes (for example, it says after 8pm, pods can start) the thing will re-sync. 

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/config/apiserver.go#L37-L48

Requests:
1. Can we ask the filter function to have a context (passed to it)
2. Can we maybe wrap the filter function and only allow pods through or some such?

node/podcontroller.go
```diff
@@ -115,6 +115,8 @@ type PodController struct {
 	// (derived from Kubernetes' cache library) -> a *knownPod struct.
 	knownPods sync.Map
 
+	eventFilterFunc func(interface{}) bool
```
sargun, line 118: It seems like this should be defined as a type?

node/pod.go
```diff
@@ -390,3 +390,15 @@ func (pc *PodController) deletePodHandler(ctx context.Context, key string) (retE
 	}
 	return nil
 }
+
+// FilterPodsForNodeName creates an event filter function that filters pod events such that pod.Sepc.NodeName matches the provided name
+// Use the return value of this as the PodEventFilterFunc in PodControllerConfig
+func FilterPodsForNodeName(name string) func(interface{}) bool {
```
sargun, line 396: This to me almost seems more of a "utils" or "helpers" thing? Also, it'd be neat if these were written in such a way that filters can be composed -- maybe if .filterfunc was an array?

node/podcontroller.go
cpuguy83: 1. How does the context get passed to the filter?
2. I was thinking something like a function that bundles them together... so like `Filter(filters ...FilterFunc) FilterFunc`

node/pod.go
cpuguy83: I was thinking something like a function that bundles them together... so like `Filter(filters ...FilterFunc) FilterFunc`

node/podcontroller.go
cpuguy83: Yep, updated.

node/podcontroller.go
cpuguy83: Added wrapper

node/pod.go
cpuguy83: Moved to nodeutils.
Also added a function to take a list of filters and make one filter.

cpuguy83:
@sargun Can you give this another glance?

cpuguy83:
Updated to add context.Context support.

PR #870, opened 2020-07-31
Add envtest

sargun:
I know it's not an impressive test. It just brings up a node, and
makes sure it registers. Let's do more in the future.


scripts/validate/gomod.sh
```diff
@@ -12,8 +12,6 @@ else
   echo "go mod tidy passed."
 fi
 
-git diff --exit-code go.mod go.sum || exit_code=$?
```
sargun, line 15: This has literally only provided headaches over the time. I don't see any good reason to have it.

node/node_test.go
```diff
@@ -625,6 +625,7 @@ func TestManualConditionsPreserved(t *testing.T) {
 func testNode(t *testing.T) *corev1.Node {
 	n := &corev1.Node{}
 	n.Name = strings.ToLower(t.Name())
+	n.Name = strings.ReplaceAll(n.Name, "/", "-")
```
sargun, line 628: Turns out node names with `/` break the kubernetes.

sargun:
@adrienjt We talked about this in your previous PR.

node/e2e_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+)
+
+var enableEnvTest = flag.Bool("envtest", false, "Enable envtest based tests")
+
+func TestMain(m *testing.M) {
```
cpuguy83, line 23: Maybne these tests should go in their own package?

sargun: I wanted to do that / thought about doing that, but there's private members of pod controller that we want to prod at while running these tests.

cpuguy83: `<-node.chReady` seems like the only private member unless I missing something.
We should have a `Ready()` function for this just like pod controller.

sargun: Okay, I'll try to figure something out.

cpuguy83 (approved):
LGTM

PR #871, opened 2020-07-31
Set Node Leader Owner Reference

sargun:
This sets / updates the node lease owner reference to the current
node. Previously, we did not set this, which had the interesting
problem of leaking node leases on clusters with node churn.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -206,8 +206,7 @@ func (n *NodeController) Run(ctx context.Context) error {
 		return n.controlLoop(ctx)
 	}
 
-	n.lease = newLease(n.lease)
-	setLeaseAttrs(n.lease, n.n, n.pingInterval*5)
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -610,21 +610,60 @@ func newLease(base *coord.Lease) *coord.Lease {
...
+		d := int32(leaseRenewalInterval.Seconds()) * 25
+		lease.Spec.LeaseDurationSeconds = &d
+	}
+
+	if lease.Name == "" {
+		lease.Name = node.Name
 	}
-	if l.Spec.HolderIdentity == nil {
-		l.Spec.HolderIdentity = &n.Name
+	if lease.Spec.HolderIdentity == nil {
```
cpuguy83, line 623: Oh wow... I remember setting this in a prior implementation... 🤔

PR #872, opened 2020-08-17
Delete and recreate lease on conflict

sargun:
This takes a somewhat hamfisted approach at dealing with lease
conflicts. This can happen if "someone" changes the lease underneath
us. Again, this should happen rarely, but it can happen (And does
happen in production systems).

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -424,7 +424,8 @@ func ensureLease(ctx context.Context, leases v1beta1.LeaseInterface, lease *coor
 		case errors.IsNotFound(err):
 			log.G(ctx).WithError(err).Info("Node lease not supported")
 			return nil, err
-		case errors.IsAlreadyExists(err):
+		case errors.IsAlreadyExists(err), errors.IsConflict(err):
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -424,7 +424,7 @@ func ensureLease(ctx context.Context, leases v1beta1.LeaseInterface, lease *coor
 		case errors.IsNotFound(err):
 			log.G(ctx).WithError(err).Info("Node lease not supported")
 			return nil, err
-		case errors.IsAlreadyExists(err):
+		case errors.IsAlreadyExists(err), errors.IsConflict(err):
 			if err := leases.Delete(ctx, lease.Name, metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
```
cpuguy83, line 428: If there's a conflict, likely there should be a warning log.

cpuguy83:
Minor comment, but seems fine.

node/node.go
sargun: I think in either case (already exists, or conflict) there should be a warning.

sargun:
@cpuguy83 Added a warning message.

PR #879, opened 2020-09-17
docs(usage): Update Helm installation instructions and various chart updates

lachie83:
Goal - successfully run through Helm installation steps on the usage docs.

FYI - I've bumped the minor version of the chart as these are non-backwards compatible changes due to the Deployments API change.

Signed-off-by: Lachlan Evenson <lachlan.evenson@microsoft.com>

### DIFF
--- a/charts/virtual-kubelet/Chart.yaml
+++ b/charts/virtual-kubelet/Chart.yaml
@@ -1,5 +1,5 @@
 name: virtual-kubelet
-version: 0.5.0
+version: 0.6.0
 appVersion: 0.5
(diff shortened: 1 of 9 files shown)


pires (approved):

pires:
Thanks, @lachie83! Most likely this slipped because there were no review requests.

cpuguy83:
I would rather remove this since the chart only works for ACI and the ACI repo already has its own helm charts. WDYT?

cpuguy83:
Closing since these charts live int he ACI repo.

PR #891, opened 2020-10-12
docs(e2e): make example e2e.go compilable

DanielMSchmidt:
Hi there,

I am currently in the process of creating a provider (just for fun, nothing serious) and I am following the e2e docs (thank you so much for providing e2e tests 💯).
I found that the embedded go code for the provider is not actually compilable, so I thought a quick fix might spare others a few minutes of research time 👍


cpuguy83 (approved):
LGTM

Thanks!

PR #862, opened 2020-07-27
Add helpers for common setup code

cpuguy83:
- Create a clientset
- Setup pod informer filters
- Setup node lease client
- Add V2 naive node provider for accepting node status updates and use that to wait for node readiness in e2e tests

This splits up some stuff from #859

PR #892, opened 2020-10-23
Docs on E2E tests: ensure tests can start with simple provider

DanielMSchmidt:
If a provider does not provide a metrics endpoint on port 10255 the E2E tests won't start as the pod is not getting ready.

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -14,7 +14,7 @@ To run the E2E test suite, three things are required:
 
 ### Skaffold Folder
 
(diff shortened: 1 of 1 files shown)


cpuguy83:
I think we have a solution for this https://github.com/virtual-kubelet/virtual-kubelet/pull/862

cpuguy83:
There we go, merged #862, so now the node just needs to be in the "Ready" state.

cpuguy83:
This is not needed anymore since we changed detection of node readiness.

PR #894, opened 2020-10-24
Move env var resolver into its own package and Fix resolution of env vars with a value of ""

sargun:
Handle empty environment variables
    
    This is a pretty large refactor of the env code in order to make it
    so that the individual resolvers for environment variables are their
    own functions.
    
    This has two benefits:
     * The code is much shorter, and much more readable
     * Static linters have a much easier time with exhaustiveness checking

### DIFF
--- a/env/env.go
+++ b/env/env.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package node
+package env
 
 import (
 	"context"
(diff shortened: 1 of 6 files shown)


cpuguy83:
This is fairly difficult to effectively review.

I don't think this deserves its own package. The other env PR moves it to a `podutil` package (like `io` -> `ioutil`).

Other than that, code looks OK.
There is a test for the case, and other tests all seem to pass, so 👍

sargun:
@cpuguy83 What do you mean by it doesn't deserve its own package, but the other one moved it to a podutils package? Do you think we should create an envutils package? Or a podutils package that we stuff all of our podutils-related stuff in?

cpuguy83:
> Or a podutils package that we stuff all of our podutils-related stuff in?

That one.

sargun:
@cpuguy83 See: https://github.com/virtual-kubelet/virtual-kubelet/pull/900

PR #895, opened 2020-10-28
Bug: when initialzing vk, use empty clientcmd.ConfigOverrides instead of nil

charleszheng44:
The `NewNonInteractiveClientConfig`(`k8s.io/client-go@v0.18.4/tools/clientcmd/merged_client_builder.go`)  will be called inside the `ClientConfig()`(line 33), whose second argument is `config.overrides.CurrentContext`. Using `nil` here will cause `nil pointer dereference` runtime error.


sargun (approved):

cpuguy83 (approved):
LGTM

PR #898, opened 2020-11-04
Don't skip pods status update if provider failed once but recovered with retry

### LINKED ISSUE #399: Failed pods not reconciled
When a pod fails we set the reason "ProviderFailed" but the pod state still says "Pending".
This state never gets reconciled and a pod deployment will be short until the pod is either deleted by the user or it's state updated to Failed.

turkenh:
More context [here](https://github.com/virtual-kubelet/virtual-kubelet/issues/399#issuecomment-704186791).

Closes #399

Signed-off-by: Hasan Turken <turkenh@gmail.com>

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -195,8 +195,7 @@ func (pc *PodController) deletePod(ctx context.Context, pod *corev1.Pod) error {
 
 func shouldSkipPodStatusUpdate(pod *corev1.Pod) bool {
 	return pod.Status.Phase == corev1.PodSucceeded ||
-		pod.Status.Phase == corev1.PodFailed ||
-		pod.Status.Reason == podStatusReasonProviderFailed
+		pod.Status.Phase == corev1.PodFailed
 }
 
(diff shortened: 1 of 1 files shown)


cpuguy83:
This seems mostly there because it was already there.
I think this is fine.

I'd argue we should never skip a pod status update... however we have the case where we ping the backend every 5 seconds and send a pod status update.... maybe we could move this function call to only run in that case.

sargun (approved):

sargun:
Sweet.

sargun:
AFAICT, this should only come up in the sync case really -- or if it happens at VK startup time, because async notifications should never get this set.

sargun:
@cpuguy83 ^

cpuguy83:
@turkenh Looks like you need to sign the CNCF CLA?

turkenh:
> @turkenh Looks like you need to sign the CNCF CLA?

@cpuguy83 now it is fixed, thanks.

sargun:
Thanks for signing the CNCF CLA.

PR #900, opened 2020-11-06
Move env var code into its own package

sargun:
This creates a new package -- podutils. The env var related code
doesn't really have any business being part of the node package,
and to create a separation of concerns, faster tests, and just
general code isolation and cleanliness, we can move the env
var related code into this package. This change is purely hygiene,
and not logic related.

### DIFF
--- a/internal/podutils/env.go
+++ b/internal/podutils/env.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package node
+package podutils
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


cpuguy83 (approved):
LGTM

node/podutils/env.go
```diff
@@ -71,9 +71,8 @@ const (
 
 var masterServices = sets.NewString("kubernetes")
 
-// populateEnvironmentVariables populates the environment of each container (and init container) in the specified pod.
-// TODO Make this the single exported function of a "pkg/environment" package in the future.
-func populateEnvironmentVariables(ctx context.Context, pod *corev1.Pod, rm *manager.ResourceManager, recorder record.EventRecorder) error {
+// PopulateEnvironmentVariables populates the environment of each container (and init container) in the specified pod.
+func PopulateEnvironmentVariables(ctx context.Context, pod *corev1.Pod, rm *manager.ResourceManager, recorder record.EventRecorder) error {
```
cpuguy83, line 75: Oh, maybe one thing here. `resource` is an internal package.

cpuguy83 (changes requested):

node/podutils/env.go
sargun: I don't see `resource` being referenced?

sargun: Or are you suggesting we make this an internal package?

cpuguy83: Sorry, `manager.ResourceManager` (brain fart).

sargun: Fixed. I think.

cpuguy83: Change works for me. Would still like to get something like @justnoise's change in that makes this a public interface that providers can choose to use... but this is good for now.

Thanks!

cpuguy83 (approved):

PR #901, opened 2020-11-06
Upgrade to golangci-lint v1.32.2

sargun:


cpuguy83 (approved):
LGTM

I'm still not sure I like this error handling, but it's way better.

cpuguy83:
You do have an extra change that's not golint, can you split that?

sargun:
@cpuguy83 that extra change was an issue detected by golang lint. Previously, we were ignoring the error.

sargun:
@cpuguy83 I split it into two commits.

Issue #899, opened 2020-11-06
Pod status out of sync after being marked as not ready by controller manager


turkenh:
---

### Environment summary

Provider: Custom provider for Kubernetes

Version: v1.3.0

K8s Master Info: GKE

Install Method: Raw manifest

### Issue Details

After pods created, if pod status in k8s updated by another entity (e.g. controller manager), status never gets synced from provider, because of this check: https://github.com/virtual-kubelet/virtual-kubelet/pull/824

### Repro Steps

1. Create pod in k8s
2. Provider creates the pod and syncs its status back
3. Pod in k8s ready/running, all fine
4. Virtual kubelet fails to update node status for some time for whatever reason (e.g. network connectivity issues)
5. Virtual node marked as `NotReady` with message: `Kubelet stopped posting node status`
6. kube-controller-manager of k8s, goes and marks all pods as Ready = false:

```
I1105 12:40:27.046070       1 controller_utils.go:139] Updating ready status of pod busybox-545fd8cc6c-zmwdb to false
```

7. Virtual kubelet never sync's status of pod in provider back to k8s because of [this check ](https://github.com/virtual-kubelet/virtual-kubelet/pull/824), hence pod stuck as `NotReady`.

Please note, this is just one possible case, there could be other cases where status of pod in k8s could be updated by another entity than provider/virtual kubelet.

/cc @cwdsuzhou

sargun:
@cwdsuzhou I suggest we completely remove the check, or we add an additional check when we get new pods from K8s to see if our pod status does not match theirs and if we skipped an update, and if so, trigger re-update on our side.

cwdsuzhou:
IMO we should not remove this check. Maybe we should trigger re-renque of pod in provider if pod status in K8s change rather than remove this change. If remove this change, we would get more and more requests of pod status update.

sargun:
@cwdsuzhou Did you see https://github.com/virtual-kubelet/virtual-kubelet/pull/902?

turkenh:
I think this is fixed now with #902

PR #902, opened 2020-11-06
Fix issue #899: Pod status out of sync

sargun:
As described in the issue, if the following sequence happens, we fail to properly
update the pod status in api server:

1. Create pod in k8s
2. Provider creates the pod and syncs its status back
3. Pod in k8s ready/running, all fine
4. Virtual kubelet fails to update node status for some time for whatever reason (e.g. network connectivity issues)
5. Virtual node marked as NotReady with message: Kubelet stopped posting node status
6. kube-controller-manager of k8s, goes and marks all pods as Ready = false:
7. Virtual kubelet never sync's status of pod in provider back to k8s

### DIFF
--- a/node/podcontroller.go
+++ b/node/podcontroller.go
@@ -146,6 +146,7 @@ type knownPod struct {
 	sync.Mutex
 	lastPodStatusReceivedFromProvider *corev1.Pod
 	lastPodUsed                       *corev1.Pod
+	lastPodStatusUpdateSkipped        bool
 }
 
 // PodControllerConfig is used to configure a new PodController.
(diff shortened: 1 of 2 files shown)


sargun:
@cwdsuzhou Please take a look.

sargun:
@cwdsuzhou Please see: https://github.com/virtual-kubelet/virtual-kubelet/pull/902

cwdsuzhou:
@sargun  Thanks, seems similar with my thought. Looks good!

But I think  we can wait for https://github.com/virtual-kubelet/virtual-kubelet/pull/874 to be merged first because this PR would also go to `updateStatus.

node/podcontroller.go
```diff
@@ -146,6 +146,7 @@ type knownPod struct {
 	sync.Mutex
 	lastPodStatusReceivedFromProvider *corev1.Pod
 	lastPodUsed                       *corev1.Pod
+	lastPodStatusUpdateSkipped        bool
```
cwdsuzhou, line 149: I am not sure if we need this field

node/podcontroller.go
```diff
@@ -304,6 +305,26 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
...
+				if !ok {
+					// Pods are only ever *added* to knownPods in the above AddFunc, and removed
+					// in the below *DeleteFunc*
+					panic("Pod not found in known pods. This should never happen.")
+				}
+
+				kPod := obj.(*knownPod)
+				kPod.Lock()
+				if kPod.lastPodStatusUpdateSkipped && !cmp.Equal(newPod.Status, kPod.lastPodStatusReceivedFromProvider.Status) {
+					// The last pod from the provider -> kube api server was skipped, but we see they no longer match.
```
cwdsuzhou, line 318: A scene is as follow: 

1. Pod in K8s status A
Pod in Provider status A

2. Pod  in Provider Status change to B
Pod in K8s synced to B

3. Pod in K8s changed to C due to some resom.

Here `lastPodStatusUpdateSkipped` is false. So  we would equeue pod in Provider.

IMO, we can remove  `lastPodStatusUpdateSkipped `

sargun: I don't think we should do that, otherwise it will lead to a retry storm. If something outside of VK updates the pod status, it will cause use to update the api server, and if that thing tries to restore the status change, we will do it again.

cwdsuzhou: Thanks for your explanation!

node/podcontroller.go
```diff
@@ -304,6 +305,26 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 			if key, err := cache.MetaNamespaceKeyFunc(newPod); err != nil {
 				log.G(ctx).Error(err)
 			} else {
+				obj, ok := pc.knownPods.Load(key)
+				if !ok {
+					// Pods are only ever *added* to knownPods in the above AddFunc, and removed
+					// in the below *DeleteFunc*
+					panic("Pod not found in known pods. This should never happen.")
```
pires, line 312: Seems a bit extreme. What's the expected result by doing this? VK crashing?

pires (commented):
Overall LGTM but that panic is raising my eyebrow.

node/podcontroller.go
sargun: Yeah, this indicates a critical bug / internal data state corruption.

pires: Gotcha!

cpuguy83 (approved):
LGTM

pires:
What was the force-push all about?

sargun:
@pires re-base. There had been some other changes merged in (I think the allow pod delete from provider) that required it.

sargun:
unfortunately, AFAICT, there's no nice way in Github to hand-rebase and commit to master :(

PR #908, opened 2020-11-12
Fix race between k8s and provider when deleting pod

### LINKED ISSUE #867: Pod recreated during pod.DeletionGracePeriodSeconds will be forcibly deleted
### steps to reproduct it:
1. Run `kubectl delete po testpod` (pod.DeletionGracePeriodSeconds=30s), [these](https://github.com/virtual-kubelet/virtual-kubelet/blob/5a39c167a67dfc636325f2bc07aac96b5f5157da/node/podcontroller.go#L484) code were called to guarantee it would be deleted in the provider and Kubernetes.
```go
	// Check whether the pod has been marked for deletion.
	// If it does, guarantee it is deleted in the provider and Kubernetes.
	if pod.DeletionTimestamp != nil {
		log.G(ctx).Debug("Deleting pod in provider")
		if err := pc.deletePod(ctx, pod); errdefs.IsNotFound(err) {
			log.G(ctx).Debug("Pod not found in provider")
		} else if err != nil {
			err := pkgerrors.Wrapf(err, "failed to delete pod %q in the provider", loggablePodName(pod))
			span.SetStatus(err)
			return err
		}

		pc.deletionQ.AddAfter(key, time.Second*time.Duration(*pod.DeletionGracePeriodSeconds))
		return nil
	}
```
2. In my case, the old pod was successfully deleted in 5s (shorter than pod.DeletionGracePeriodSeconds)
3. Run `kubectl apply -f testpod.yaml`  immediately, the new pod with the same name and namespace was created
4. After the time of pod.DeletionGracePeriodSeconds, [deletePodHandler](https://github.com/virtual-kubelet/virtual-kubelet/blob/5a39c167a67dfc636325f2bc07aac96b5f5157da/node/pod.go#L342) will check whether the pod has been deleted in the kubernetes. However, it got the new one (as the name and namespaces are same).  It will delete the pod directly without any other checks.  
```go
	// If the pod has been deleted from API server, we don't need to do anything.
	k8sPod, err := pc.podsLister.Pods(namespace).Get(name)
	if errors.IsNotFound(err) {
		return nil
	}
	if err != nil {
		span.SetStatus(err)
		return err
	}

	if running(&k8sPod.Status) {
		log.G(ctx).Error("Force deleting pod in running state")
	}

	// We don't check with the provider before doing this delete. At this point, even if an outstanding pod status update
	// was in progress,
	err = pc.client.Pods(namespace).Delete(ctx, name, *metav1.NewDeleteOptions(0))
	if errors.IsNotFound(err) {
		return nil
	}
	if err != nil {
		span.SetStatus(err)
		return err
	}
```

As a result， the new pod was deleted unexpectedly.

### LINKED ISSUE #907: Race issue between k8s and provider when deleting pod
---

### Environment summary

Provider (e.g. ACI, AWS Fargate)

k8s

Version (e.g. 0.1, 0.2-beta)

### Issue Details
1. Create Pod
2. Delete pod and re-create the pod in k8s. 
3. pod re-created may be deleted again by provider

https://github.com/virtual-kubelet/virtual-kubelet/blob/master/node/pod.go#L384

cwdsuzhou:
Fixes #867 
Fixes #907

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -17,6 +17,7 @@ package node
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
@@ -367,7 +368,8 @@ func (pc *PodController) deletePodsFromKubernetesHandler(ctx context.Context, ke
 	ctx, span := trace.StartSpan(ctx, "deletePodsFromKubernetesHandler")
 	defer span.End()
 
-	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	uid, metaKey := getUIDAndMetaNamespaceKey(key)
+	namespace, name, err := cache.SplitMetaNamespaceKey(metaKey)
 	ctx = span.WithFields(ctx, log.Fields{
 		"namespace": namespace,
 		"name":      name,
@@ -397,15 +399,25 @@ func (pc *PodController) deletePodsFromKubernetesHandler(ctx context.Context, ke
 		span.SetStatus(err)
 		return err
 	}
-
+	if string(k8sPod.UID) != uid {
+		log.G(ctx).WithField("k8sPodUID", k8sPod.UID).WithField("uid", uid).Warn("Not deleting pod because remote pod has different UID")
+		return nil
+	}
 	if running(&k8sPod.Status) {
 		log.G(ctx).Error("Force deleting pod in running state")
 	}
 
 	// We don't check with the provider before doing this delete. At this point, even if an outstanding pod status update
 	// was in progress,
-	err = pc.client.Pods(namespace).Delete(ctx, name, *metav1.NewDeleteOptions(0))
+	deleteOptions := metav1.NewDeleteOptions(0)
+	deleteOptions.Preconditions = metav1.NewUIDPreconditions(uid)
+	err = pc.client.Pods(namespace).Delete(ctx, name, *deleteOptions)
 	if errors.IsNotFound(err) {
+		log.G(ctx).Warnf("Not deleting pod because %v", err)
+		return nil
+	}
+	if errors.IsConflict(err) {
+		log.G(ctx).Warnf("There was a conflict, maybe trying to delete a Pod that has been recreated: %v", err)
 		return nil
 	}
 	if err != nil {
@@ -413,4 +425,11 @@ func (pc *PodController) deletePodsFromKubernetesHandler(ctx context.Context, ke
 		return err
 	}
 	return nil
+}
+
+func getUIDAndMetaNamespaceKey(key string) (string, string) {
+	idx := strings.LastIndex(key, "/")
+	uid := key[idx+1:]
+	metaKey := key[:idx]
+	return uid, metaKey
 }
--- a/node/pod_test.go
+++ b/node/pod_test.go
@@ -20,16 +20,20 @@ import (
 	"testing"
 	"time"
 
-	testutil "github.com/virtual-kubelet/virtual-kubelet/internal/test/util"
 	"golang.org/x/time/rate"
 	"gotest.tools/assert"
 	is "gotest.tools/assert/cmp"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
+	core "k8s.io/client-go/testing"
 	"k8s.io/client-go/util/workqueue"
+
+	testutil "github.com/virtual-kubelet/virtual-kubelet/internal/test/util"
 )
 
 type TestController struct {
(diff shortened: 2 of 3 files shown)


node/pod.go
```diff
@@ -374,7 +379,10 @@ func (pc *PodController) deletePodHandler(ctx context.Context, key string) (retE
 		span.SetStatus(err)
 		return err
 	}
-
+	if string(k8sPod.UID) != uid {
+		log.G(ctx).Errorf("Pod uid: %v, desired uid: %v", k8sPod.UID, uid)
```
sargun, line 383: Please use withField

sargun: Also, make this error message "make more sense" - something like:

log.G(ctx).WithField("k8sPodUID", k8sPod.UID).WithField("uid", uid).Warn("Not deleting pod because remote pod has different UID")

node/pod.go
```diff
@@ -344,7 +345,11 @@ func (pc *PodController) deletePodHandler(ctx context.Context, key string) (retE
 	ctx, span := trace.StartSpan(ctx, "processDeletionReconcilationWorkItem")
 	defer span.End()
 
-	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	idx := strings.LastIndex(key, "-")
```
sargun, line 348: I do not recommend we change out the key format. That seems like it could quickly, and easily generate confusion.

sargun (changes requested):
I believe this is the wrong approach because it still allows for a race The deletePodHandler can read from the knownPods map, and get the pod UID / (and resource version?) from there. 

In that, it can populate DeleteField's Precondition:

https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#DeleteOptions

In that, you can specify a UID precondition:

https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Preconditions

Thanks for addressing this work!

sargun:
P.P.S.

There is a slight problem introduced, in that in AddFunc we get a duplicate pod in known pods potentially because a pod with the same key has been added. Perhaps we introduce another syncMap of deleting pods that things are transitioned to? Even at that, I can imagine this being confusing.

node/pod.go
cwdsuzhou: have changed "-" to "/"

cwdsuzhou:
> I believe this is the wrong approach because it still allows for a race The deletePodHandler can read from the knownPods map, and get the pod UID / (and resource version?) from there.
> 
> In that, it can populate DeleteField's Precondition:
> 
> https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#DeleteOptions
> 
> In that, you can specify a UID precondition:
> 
> https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Preconditions
> 
> Thanks for addressing this work!

I do not think race still exists. But use `Preconditions` may be better

node/pod.go
cwdsuzhou: done

node/pod.go
```diff
@@ -344,7 +345,10 @@ func (pc *PodController) deletePodHandler(ctx context.Context, key string) (retE
 	ctx, span := trace.StartSpan(ctx, "processDeletionReconcilationWorkItem")
 	defer span.End()
 
-	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	idx := strings.LastIndex(key, "/")
```
sargun, line 348: I'm reluctant to mess with this in any way. I yield to @cpuguy83 if you want to pass messages through the queue like this.

node/pod.go
```diff
@@ -374,17 +378,25 @@ func (pc *PodController) deletePodHandler(ctx context.Context, key string) (retE
 		span.SetStatus(err)
 		return err
 	}
-
+	if string(k8sPod.UID) != uid {
```
sargun, line 381: Why not just rely on the precondition check?

cwdsuzhou: If we can not pass this check, we do not need send a request to k8s again.

node/podcontroller.go
```diff
@@ -313,10 +313,11 @@ func (pc *PodController) Run(ctx context.Context, podSyncWorkers int) (retErr er
 			if key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(pod); err != nil {
 				log.G(ctx).Error(err)
 			} else {
+				k8sPod := pod.(*corev1.Pod)
 				pc.knownPods.Delete(key)
 				pc.k8sQ.AddRateLimited(key)
 				// If this pod was in the deletion queue, forget about it
-				pc.deletionQ.Forget(key)
+				pc.deletionQ.Forget(fmt.Sprintf("%v/%v", key, k8sPod.UID))
```
sargun, line 320: Maybe a different approach to stuffing the key in here:

1. You add the pod to the deletionQ
2. The pod gets popped off the deletionQ
3. When the pod gets popped off the deletionQ, you load it from knownPods
4. You check pod.DeletionTimestamp != nil && !running(&pod.Status) -- Maybe even DeletionTimestamp >= time.Now()?

sargun:
@cwdsuzhou I made a comment on how I think this can be simpler.

node/podcontroller.go
cwdsuzhou: Thanks for you suggestions!

But I think pod UID is the real points we concerns. If we did not check uid, we may incorrectly delete a pod.

cwdsuzhou:
@cpuguy83  PTAL thanks

node/pod.go
pires: The request is not send to API server directly but instead to the local cache which IIRC will validate the `deleteOptions`.

pires (changes requested):
Please, implement at least one e2e test that proves this fix.

cwdsuzhou:
> > Please, implement at least one e2e test that proves this fix.
> 
> Thanks! IMO this may be hard to add e2e tests since there are chances that the pod controller works correctly.
> 
> Will have a try

Add some unit tests

node/pod.go
sargun: @pires Are there any docs about that? Can't that lead to a different kind of race?

node/pod.go
```diff
@@ -397,15 +401,25 @@ func (pc *PodController) deletePodsFromKubernetesHandler(ctx context.Context, ke
...
 	}
 
 	// We don't check with the provider before doing this delete. At this point, even if an outstanding pod status update
 	// was in progress,
-	err = pc.client.Pods(namespace).Delete(ctx, name, *metav1.NewDeleteOptions(0))
+	deleteOptions := metav1.NewDeleteOptions(0)
+	deleteOptions.Preconditions = metav1.NewUIDPreconditions(uid)
+	err = pc.client.Pods(namespace).Delete(ctx, name, *deleteOptions)
 	if errors.IsNotFound(err) {
+		log.G(ctx).Warnf("Not deleting pod because %v", err)
```
sargun, line 418: WithError?

node/pod_test.go
```diff
@@ -20,16 +20,20 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
+	core "k8s.io/client-go/testing"
 	"k8s.io/client-go/util/workqueue"
+
```
sargun, line 35: Please drop this new line

sargun (commented):
Other than the mechanism of adding metadata to the queue item to dedupe it, this looks okay to me apart from some minor nits. I yield to @pires or @cpuguy83 since they originally built the WQ based mechanism

cwdsuzhou:
@sargun PR addressed

sargun:
@cpuguy83 ^

pires:
I plan to review this over the weekend.

pires:
I didn't get the chance to review, I'm sorry. Will try again during the week or weekend.

cwdsuzhou:
> I didn't get the chance to review, I'm sorry. Will try again during the week or weekend.

Hi @pires  could you help review this PR this week?  thanks!

node/pod.go
pires: I just realized we're calling the `client` directly here, so it's a REST call and no cache is involved.

pires: I think it's fine to send a request to Kubernetes API as this is a corner case and after the key is processed it shouldn't show up again.

node/pod.go
```diff
@@ -397,15 +401,25 @@ func (pc *PodController) deletePodsFromKubernetesHandler(ctx context.Context, ke
...
-	err = pc.client.Pods(namespace).Delete(ctx, name, *metav1.NewDeleteOptions(0))
+	deleteOptions := metav1.NewDeleteOptions(0)
+	deleteOptions.Preconditions = metav1.NewUIDPreconditions(uid)
+	err = pc.client.Pods(namespace).Delete(ctx, name, *deleteOptions)
 	if errors.IsNotFound(err) {
+		log.G(ctx).WithError(err).Error("Not deleting pod")
+		return nil
+	}
+	if errors.IsConflict(err) {
+		log.G(ctx).WithError(err).Error("Not deleting pod")
```
pires, line 422: I'd use a .`Warn` stating why the conflict is happening, eg _there was a conflict, maybe trying to delete a Pod that has been recreated_.

pires (changes requested):
I agree with @sargun initial review and I ask for the removal of code that injects/reads Pod UID into queue keys and instead simply pass the appropriate `DeleteOptions` with a `Precondition` that ensures the Pod UID is set.

node/pod.go
pires: That said, the Pod UID must come from somewhere and I'm not sure where because:
- `knownPods` shouldn't have the _old_ Pod around
- there's no way to access the provider's `deletedPods` map

So, ignore my last comment.

pires (approved):
I don't like most of this change but it does seems to fix this particular issue, and, right now, I can't think of any solution that doesn't require some deeper thinking into how we handle Pod events. And I do think that it will take some time to do that, including coming up with a design that covers as many scenarios as we can come up with, before any execution is taken on.

pires:
@sargun @cpuguy83 is there something you want to add?

node/pod.go
cwdsuzhou: done, log update

cwdsuzhou:
> I don't like most of this change but it does seems to fix this particular issue, and, right now, I can't think of any solution that doesn't require some deeper thinking into how we handle Pod events. And I do think that it will take some time to do that, including coming up with a design that covers as many scenarios as we can come up with, before any execution is taken on.

Thanks,  this is not the best solution, but this can solve the issue without changing the vk designation too much.

cwdsuzhou:
@sargun @pires @cpuguy83 
do you have any other suggestions ? or could you help approve this PR?

Thanks!

pires:
@sargun implicitly LGTM'ed and I have done it explicitly so I'm merging.

PR #913, opened 2020-11-16
Refactor node ping controller

sargun:
Refactor the way that the that node_ping_controller works
    
    This moves node ping controller to using the new internal lock
    API.
    
    The reason for this is twofold:
    * The channel approach that was used to notify other
      controllers of changes could only be used once (at startup),
      and couldn't be used in the future to broadcast node
      ping status. The idea idea is here that we could move
      to a sync.Cond style API and only wakeup other controllers
      on change, as opposed to constantly polling each other
    * The problem with sync.Cond is that it's not context friendly.
      If we want to do stuff like wait on a sync.cond and use a context
      or a timer or similar, it doesn't work whereas this API allows
      context cancellations on condition change.
    
    The idea is that as we have more controllers that act as centralized
    sources of authority, they can broadcast out their state.

### DIFF
--- a/internal/lock/monitor.go
+++ b/internal/lock/monitor.go
@@ -1,0 +1,98 @@
+package lock
+
+import (
+	"sync"
+)
+
+// NewMonitorVariable instantiates an empty monitor variable
+func NewMonitorVariable() MonitorVariable {
+	mv := &monitorVariable{
+		versionInvalidationChannel: make(chan struct{}),
+	}
+	return mv
+}
+
+// MonitorVariable is a specific monitor variable which allows for channel-subscription to changes to
+// the internal value of the MonitorVariable.
+type MonitorVariable interface {
+	Set(value interface{})
+	Subscribe() Subscription
+}
+
+// Subscription is not concurrency safe. It must not be shared between multiple goroutines.
+type Subscription interface {
+	// On instantiation, if the value has been set, this will return a closed channel. Otherwise, it will follow the
+	// standard semantic, which is when the Monitor Variable is updated, this channel will close. The channel is updated
+	// based on reading Value(). Once a value is read, the channel returned will only be closed if a the Monitor Variable
+	// is set to a new value.
+	NewValueReady() <-chan struct{}
+	// Value returns a value object in a non-blocking fashion. This also means it may return an uninitialized value.
+	// If the monitor variable has not yet been set, the "Version" of the value will be 0.
+	Value() Value
+}
+
+type Value struct {
+	Value   interface{}
+	Version int64
+}
+
+type monitorVariable struct {
+	lock         sync.Mutex
+	currentValue interface{}
+	// 0 indicates uninitialized
+	currentVersion             int64
+	versionInvalidationChannel chan struct{}
+}
+
+func (m *monitorVariable) Set(newValue interface{}) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	m.currentValue = newValue
+	m.currentVersion++
+	close(m.versionInvalidationChannel)
+	m.versionInvalidationChannel = make(chan struct{})
+}
+
+func (m *monitorVariable) Subscribe() Subscription {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	sub := &subscription{
+		mv: m,
+	}
+	if m.currentVersion > 0 {
+		// A value has been set. Set the first versionInvalidationChannel to a closed one.
+		closedCh := make(chan struct{})
+		close(closedCh)
+		sub.lastVersionReadInvalidationChannel = closedCh
+	} else {
+		// The value hasn't yet been initialized.
+		sub.lastVersionReadInvalidationChannel = m.versionInvalidationChannel
+	}
+
+	return sub
+}
+
+type subscription struct {
+	mv                                 *monitorVariable
+	lastVersionRead                    int64
+	lastVersionReadInvalidationChannel chan struct{}
+}
+
+func (s *subscription) NewValueReady() <-chan struct{} {
+	/* This lock could be finer grained (on just the subscription) */
+	s.mv.lock.Lock()
+	defer s.mv.lock.Unlock()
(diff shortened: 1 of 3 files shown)


sargun:
This adds a brand new locking primitive / implementation that makes it much easier for lots of people to subscribe in a context safe way to changes.

sargun:
I'd say 90% of this is the change to add the new broadcast based locking primitive. Most of that code is tests.

The test of this is just shuffling around files.

cpuguy83 (changes requested):
I'm not a fan of splitting packages.
A package is a grouping, and these items belong grouped together.

Can you add, as part of the commit message, the reasoning for moving to the new locking behavior?

sargun:
@cpuguy83 There are two package splits here. There is one package split for the controllers, and there is another package split for the types.

I can add comments to the commit / merge message for both.

The package splits are basically a matter of simplifying testing, and interfaces. When bringing the k8s code into our repo, it was not easy to figure out what part of the interface was public and what was private. In the k8s repo, they split their controllers into packages.

This led to a cascading effect because the leasecontroller package has a dependency on the PingResult. If Pingresult stayed in node/ it resulted in resulted in an import loop.


Another part of it is testing and linting speed.

sargun:
@cpuguy83 Slimmed it down to just the change related to node ping controller.

pires (approved):
LGTM and thanks for the detailed commit message. I understand better what the purpose is now. A suggestion would be to add said reasoning to a `lock/doc.go` file.

lock/cond.go
```diff
@@ -0,0 +1,162 @@
...
+
+type Ticket interface {
+	// Release releases the ticket. The ticket must not be released multiple times.
+	Release()
+	// Wait waits for someone to wakeup the context. It is not concurrency safe [in that multiple goroutines cannot
+	// wait on the same ticket]. If the condition is woken up, no error is returned, the ticket remains valid.
+	// The ticket now holds an exclusive lock on the condition, and it must be released for others to
+	// make progress.
+	// Otherwise, ctx.Err() is returned by Wait, and the ticket is rendered invalid.
+	Wait(context.Context) error
```
cpuguy83, line 19: It seems that `Wait` and `Waiter` are exclusive in the implementation.
I'm not sure in what case another package would want to call `Waiter()` and seems like `Waiter` does not need to be exposed as an interface?

lock/cond.go
```diff
@@ -0,0 +1,162 @@
...
+import (
+	"container/list"
+	"context"
+	"sync"
+
+	"golang.org/x/sync/semaphore"
+)
+
+type Ticket interface {
+	// Release releases the ticket. The ticket must not be released multiple times.
```
cpuguy83, line 12: We are (transitively) violating this in the ping controller with `defer ticket.Release()` followed by `ticket.Wait()`.

cpuguy83 (changes requested):
At the moment we don't really seem to be taking advantage of this new interface since it's still being used in the same way as the channel implementation (lock => check if value is set => wait).
Is there some other plan here?

If we take this in, I'd prefer it to be in an internal package as well.

lock/cond_test.go
```diff
@@ -0,0 +1,361 @@
+package lock_test
```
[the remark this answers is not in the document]
sargun: I can move it into an internal package.

lock/cond.go
sargun: Waiter can be used for 1:N style broadcast with channels. This isn't used right now, but it's used in a future controller (When we start doing multiple controllers listening to a change.

lock/cond.go
sargun: I don't see how it's released multiple times?

lock/cond.go
sargun: I removed the interface.

sargun:
@cpuguy83 It's used once we add the lease v1 controller because the lease v1 controller needs the broadcast interface.

sargun:
I'm going to try to simplify this.

lock/cond.go
cpuguy83: Because `Waiter()` is called which calls `Release` (as I recall).

sargun:
@cpuguy83 Greatly simplified (IMHO).

sargun:
@cpuguy83 if you could take a look at this now that it's post-holiday.

sargun:
@cpuguy83 I made further changes / improvements after our conversation in slack.

internal/lock/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"sync"
+)
+
+// MonitorFunc is passed to a MonitorVariable. A MonitorFunc receives values on the value chan. If the value is not
+// yet set / initialized, it will block until one is set. Once a value is set it will receive updates. If a value
+// is not read off of the value channel, certain intermediate values may be skipped, but the latest value
+// will be put on the channel.
+
+func NewMonitorVariable() MonitorVariable {
```
cpuguy83, line 12: Not sure why the linter isn't complaining about the doc string here and below.

internal/lock/monitor.go
```diff
@@ -0,0 +1,100 @@
+package lock
+
+import (
+	"sync"
+)
+
+// MonitorFunc is passed to a MonitorVariable. A MonitorFunc receives values on the value chan. If the value is not
```
cpuguy83, line 7: This doesn't look true anymore.

cpuguy83 (changes requested):
This looks ok, just some issues with the documentation.

internal/lock/monitor.go
pires: True, had forgotten to submit my review but glad you picked it up.

sargun:
@cpuguy83 Fixed the comments.

Also beginning to fix the linters here:
https://github.com/virtual-kubelet/virtual-kubelet/pull/923

There are two more steps (Need to upgrade the tool, and can probably use gobin's on-demand install feature now, and it looks like the linter for it (golint) needs to have it turned on manually.)

PR #917, opened 2020-11-16
Add V1 Lease Controller

sargun:


internal/leasecontroller/lease_controller.go
```diff
@@ -0,0 +1,335 @@
...
+type LeaseController interface {
+	// leaseControllerStatus returns whether the Lease Controller has successfully been able to establish a lease.
+	LeaseControllerStatus(ctx context.Context) Status
+	// Run starts the lease controller. It blocks until context is cancelled, or if the lease cannot be established.
+	Run(ctx context.Context)
+}
+
+// NewV1BetaV1LeaseController creates a new lease controller.
+//
+// shouldUpdateOrEstablishLeaseFunc is a function that may block, but will be called every lease update interval,
```
pires, line 40: `shouldUpdateOrEstablishLeaseFunc` doesn't exist (anymore?) and it's actually `Run`.

Also, my understanding of the code below is that the `Run` function is indeed blocking on `lc.poll(ctx)` which means this function may actually not run within the lease update interval, i.e. `lc.poll(ctx)` didn't complete within lease interval.

internal/leasecontroller/lease_controller.go
```diff
@@ -0,0 +1,335 @@
...
+	if err != nil {
+		return fmt.Errorf("Received error when attempting to ascertain node status: %w", err)
+	}
+
+	if pr.Error != nil {
+		return newNodeNodeReady(pr)
+	}
+
+	lc.lease.Spec.RenewTime = &metav1.MicroTime{Time: time.Now()}
+	// This is 25 due to historical reasons. It was supposed to be * 5, but...reasons
```
pires, line 159: What reasons?

internal/nodepingcontroller/node_ping_controller.go
```diff
@@ -0,0 +1,136 @@
...
+func (npc *nodePingController) Run(ctx context.Context) {
+	const key = "key"
+	sf := &singleflight.Group{}
+
+	// 1. If the node is "stuck" and not responding to pings, we want to set the status
+	//    to that the node provider has timed out responding to pings
+	// 2. We want it so that the context is cancelled, and whatever the node might have
+	//    been stuck on uses context so it might be unstuck
+	// 3. We want to retry pinging the node, but we do not ever want more than one
+	//    ping in flight at a time.
```
pires, line 62: This comment is concise and clear but the code to do it (cond, ticker, waiter) requires a bit of mental effort to follow, for a dumb programmer such as myself.

pires (commented):
Overall it does look like it works.

internal/leasecontroller/lease_controller.go
sargun: someone multiplied times 5 on input..and then on 5 again. In my limited testing in prod, suddenly changing this value doesn't seem to make other components happy.

internal/nodepingcontroller/node_ping_controller.go
pires: Spent some more time w/ said code while reviewing #913. It makes more sense now. Hopefully, having that reviewed and, eventually, merged will reduce the amount of code in this PR.

sargun:
Superceded by https://github.com/virtual-kubelet/virtual-kubelet/pull/933

PR #918, opened 2020-11-16
Remove $Q Makefile command surpression

sargun:
The whole $Q thing is unintuitive and that's not how Makefiles
typically act. It's confusing.


cpuguy83 (approved):
LGTM

PR #919, opened 2020-11-18
In envtest, configure the logger correctly

sargun:
This configures the global logger the same as the local logger,
and adds the test name. It also uses the logger with the test
context as the context logger.


cpuguy83 (approved):
LGTM

PR #922, opened 2020-12-02
Add CodeQL Security Scanning

caniszczyk:
Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,56 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '19 18 * * 3'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go' ]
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go', 'javascript' ]
```
sargun, line 31: Can you remove javascript?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
+# For most projects, this workflow file will not need changing; you simply need
```
sargun, line 1: I think you can remove this note.

sargun (changes requested):
I'm not totally sure what CodeQL does, and where the results of these actions go. Can you add a little bit to your commit comment about that?

sargun:
@caniszczyk Ping?

.github/workflows/codeql-analysis.yml
caniszczyk: removed the note

.github/workflows/codeql-analysis.yml
caniszczyk: done

PR #923, opened 2020-12-03
Enable all linters by default

sargun:
This removes the directive from .golangci.yml to disable all linters,
and fixes the relevant bugs / issues that are exposed.


cpuguy83 (approved):
LGTM

PR #926, opened 2020-12-04
Move v1beta1 lease controller

sargun:
In order to start to decompose the node package, we need
    to remove the types in order to avoid creating circular
    import depedencies. This works around breaking everyone
    but importing the type into node, and "re-exporting" it

    
    This makes reasoning about the controller interface / surface area
    easier. Eventually, we want to consume the upstream lease controller,
    and this is a step in that direction (so we can make it so that
    our API surface are looks similar to theirs.)

### DIFF
--- a/internal/nodepingcontroller/node_ping_controller.go
+++ b/internal/nodepingcontroller/node_ping_controller.go
@@ -1,4 +1,4 @@
-package node
+package nodepingcontroller
 
 import (
 	"context"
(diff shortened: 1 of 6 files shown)


sargun:
@cpuguy83 I know this is controversial, but I hope the explanation above makes it less worrying.

internal/nodepingcontroller/node_ping_controller.go
```diff
@@ -1,4 +1,4 @@
-package node
+package nodepingcontroller
```
pires, line 1: This package name makes me itchy.

node/node.go
```diff
@@ -90,7 +74,7 @@ func NewNodeController(p NodeProvider, node *corev1.Node, nodes v1.NodeInterface
 		n.statusInterval = DefaultStatusUpdateInterval
 	}
 
-	n.nodePingController = newNodePingController(n.p, n.pingInterval, n.pingTimeout)
+	n.nodePingController = nodepingcontroller.NewNodePingController(n.p, n.pingInterval, n.pingTimeout)
```
pires, line 77: A linter must be catching this as stutter. If this is really the package name we're going with, at least `nodepingcontroller.New()` would make more sense to me.

pires (commented):
I understand the goal but I lack context why we'd want to do things this way, sorry. I need to return to our meetings to make sure I catch up.

sargun:
@pires tried to make it..less stuttery.

PR #928, opened 2020-12-05
Add more golint - gocritic

sargun:
* Switch to using Makefile for lint
    
    This uses gobin in the Makefile for golint. Gobin allows for easier
    pinning of dependencies that are project specific, as in the actual
    gobin command invocation you can specify the version.

* Add Gocritic
    
    This also fixes the issues laid out by gocritic


pires:
Looking good.

pires (commented):
Looking good.

sargun:
@pires Only one 'real' change from your initial review. It moves golint to the Makefile so we can make sure local linting (via `make lint`) works with remote things.

pires (approved):
LGTM.

Unrelated to this PR but since we're adding tests I realized we could do coveralls to give us better metrics on code coverage. WDYT?

sargun:
@pires I tried to add coveralls a long time ago, but the way our current E2E test works doesn't work on coveralls for some reason. I think if we could rewrite those against vk_envtest, and just run those as traditional tests, that'd be better.

PR #931, opened 2020-12-07
Refactor V1Beta1 Lease Controller

sargun:
This refactors the v1beta1 lease controller. It makes two functional differences
to the lease controller:
* It no longer ties lease updates to node pings or node status updates
* There is no fallback mechanism to status updates

This is in preparation for V1 leases.

### DIFF
--- a/node/lease_controller.go
+++ b/node/lease_controller.go
@@ -1,0 +1,257 @@
+package node
+
+import (
+	"context"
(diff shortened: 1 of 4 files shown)


pires:
ah ah, sorry but my review just got in and I don't even know why/how this happened :-s

PR #932, opened 2020-12-07
Add v1 lease controller

sargun:


sargun:
This will be tee'd up after: https://github.com/virtual-kubelet/virtual-kubelet/pull/932

PR #941, opened 2021-01-10
log: add klog/v2

### LINKED ISSUE #924: Add klogv2 adapter for logging
Right now we only have a Logrus impl. Add klogv2 since it implements the logr interface.

pires:
Fixes #924


log/klogv2/klogv2.go
```diff
@@ -0,0 +1,95 @@
...
+//
+// We recommend reading the klog conventions to build a better understanding of levels and when they should apply
+// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md
+package klogv2
+
+import (
+	"github.com/virtual-kubelet/virtual-kubelet/log"
+	"k8s.io/klog/v2"
+)
+
```
sargun, line 28: What do you think about doing a:

```go
var _ log.Logger = (*adapter)(nil)
```

sargun: (rather than the test in the other file)

sargun (approved):

log/klogv2/klogv2.go
pires: Just following what was there but yeah I'm all in for compile time validation. On it.

sargun (approved):

pires:
Thank you, @sargun

PR #943, opened 2021-01-10
e2e: test with Kubernetes to 1.20.1

pires:
Also bumped minikube and skaffold.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.0
 jobs:
   validate:
     resource_class: xlarge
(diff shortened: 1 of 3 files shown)


pires:
Interesting find - or so I believe -, our e2e don't run on CCI Ubuntu 20.04 (`ubuntu-2004:202010-01`), I believe because [make `%` wildcard](https://makefiletutorial.com/#pattern-rules) is not properly supported.

sargun (approved):
Pretty impressed this works.

pires:
Makes two of us! Seems we did a great job when deciding to go for CCI, minikube and skaffold 😅

PR #945, opened 2021-01-12
Import kubernetes/remotecommand

miekg:
Copy/paste some more kubernetes code. This is to remove the dep on
kubernetes/kubernetes from within exec.go

See #940

Signed-off-by: Miek Gieben <miek@miek.nl>

Issue #940, opened 2021-01-08
Import of k8s.io/kubernetes breaks Bazel


miekg:
It's impossible (*) to build VK (and systemk) w/ Bazel because the entirety of kubernetes is a dependency.
It seems to be because of this one file https://github.com/virtual-kubelet/virtual-kubelet/blob/master/internal/podutils/env.go

VK probably needs more stuff than the usual k8s client, but I think it worth copy and pasting a few bits and depending on *all* of kubernetes just to get a thing. Yes, I realize 'go build' works.

(*): day of trying

miekg:
in env.go only 3 functions are used, but there are more files:
~~~
test/e2e/basic.go:	"k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
cmd/virtual-kubelet/main.go:	k8sVersion   = "v1.15.2" // This should follow the version of k8s.io/kubernetes we are importing
cmd/virtual-kubelet/internal/provider/mock/mock.go:	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
cmd/virtual-kubelet/internal/provider/provider.go:	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
node/api/exec.go:	api "k8s.io/kubernetes/pkg/apis/core"
node/api/exec.go:	"k8s.io/kubernetes/pkg/kubelet/server/remotecommand"
node/api/stats.go:	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
internal/podutils/env.go:	podshelper "k8s.io/kubernetes/pkg/apis/core/pods"
internal/podutils/env.go:	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
internal/podutils/env.go:	fieldpath "k8s.io/kubernetes/pkg/fieldpath"
internal/podutils/env.go:	"k8s.io/kubernetes/pkg/kubelet/envvars"
internal/podutils/env.go:	"k8s.io/kubernetes/third_party/forked/golang/expansion"
internal/test/e2e/framework/stats.go:	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
internal/test/e2e/framework/pod.go:	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
~~~

cpuguy83:
Yes, api/exec has definitely been on a todo to fix.
[...]

PR #948, opened 2021-01-15
Remove k8s.io/kubernetes dep from env.go

miekg:
Put 3 k8s.io/kubernetes packages in internal/kubernetes to prevent
depending on k8s.io/kubernetes

See: #940

Probably depends on the merge of #945

Signed-off-by: Miek Gieben <miek@miek.nl>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	gotest.tools v2.2.0+incompatible
 	k8s.io/api v0.18.6
 	k8s.io/apimachinery v0.18.6
+	k8s.io/apiserver v0.18.4
 	k8s.io/client-go v0.18.6
 	k8s.io/klog v1.0.0
 	k8s.io/klog/v2 v2.0.0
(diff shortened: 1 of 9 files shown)


sargun:
Can we automate this in any fashion? I know the Kubernetes project itself has some "magic" to do this. In other projects, I've written a mini-tool to do this (example: https://github.com/Netflix/titus-executor/blob/master/fix/main.go)

miekg:
[ Quoting <notifications@github.com> in "Re: [virtual-kubelet/virtual-kubele..." ]
>Can we automate this in any fashion? I know the Kubernetes project itself has
>some "magic" to do this. In other projects, I've written a mini-tool to do this
>(example: https://github.com/Netflix/titus-executor/blob/master/fix/main.go)

Nice - this hadn't crossed my mind. Probably worth adding a comment in each file which
upstream hash we've copied and (maybe?) the exact upstream directory?

(and copying the actual files, etc, etc)

/Miek

--
Miek Gieben

sargun:
Yeah, that's a good idea, but I have no idea how to do that.

sargun:
Any luck investigating automation?

cpuguy83:
@sargun I think the best bet long term is to make a new module separate from k/k and get k/k to use it.
Parsing k8s yaml seems like a really useful thing outside of the kubelet codebase.

PR #951, opened 2021-01-27
Update mailing list link

Jeffwan:
image

It redirects me to `https://github.com/virtual-kubelet/virtual-kubelet/blob/master/virtualkubelet-dev@lists.cncf.io` which is 404 page. I change to mail list page instead.


pires:
OK this works but I think what was missing originally was a `mailto:` scheme before the email address.

pires (approved):
LGTM

cpuguy83 (approved):
LGTM

PR #953, opened 2021-02-02
Split out rate limiter per workqueue

sargun:
I would say this solves a somewhat worrying / urgent bug.

If you share a ratelimiter between workqueues, it breaks.

WQ1: Starts processing item (When)
WQ1: Fails to process item (When)
WQ1: Fails to process item (When)
WQ1: Fails to process item (When)
--- At this point we've backed off a bit ---
WQ2: Starts processing item (with same key, When)
WQ2: Succeeds at processing item (Forget)
WQ1: Fails to process item (When) ---> THIS RESULTS IN AN ERROR

This results in an error because it "forgot" the previous
rate limit.

CC: @cwdsuzhou

### DIFF
--- a/node/pod_test.go
+++ b/node/pod_test.go
@@ -44,11 +44,6 @@ func newTestController() *TestController {
 	rm := testutil.FakeResourceManager()
 	p := newMockProvider()
 	iFactory := kubeinformers.NewSharedInformerFactoryWithOptions(fk8s, 10*time.Minute)
-	rateLimiter := workqueue.NewMaxOfRateLimiter(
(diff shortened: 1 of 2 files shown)


cwdsuzhou:
Thanks @sargun , looks good

pires (approved):
LGTM

cpuguy83:
I knew this would give us some trouble. I'd rather just remove it since we are making a change like this anyway. How useful is providing a custom implementation here?

sargun:
@cpuguy83 How else would you configure it outside of the VK package?

PR #955, opened 2021-02-16
Fix pod status update

sargun:
We had added an optimization that made it so we dedupe pod status updates
from the provider. This ignored two subfields that could be updated along
with status.

Because the details of subresource updating is a bit API server centric,
I wrote an envtest which checks for this behaviour.

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -215,14 +215,7 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes 
 	}
 	kPod := obj.(*knownPod)
 	kPod.Lock()
-
 	podFromProvider := kPod.lastPodStatusReceivedFromProvider.DeepCopy()
-	if cmp.Equal(podFromKubernetes.Status, podFromProvider.Status) && podFromProvider.DeletionTimestamp == nil {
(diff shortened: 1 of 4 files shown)


node/pod.go
```diff
@@ -216,7 +216,10 @@ func (pc *PodController) updatePodStatus(ctx context.Context, podFromKubernetes
 	kPod.Lock()
 
 	podFromProvider := kPod.lastPodStatusReceivedFromProvider.DeepCopy()
-	if cmp.Equal(podFromKubernetes.Status, podFromProvider.Status) && podFromProvider.DeletionTimestamp == nil {
+	if cmp.Equal(podFromKubernetes.Status, podFromProvider.Status) &&
+		podFromProvider.DeletionTimestamp == nil &&
+		cmp.Equal(podFromKubernetes.ObjectMeta.Annotations, podFromProvider.ObjectMeta.Annotations) &&
+		cmp.Equal(podFromKubernetes.ObjectMeta.Labels, podFromProvider.ObjectMeta.Labels) {
```
pires, line 222: I suggest including finalizers too.

pires (approved):
LGTM!

PR #962, opened 2021-04-09
Add the ability to dictate custom retries

sargun:
Our current retry policy is naive and only does 20 retries. It is
also based off of the rate limiter. If the user is somewhat aggressive in
rate limiting, but they have a temporary outage on API server, they
may want to continue to delay.

In facts, K8s has a built-in function to suggest delays:
https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#SuggestsClientDelay

### DIFF
--- a/internal/queue/queue.go
+++ b/internal/queue/queue.go
@@ -35,6 +35,9 @@ const (
 	MaxRetries = 20
 )
 
+// ShouldRetryFunc is a mechanism to have a custom retry policy
+type ShouldRetryFunc func(ctx context.Context, key string, timesTried int, originallyAdded time.Time, err error) (*time.Duration, error)
+
 // ItemHandler is a callback that handles a single key on the Queue
 type ItemHandler func(ctx context.Context, key string) error
 
@@ -61,6 +64,8 @@ type Queue struct {
 
 	// wakeup
 	wakeupCh chan struct{}
+
+	retryFunc ShouldRetryFunc
 }
 
 type queueItem struct {
@@ -83,9 +88,12 @@ func (item *queueItem) String() string {
 
 // New creates a queue
 //
-// It expects to get a item rate limiter, and a friendly name which is used in logs, and
-// in the internal kubernetes metrics.
-func New(ratelimiter workqueue.RateLimiter, name string, handler ItemHandler) *Queue {
+// It expects to get a item rate limiter, and a friendly name which is used in logs, and in the internal kubernetes
+// metrics. If retryFunc is nil, the default retry function.
+func New(ratelimiter workqueue.RateLimiter, name string, handler ItemHandler, retryFunc ShouldRetryFunc) *Queue {
+	if retryFunc == nil {
+		retryFunc = DefaultRetryFunc
+	}
 	return &Queue{
 		clock:                    clock.RealClock{},
 		name:                     name,
@@ -96,6 +104,7 @@ func New(ratelimiter workqueue.RateLimiter, name string, handler ItemHandler) *Q
 		handler:                  handler,
 		wakeupCh:                 make(chan struct{}, 1),
 		waitForNextItemSemaphore: semaphore.NewWeighted(1),
+		retryFunc:                retryFunc,
 	}
 }
 
@@ -104,7 +113,7 @@ func (q *Queue) Enqueue(ctx context.Context, key string) {
 	q.lock.Lock()
 	defer q.lock.Unlock()
 
-	q.insert(ctx, key, true, 0)
+	q.insert(ctx, key, true, nil)
 }
 
 // EnqueueWithoutRateLimit enqueues the key without a rate limit
@@ -112,7 +121,7 @@ func (q *Queue) EnqueueWithoutRateLimit(ctx context.Context, key string) {
 	q.lock.Lock()
 	defer q.lock.Unlock()
 
-	q.insert(ctx, key, false, 0)
+	q.insert(ctx, key, false, nil)
 }
 
 // Forget forgets the key
@@ -142,9 +151,20 @@ func (q *Queue) Forget(ctx context.Context, key string) {
 	span.WithField(ctx, "status", "notfound")
 }
 
+func durationDeref(duration *time.Duration, def time.Duration) time.Duration {
+	if duration == nil {
+		return def
+	}
+
+	return *duration
+}
+
 // insert inserts a new item to be processed at time time. It will not further delay items if when is later than the
 // original time the item was scheduled to be processed. If when is earlier, it will "bring it forward"
-func (q *Queue) insert(ctx context.Context, key string, ratelimit bool, delay time.Duration) *queueItem {
+// If ratelimit is specified, and delay is nil, then the ratelimiter's delay (return from When function) will be used
+// If ratelimit is specified, and the delay is non-nil, then the delay value will be used
+// If ratelimit is false, then only delay is used to schedule the work. If delay is nil, it will be considered 0.
+func (q *Queue) insert(ctx context.Context, key string, ratelimit bool, delay *time.Duration) *queueItem {
 	ctx, span := trace.StartSpan(ctx, "insert")
 	defer span.End()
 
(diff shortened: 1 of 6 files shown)


node/podcontroller.go
```diff
@@ -178,10 +178,18 @@ type PodControllerConfig struct {
 
 	// SyncPodsFromKubernetesRateLimiter defines the rate limit for the SyncPodsFromKubernetes queue
 	SyncPodsFromKubernetesRateLimiter workqueue.RateLimiter
+	// SyncPodsFromKubernetesShouldRetryFunc allows for a custom retry policy for the SyncPodsFromKubernetes queue
+	SyncPodsFromKubernetesShouldRetryFunc queue.ShouldRetryFunc
+
 	// DeletePodsFromKubernetesRateLimiter defines the rate limit for the DeletePodsFromKubernetesRateLimiter queue
 	DeletePodsFromKubernetesRateLimiter workqueue.RateLimiter
+	// DeletePodsFromKubernetesShouldRetryFunc allows for a custom retry policy for the SyncPodsFromKubernetes queue
+	DeletePodsFromKubernetesShouldRetryFunc queue.ShouldRetryFunc
```
cpuguy83, line 187: Can't use an internal type for external config.

cpuguy83: Can you describe the default behavior (when no retry func is provided) in the doc string?

internal/queue/queue.go
```diff
@@ -61,8 +61,16 @@ type Queue struct {
...
 	wakeupCh chan struct{}
+
+	retryFunc ShouldRetryFunc
 }
 
+// ShouldRetryFunc is passed a key, the number of attempts that this item has already had, and the (potentially wrapped)
+// error from the queue handler. If the client should retry, the error value should be nil. Optionally, the function
+// may return a value for the delay, to override the default rate limit backoff. The returned error is bubbled up,
+// and can just be the errors that was passed in, if the function does not care to decorate the error.
+type ShouldRetryFunc func(ctx context.Context, key string, timesTried int, originallyAdded time.Time, err error) (*time.Duration, error)
```
cpuguy83, line 72: Can you explain the usage of the duration pointer instead of value?

internal/queue/queue.go
```diff
@@ -198,15 +231,16 @@ func (q *Queue) insert(ctx context.Context, key string, ratelimit bool, delay ti
...
+		actualDelay := q.ratelimiter.When(key)
+		// Check if delay is overridden
+		if delay != nil {
+			actualDelay = *delay
 		}
-		ratelimitDelay := q.ratelimiter.When(key)
-		span.WithField(ctx, "delay", ratelimitDelay.String())
-		val.plannedToStartWorkAt = val.plannedToStartWorkAt.Add(ratelimitDelay)
-		val.delayedViaRateLimit = &ratelimitDelay
+		span.WithField(ctx, "delay", actualDelay.String())
```
cpuguy83, line 239: We should special case stringers in the tracing so it only does the conversion when we are actually collecting the trace.

cpuguy83 (changes requested):

node/podcontroller.go
sargun: Where should I declare it?

internal/queue/queue.go
sargun: I think this explains it. Or should I be explicit about nil vs. 0 value?

> Optionally, the function
> may return a value for the delay, to override the default rate limit backoff.

internal/queue/queue.go
sargun: How do you think we should special case nil duration? or should nil always just be nil

node/podcontroller.go
sargun: The only thing I can think of (that wouldn't introduce an import loop), is declare a new API package called "queue" under root and declare the ShouldRetryFunc there. Opinion?

cpuguy83: Or define 2 types one for external one for internal.

cpuguy83: Did you have a concern with defining the type twice? I see there's a new package now.

sargun: I do not think we should define types twice. It goes against DRY.

cpuguy83: Seems like the code contortions is worse than a repeat :(
And if there's ever a change it will fail to compile anyway.

sargun: What do you think of moving the queue package out of internal?

sargun: Also, wait, are you sure if I "re-export" the tag in podcontroller.go, it doesn't work? A la:

var ShouldRetryFunc = queue.ShouldRetryFunc

...

cpuguy83: That may work well (or rather `type ShouldRetryFunc = queue.ShouldRetryFunc`

sargun: Done.

cpuguy83: Checked this on godoc, looks good.

internal/queue/queue.go
cpuguy83: We could say use `-1` (or < 0) instead of `nil` to avoid all these pointer conversions.

sargun: There's already code (And tests for it), that allow you to pass a negative delay from the rate limiter back and result in it being processed immediately. That would regress / change this behaviour.

cpuguy83: I don't understand why we did that since 0 should mean no delay. But it is what it is.

node/queue.go
```diff
@@ -0,0 +1,29 @@
...
+// * The number of attempts that this item has already had (and failed)
+// * The (potentially wrapped) error from the queue handler.
+//
+// The return value is an error, and optionally an amount to delay the work.
+// If an error is returned, the work will be aborted, and the returned error is bubbled up. It can be the error that
+// was passed in or that error can be wrapped.
+//
+// If the work item should be is to be retried, a delay duration may be specified. If the value is nil, it will fall
+// back to the default behaviour of the queue, and use the ratelimiter that's configured to determine when to start work.
+type ShouldRetryFunc = queue.ShouldRetryFunc
```
cpuguy83, line 22: What happens when the returned delay is a negative value?

internal/queue/queue.go
sargun: It's the way the kubernetes ratelimiters work, and we kept the contract the same.

node/queue.go
sargun: Fixed.

cpuguy83: Can you point me to what changed? I'm not noticing.

sargun: huh, no idea why it didn't update inline:

> 
> ShouldRetryFunc is a mechanism to have a custom retry policy
> 
>  it is passed metadata about the work item when the handler returns an error. It returns the following:
>  * The key
>  * The number of attempts that this item has already had (and failed)
>  * The (potentially wrapped) error from the queue handler.
> 
>  The return value is an error, and optionally an amount to delay the work.
>  If an error is returned, the work will be aborted, and the returned error is bubbled up. It can be the error that
>  was passed in or that error can be wrapped.
> 
>  If the work item should be is to be retried, a delay duration may be specified. The delay is used to schedule when
>  the item should begin processing relative to now, it does not necessarily dictate when the item will start work.
>  Items are processed in the order they are scheduled. If the delay is nil, it will fall  back to the default behaviour
>  of the queue, and use the rate limiter that's configured to determine when to start work.

cpuguy83: I'm not seeing where we are calling out what a negative value means, only nil.

sargun: It doesn't call out the negative value explicitly, but it says that the ordering is based on now + delay. If the delay is negative, your task will be scheduled in the past, thus pushing it to the front of the queue. Do you think we need to be explicit about it?

cpuguy83: I think it is worth mentioning.

sargun: Added: 
> // If the delay is negative, the item will be scheduled "earlier" than now. This will result in the item being executed
> // earlier than other items in the FIFO work order.

cpuguy83:
Looks like there's a test timeout.

sargun:
Huh, didn't happen again.

CC: @cpuguy83

sargun:
I literally ran:
 go test -run=TestPodLifecycle/Async/testDanglingPodScenarioWithDeletionTimestamp/mockProvider -v -count=1000 -parallel=8 -race ./node 2>&1|tee test.log

and no errors.

sargun:
There's somehow a situation where "time="2021-04-14T20:50:38Z" level=warning msg="skipping sync of pod \"default/my-pod\" in \"Succeeded\" phase"" is happening, and this results in it not being put on the deletionQ. 

I don't see how this is possible.

cpuguy83 (approved):
LGTM

PR #965, opened 2021-04-15
Add logging to help debug queueing problems

sargun:
This adds *logs* for information about which queue is
being used / invoked. Previously this information was
in traces.


internal/queue/queue.go
```diff
@@ -376,6 +377,7 @@ func (q *Queue) handleQueueItem(ctx context.Context) bool {
 	// We do this as the delayed nature of the work Queue means the items in the informer cache may actually be more u
 	// to date that when the item was initially put onto the workqueue.
 	ctx = span.WithField(ctx, "key", qi.key)
+	ctx = log.WithLogger(ctx, log.G(ctx).WithField("key", qi.key))
```
cpuguy83, line 380: Shouldn't the logger already have this key? If not that seems like a bug in the span implementation.

cpuguy83: ping

PR #967, opened 2021-04-30
Move some boiler plate startup logic to nodeutil

cpuguy83:
This makes a controller that handles the startup for the node and pod
controller.
Later if we add an "api controller" it can also be added here.

This is just part of reducing some of the boiler plate code so it is
easier to get off of node-cli.

---

Moves some more stuff from #859

### DIFF
--- a/cmd/virtual-kubelet/internal/commands/root/root.go
+++ b/cmd/virtual-kubelet/internal/commands/root/root.go
@@ -199,32 +199,19 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
 	}
 	defer cancelHTTP()
 
-	go func() {
-		if err := pc.Run(ctx, c.PodSyncWorkers); err != nil && errors.Cause(err) != context.Canceled {
-			log.G(ctx).Fatal(err)
-		}
-	}()
+	cm := nodeutil.NewControllerManager(nodeRunner, pc)
+	go cm.Run(ctx, c.PodSyncWorkers) // nolint:errcheck
 
-	if c.StartupTimeout > 0 {
-		ctx, cancel := context.WithTimeout(ctx, c.StartupTimeout)
-		log.G(ctx).Info("Waiting for pod controller / VK to be ready")
-		select {
-		case <-ctx.Done():
-			cancel()
-			return ctx.Err()
-		case <-pc.Ready():
-		}
+	defer func() {
+		log.G(ctx).Debug("Waiting for controllers to be done")
 		cancel()
-		if err := pc.Err(); err != nil {
-			return err
-		}
-	}
-
-	go func() {
-		if err := nodeRunner.Run(ctx); err != nil {
-			log.G(ctx).Fatal(err)
-		}
+		<-cm.Done()
 	}()
+
+	log.G(ctx).Info("Waiting for controller to be ready")
+	if err := cm.WaitReady(ctx, c.StartupTimeout); err != nil {
+		return err
+	}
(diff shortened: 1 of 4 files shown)


cmd/virtual-kubelet/internal/commands/root/root.go
```diff
@@ -199,40 +199,31 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
...
-		log.G(ctx).Info("Waiting for pod controller / VK to be ready")
-		select {
-		case <-ctx.Done():
-			cancel()
-			return ctx.Err()
-		case <-pc.Ready():
-		}
+	defer func() {
+		log.G(ctx).Debug("Waiting for controllers to be done")
 		cancel()
```
sargun, line 207: Doesn't this cancel normally wait for pc.ready, or ctx.done?

sargun: Er, read further and it was more clear.

node/node.go
```diff
@@ -224,6 +225,8 @@ type NodeController struct { // nolint:golint
 	nodeStatusUpdateErrorHandler ErrorHandler
 
 	chReady chan struct{}
+	chDone  chan struct{}
```
sargun, line 228: Can you add a comment about what this does?

node/node.go
```diff
@@ -249,7 +252,12 @@ const (
 // node status update (because some things still expect the node to be updated
 // periodically), otherwise it will only use node status update with the configured
 // ping interval.
-func (n *NodeController) Run(ctx context.Context) error {
+func (n *NodeController) Run(ctx context.Context) (retErr error) {
+	defer func() {
+		n.err = retErr
```
sargun, line 257: I believe this can trigger the go race detector, as the race detector doesn't treat (closing) channels as synchronization points.

node/nodeutil/controller.go
```diff
@@ -0,0 +1,123 @@
...
+	NodeController *node.NodeController
+	PodController  *node.PodController
+
+	mu    sync.Mutex
+	ready chan struct{}
+	done  chan struct{}
+	err   error
+}
+
+func (c *ControllerManager) init() {
```
sargun, line 22: Isn't this better done in a constructor New()? And hide the type from direct initialization>

cpuguy83: Yeah, originally was wanting to avoid a constructor, then realized I needed a mutex 🤷 oh well.
Updated to add constructor.

node/node.go
cpuguy83: Indeed it does. Fixed to use a mutex.

node/node.go
cpuguy83: Done.

node/nodeutil/controller.go
```diff
@@ -0,0 +1,136 @@
...
+	nc *node.NodeController
+	pc *node.PodController
+
+	ready chan struct{}
+	done  chan struct{}
+	err   error
+}
+
+// NewControllerManager creates a new ControllerManager.
+func NewControllerManager(nc *node.NodeController, pc *node.PodController) (*ControllerManager, error) {
```
pires, line 25: If this is a pure ctor without any logic, why return error?

pires (approved):
LGTM

node/nodeutil/controller.go
cpuguy83: Thanks, updated this.

PR #969, opened 2021-05-05
Copy stats types from upstream.

### LINKED ISSUE #946: `stats.Summary` doesn't exists anymore in upstream kubernetes
This is a follow up from #940

In `node.go`: 
~~~
node/api/stats.go:type PodStatsSummaryHandlerFunc func(context.Context) (*stats.Summary, error)
~~~
`stats.Summary` is used, this pulls in kubernetes/kubernetes. While looking into copying a few files from k8s into VK it seems `stats.Summary` doesn't exist anymore - I'm unclear what the new shiny is though.

I'm inclined to remove this. Is that the best way forward?
Also happy to be pointed to the new shiny and seeing how we can incorporate that.

Is this used btw?

cpuguy83:
This drops another dependency on k8s.io/kubernetes.
This does have the unfortunate side effect that implementers will now
get a compile error until they update their code to use the new type.

Just as a note:

The stats types have moved to k8s.io/kubelet, however the stats types
are only there as of v1.20.
Currently we support older versions than v1.20, and even our go.mod
imports from v1.19.

For now we copy the types in. Later we can remove the type defs and
change them to type aliases to the k8s.io/kubelet types (which prevents
another compile time issue).

Anything relying on type assertions to determine if something implements
this method will, unfortunately, be broken and it will be hard to notice
until runtime. We need to make sure to call this out in the release
notes.

Fixes #946

### DIFF
--- a/node/api/statsv1alpha1/types.go
+++ b/node/api/statsv1alpha1/types.go
@@ -1,0 +1,345 @@
+/*
+Copyright 2015 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package statsv1alpha1
+
+import (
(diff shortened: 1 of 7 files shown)


pires (approved):

node/api/statsv1alpha1/types.go
```diff
@@ -0,0 +1,345 @@
+/*
+Copyright 2015 The Kubernetes Authors.
```
miekg, line 2: do we care this is a one-time copy and we lack automation to keep it up to date?

cpuguy83: No, we'll eventually move to use the package in k8s.io/kubelet.

PR #972, opened 2021-05-17
Remove remaining deps on k8s.io/kubernetes

cpuguy83:
These are mostly helper code for setting up env vars.
There are 2 files copied verbatim, although much of this downward api
stuff is a copy.
We may want to pull this out and do a direct copy of the the code so it
is easier to update and work with upstream to have a shared package that
lives outside of k8s.io/kubernetes for downward api.

This is simpler than #948 since that one is bringing in more deps on k8s.io/kubernetes by bringing in functions we don't need.
Closes #948


cpuguy83:
@miekg

sargun:
Is this ready for review?

cpuguy83:
Yep.

sargun:
Can you add a file in that folder, or a doc or something that indicates where this was vendored from? Also, in your commit, can you point the the upstreams Kubernetes commit you copied them from?

cpuguy83:
Updated.

sargun (approved):

miekg (approved):

PR #973, opened 2021-05-17
Output test results in junit and export to circle

cpuguy83:
This adds an override for `go test` so in CI we can use `gotestsum` which outputs junit formatted results which will be consumed by circleci.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
   test:
     resource_class: xlarge
     docker:
-      - image: circleci/golang:1.15
+      - image: circleci/golang:1.16
         environment:
           GO111MODULE: "on"
     working_directory: /go/src/github.com/virtual-kubelet/virtual-kubelet
@@ -42,13 +42,22 @@ jobs:
       - run:
           name: Build
           command: V=1 make build
+      - run: go install gotest.tools/gotestsum@latest
       - run:
           name: Tests
-          command: V=1 CI=1 make test envtest
+          environment:
+            GOTEST: gotestsum -- -timeout=9m
+            GOTESTSUM_JUNITFILE: output/unit/results.xml
+            GODEBUG: cgocheck=2
+          command: |
+            mkdir -p output/unit
+            V=1 make test envtest
       - save_cache:
           key: test-{{ checksum "go.mod" }}-{{ checksum "go.sum" }}
           paths:
             - "/go/pkg/mod"
+      - store_test_results:
+          path: output
 
   e2e:
     machine:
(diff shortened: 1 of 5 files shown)


cpuguy83:
Example test insights report: https://app.circleci.com/insights/github/virtual-kubelet/virtual-kubelet/workflows/validate_and_test/tests?branch=pull/973

Also failure reports, much better than sifting through the logs:

Job list view:

image

Detail view:

image
image

.circleci/config.yml
```diff
@@ -42,13 +42,20 @@ jobs:
...
           command: V=1 make build
+      - run: make gotestsum
       - run:
           name: Tests
-          command: V=1 CI=1 make test envtest
+          environment:
+            GOTEST: gotestsum -- -timeout=9m
+            GOTESTSUM_JUNITFILE: output/unit/results.xml
+            GODEBUG: cgocheck=2
+          command: mkdir -p output/unit; V=1 make test envtest
```
sargun, line 52: why not make this multiline as well?

.circleci/config.yml
```diff
@@ -121,13 +128,21 @@ jobs:
...
+            GOTEST: gotestsum --
           command: |
             mkdir $HOME/.go
             export PATH=$HOME/.go/bin:${PATH}
             curl -fsSL -o "/tmp/go.tar.gz" "https://dl.google.com/go/go1.15.6.linux-amd64.tar.gz"
             tar -C $HOME/.go --strip-components=1 -xzf "/tmp/go.tar.gz"
             go version
+            mkdir -p output/e2e
+            export GOTESTSUM_JUNITFILE="$(pwd)/output/e2e/results.xml"
+            export PATH="${GOPATH}/bin:${PATH}"
```
sargun, line 141: Any reason not to use gobin?

Makefile
```diff
@@ -196,3 +186,6 @@ fmt:
 .PHONY: lint
 lint: $(gobin_tool)
 	gobin -run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.33.0 run ./...
+
+gotestsum: $(gobin_tool)
+	gobin gotest.tools/gotestsum@v1.6.4
```
sargun, line 191: shouldn't it just be:
`gobin -run gotest.tools/gotestsum@v1.6.4 -- ` -- versus installing it into the path?

sargun (commented):
Go 1.17 is supposed to move everything to `go run`:
https://github.com/golang/go/issues/42088

.circleci/config.yml
cpuguy83: Just because it's short.

Makefile
cpuguy83: Then I have to re-execute the Makefile to run gotestsum.

.circleci/config.yml
cpuguy83: As in `gobin -run`?

sargun: Yep

cpuguy83: I'm just going to update to go1.16 and use `go install`

cpuguy83: *In the circle config*

pires (approved):
Nice!

cpuguy83:
@sargun Are you ok with these changes?

PR #974, opened 2021-05-19
Empty slice declaration via literal

Akiros001:


pires:
Closing given lack of description why we should change this.

PR #975, opened 2021-06-03
Make ControllerManager more useful

cpuguy83:
This changes `ControllerManager` to `Node`.

`Node` is created from a client where the VK lib is responsible for
creating all the things except the client (unless client is nil, then we
use the env client).

This should be a good replacement for node-cli.  It offers a simpler
API.  *It only works with leases enabled* since this seems always
desired, however an option could be added to disable if needed.

The intent of this is to provide a simpler way to get a vk node up and
running while also being extensible. We can slowly add options, but
they should be focussed on a use-case rather than trying to support
every possible scenario... in which case the user can just use the
controllers directly.


cpuguy83:
I think this is ready to go.

cpuguy83:
So the goal here is to be able to do `nodeutil.FromClient(client, provider)`  and have a fully running VK instance.

The nitty gritty details are a bit more than right now that because we need TLS for the http server, and preferably the default would include auth on that. Providers are also typically bootstrapped with the informers created for the pod controller so there was also a chicken/egg problem to deal with.

So instead it currently looks something like this: 
```go
nodeutil.FromClient(client, newProviderFunc, 
    nodeutil.WithTLSConfig( // gives us a standard "secure-ish" tls config to start with
        nodeutil.WithKeyPairFromFile(cert, key), // inject our cert/key into the tls config
        nodeutil.WithCA(pem), // turns on mTLS in the TLS config
    ),
    WithHanlder(nodeutil.WithAuth(myAuth, mux)), // adds our HTTP handler for the API.
    nodeutil.AttachProviderRoutes(mux), // sets up routing to go to the provider.
)
```

With some extra options to set things like informer resync, stream timeouts, etc.
Originally I wanted to create our own http handler if none is provided, however opted to not do that for now since we can't automatically create auth yet and I'd prefer to not have no-auth as the default.

What might be interesting later is to have a default TLS location on disk for cert, key, and ca cert so we can just set that up.
And then of course TLS bootstrapping.
If we like this and can merge it in, I can work on TLS bootstrapping.
However, as is this basically replaces all of the functionality except for the actual CLI flags/commands from node-cli and is significantly more versatile and easier to work with (IMHO).

This moves (and improves upon) a bunch of code that was behind the internal CLI stuff so it is now reusable and "supported" going forward.

cpuguy83:
I have a branch which migrates the ACI provider to use this: https://github.com/virtual-kubelet/azure-aci/pull/150

PR #976, opened 2021-06-10
Fix pod metadata overwritten caused by updateStatus

clarklee92:
For most of the types(like pod) defined early on in Kubernetes, changes to metadata were allowed via function `UpdateStatus`.  [https://github.com/kubernetes/kubernetes/issues/92022](url)
The way to update pod status in VK happens to be updateStatus, which may cause the change to metadata like `Labels` or `Annotations` overwritten back to the old values(This is not inevitable, because the default interval of heartbeat reporting is 5 seconds, and the metadata of pod may be collected after modification). 

Signed-off-by: clarklee92 <clarklee1992@hotmail.com>


clarklee92:
@miekg

clarklee92:
Re-test may be required, testing log indicates that the test node connection has timed out. I tested the UT locally and verified it in production.

pires (approved):
LGTM thank you very much for your contribution.

PR #980, opened 2021-09-03
Queue: Use sync.Cond instead of wake channel

cpuguy83:
The reason the channel was used is because it enables us to select on
multiple things at once to wait for.
Really we don't need to wait for multiple things to happen since we can
use a condition variable to wakeup for any number of reasons.
Since the queue is already waiting for the `ctx.Done()` we can use that
to signal to workers to wake up and exit.


cpuguy83:
This never worked the way I was hoping, so closing.

PR #988, opened 2021-12-09
fix(*): update pod status func change to patch

cbsfly:
Currently, VK uses the `UpdatePodStatus` method to update Pod Status, which updates the Pod metadata part synchronously.
For example, if the user manually modified the Pod's label or annotation when updating the Pod Status, this modification will be overwritten.
Refer to the kubelet implementation and replace the UpdatePodStatus method with the Patch method.

refer to:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/status/status_manager.go#L623
https://github.com/kubernetes/kubernetes/blob/master/pkg/util/pod/pod.go#L33

### DIFF
--- a/node/pod.go
+++ b/node/pod.go
@@ -15,7 +15,9 @@
 package node
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


cbsfly:
@cpuguy83

cpuguy83:
Thank you , this looks like a duplicate of https://github.com/virtual-kubelet/virtual-kubelet/pull/976 so I'll close this one.
If I've made a horrible mistake please let me know 😇

PR #999, opened 2022-05-30
Support environment variable values from 'status.hostIP', 'status.podIP' and 'status.podIPs'

roy2220:
'status.hostIP', 'status.podIP' and 'status.podIPs' are available to containers through environment variables.

Refer to: https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/


cpuguy83:
The problem here is that this function is called before the pod is even created in the provider.
There is no pod status or pod IP yet.

helayoty:
This might be better on the provider implementation side.

cpuguy83:
Agreed we can't do this in the core of VK.
What I really would like to see is to make this a library that providers can call themselves rather than having the core do it like we do today.

I'm going to close this PR because it's not going to solve the problem.
Also as a note, we can't accept PR's until the CLA is signed.

PR #1000, opened 2022-06-01
webhook auth crash

lhzzz:
in function ```setAuth(c.NodeName, apiConfig)```, if the api.CACertPath is not empty, it will return ```nodeutil.WebhookAuth``` , and then ```WebhookAuth``` will use the ```client kubernetes.Interface``` ,but the client is nil,  it will crash on statement ```client.AuthenticationV1().TokenReviews()``` 
image

so to solve it , i add a  client init before the auth option.
if i have something wrong , plz tell me, thanks.


cpuguy83:
Thanks for this PR.
Unfortunately we can't merge until the CLA is signed.
Will you be able to sign the CLA?

PR #987, opened 2021-11-22
expose new route /metrics/resource

shsjshentao:
According to https://github.com/kubernetes/kubernetes/issues/68522, "/stats/summary" will be deprecated in hpa.

metrics-server has changed to /metrics/resource instead of /stats/summary
https://github.com/kubernetes-sigs/metrics-server/commit/a2d732e5cdbfd93a6ebce221e8df0e8b463eecc6

Signed-off-by: blade <blade.shen@ucloud.cn>

PR #1075, opened 2023-02-16
Fnuarnav/virtual kubelet metrics update proposal

fnuarnav:
Added Proposal for Virtual Kubelet Metrics Update
 google doc here: https://docs.google.com/document/d/1GSTR8YJZE6ltVML4Gr1FqZeajUPTJswTFk0D6sCq0gI/edit

PR #1082, opened 2023-03-07
feat: Implement new metrics endpoint for k8s 1.24+

fnuarnav:
Added handlers for `/metrics/resource` endpoint

Refs #1075 

Closes #987

### DIFF
--- a/test/e2e/basic.go
+++ b/test/e2e/basic.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"testing"
 	"time"
-	"to"
 
 	"github.com/virtual-kubelet/virtual-kubelet/internal/podutils"
 	stats "github.com/virtual-kubelet/virtual-kubelet/node/api/statsv1alpha1"
--- a/test/e2e/basic.go
+++ b/test/e2e/basic.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"testing"
 	"time"
+	"to"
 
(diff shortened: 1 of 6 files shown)


pires (commented):
Overall LGTM but can you check why e2e fails?

```
--- FAIL: TestEndToEnd (7.87s)
    suite.go:62: True
    --- FAIL: TestEndToEnd/TestGetMetricsResource (0.80s)
        basic.go:139: the server could not find the requested resource (get pods https:vkubelet-mock-0:10250)
FAIL
```

test/e2e/basic.go
```diff
@@ -111,6 +111,64 @@ func (ts *EndToEndTestSuite) TestGetStatsSummary(t *testing.T) {
...
+		}
+	}()
+
+	// Wait for the "nginx-" pod to be reported as running and ready.
+	if _, err := f.WaitUntilPodReady(pod.Namespace, pod.Name); err != nil {
+		t.Fatal(err)
+	}
+
+	// Grab the stats from the provider.
+	metricsResourceResponse, err := f.GetMetricsResource(ctx)
```
pires, line 137: IIUC in the proposal (#1075) it's mentioned testing metrics retrieval would be done with `kubectl`, which is not the case here. I prefer the code approach as provided here. However, I want to mention that [our CI still runs an outdated version of Kubernetes](https://github.com/virtual-kubelet/virtual-kubelet/blob/master/.github/workflows/ci.yml#L66) and we really can't be sure this works with 1.24+.

PR #1085, opened 2023-03-14
cmd: fix nil pointer during node setup

pires:
I run into this during some testing. After implementing a fix, I realized there was a longstanding PR but it is blocked by the CLA and there hasn't been any response to Brian's request for the CLA to be signed.

Closes #1000.


pires:
Since similar code was approved by Brian, and since we don't seem to have any active contributors, I'll take the liberty to self-merge when all checks are green.

PR #1099, opened 2023-04-03
Fix GetResourceMetrics to return metrics in Prometheus Text Format

### LINKED ISSUE #1098: Metrics not served in Prometheus Text Format
---

The `metrics/resource` should publish the metrics in Prometheus Text format instead of a Json

### Environment summary

Version (e.g. 0.1, 0.2-beta)
` v1.8.1-0.20230328122244-b7030b9dc56c`

Self created helm chart

### Issue Details
The new metrics resource endpoint serializes the metrics in a [Json Format](https://github.com/virtual-kubelet/virtual-kubelet/blob/9c32bfb0aead58d46a1500466b2915d00dcc3d9f/node/api/metrics.go#L43), However, these metrics should instead be served in Promethues text format.

### Repo Steps
- Deploy a Virtual Kubelet provider built with virtual-kubelet package version ` v1.8.1-0.20230328122244-b7030b9dc56c`
- Use `kubetctl proxy`, and navigate to `http://localhost:8001/api/v1/nodes/virtual-kubelet-node/proxy/metrics/resource` . The metrics get calls returns a Json.
- Metrics Server parser expects them to be in [Prometheus text format](https://github.com/kubernetes-sigs/metrics-server/blob/2487f2149c8a5b04559cd1948c17a4568cca3ed8/pkg/scraper/client/resource/decode.go#L61-L85)

sanchitmehta:
Fixes #1098


pires:
There's an end-to-end test that expects JSON, which needs fixing before merging this.

OTOH, I'm waiting on feedback by the author of the feature (who wrote a proposal that was accepted by us) @fnuarnav @helayoty.

pires:
Thank you very much @sanchitmehta this was merged as part of #1101

PR #1102, opened 2023-04-06
add portforwarding support to node/api

### LINKED ISSUE #170: Feature: Implement Port forwarding

aka-somix:
## Summary
This PR adds support for PortForwarding implementation in virtual kubelet.

## Added
-  Portforward package brought from kubernetes repository (`/internal/kubernetes/portforward`)
- Helpers to support PortForward http requests from API Server

## Issues involved
This should close #170 , **LIQO provider** has already an implementation in place for this feature, you can look at it [here](https://github.com/liqotech/liqo/blob/master/pkg/virtualKubelet/reflection/workload/podns.go)


node/api/portforward.go
```diff
@@ -0,0 +1,116 @@
...
+// WithPortForwardCreationTimeout sets the creation timeout for a container exec stream
+func WithPortForwardCreationTimeout(dur time.Duration) PortForwardHandlerOption {
+	return func(cfg *PortForwardHandlerConfig) {
+		cfg.StreamCreationTimeout = dur
+	}
+}
+
+// HandleContainerExec makes an http handler func from a Provider which forward ports to a container
+// Note that this handler currently depends on gorrilla/mux to get url parts as variables.
+func HandlePortForward(h PortForwardHandlerFunc, opts ...PortForwardHandlerOption) http.HandlerFunc {
```
pires, line 62: Please, verify comments match function names.

pires (approved):
Overall LGTM but:
- It would be great to have some form of end-to-end testing proving the feature;
- I would add a notice that SPDY is deprecated and the community has been looking - for a long while - for supporting HTTP/2 https://github.com/kubernetes/kubernetes/issues/7452#issuecomment-1490565720

pires:
@aka-somix, on the first bullet, please, take a look and report back for our own knowledge. Maybe even add a comment in the code detailing your findings. On the second bullet, yes, the TODO/notice is enough for me.

helayoty:
> 

@windyear, thanks for pointing this out. Could you please open a PR in this repo so we can review/approve it?

ssoBAekiL:
@pires , @helayoty in order to implement the e2e test for port forwarding we also require the piece of code developed by @windyear so to be able to proceed we see two possible paths:

- First ours and his PR are merged and we implement e2e in a new PR
- We merge windyear's commint into our branch and then all of the changes are brought in our PR

What do you think?

pires:
The latter.

helayoty (approved):

PR #1103, opened 2023-04-07
Support Kind in Makefile check

kannon92:
I was trying to play around with `virtual-kubelet` and I see that there is a check for kubectl context against docker-desktop and minikube.  

I tend to like to use `kind` so I am attempting to use it for virtual-kubelet testing.  

This is my first issue I discovered when running `make`.  

Generally, kind has a context that is the following `kind-$KIND-CLUSTER-NAME$` so I do a regular expression check against context.

### DIFF
--- a/Makefile.e2e
+++ b/Makefile.e2e
@@ -1,3 +1,10 @@
+
+# skaffold checks for kubectl context
+# For minikube, docker-for-desktop and docker-desktop the context matches above names
+# If one wants to use kind, kind gives context names based on the cluster-name
+# But as of now they match the following syntax: kind-*
+# The first check verifies that this is a kind kubernetes context
+# Second check verifies the other ones
 .PHONY: skaffold.validate
 skaffold.validate: kubectl_context := $(shell kubectl config current-context)
 skaffold.validate:
--- a/Makefile.e2e
+++ b/Makefile.e2e
@@ -1,7 +1,10 @@
 .PHONY: skaffold.validate
 skaffold.validate: kubectl_context := $(shell kubectl config current-context)
 skaffold.validate:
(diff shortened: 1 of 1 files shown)


Makefile.e2e
```diff
@@ -1,7 +1,10 @@
 .PHONY: skaffold.validate
 skaffold.validate: kubectl_context := $(shell kubectl config current-context)
 skaffold.validate:
-	@if [[ ! "minikube,docker-for-desktop,docker-desktop" =~ .*"$(kubectl_context)".* ]]; then \
+
+	@if [[ "$(kubectl_context)" =~ .*"kind".* ]]; then \
+		true; \
+	elif [[ ! "minikube,docker-for-desktop,docker-desktop" =~ .*"$(kubectl_context)".* ]]; then \
```
pires, line 7: Wouldn't it work if we just added `kind` to this list of words?

pires (commented):
TBH I don't recall why we even added this check, maybe to avoid people messing w/ non-local clusters, eg production.

Makefile.e2e
kannon92: Nah, kind gives you a context name based on the name of your cluster.

default cluster name would be ‘kind’ and context would be ‘kind-kind’. But if someone names it diffeeenrly than the test breaks.

I moved it to a regular expression check for this reason.  I think it’s always kind-* if you are using a kind cluster. 

My bash isn’t that great either so I’m open for a better way of structuring this code.

pires: Could you, please, describe your rationale as a comment in the code? That would be helpful.

pires (approved):
LGTM

PR #1143, opened 2023-08-04
Add error handler to recreate virtual node when it's deleted

zhenqxuMSFT:
Currently the virtual node is created during virtual kubelet start up. We hit a situation that the virtual node might be deleted by kubernetes control plane due to lost heartbeat. The virtual node won't be recovered until we restart the virtual kubelet.

This PR adds an error handler to nodeController so it has a chance to recreate the node when the virtual node is gone.

### DIFF
--- a/node/nodeutil/controller.go
+++ b/node/nodeutil/controller.go
@@ -15,7 +15,6 @@ import (
 	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"github.com/virtual-kubelet/virtual-kubelet/node"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


zhenqxuMSFT:
@pires @cpuguy83 can I get some reviewers assigned please? Thanks!

cpuguy83:
We had this in the past and removed it.
I think it is best to leave this as opt-in.

zhenqxuMSFT:
@cpuguy83 May I know the reason behind?

zhenqxuMSFT:
@cpuguy83 I changed the code to be opt-in. Could you help review it?

cpuguy83 (approved):

PR #1145, opened 2023-08-19
feat: Add methods to get length of PodController queues

jfernandez:
This PR adds three new methods to the PodController struct. These methods return the current length of the SyncPodsFromKubernetes, DeletePodsFromKubernetes, and SyncPodStatusFromProvider queues, respectively. These methods will provide valuable insights into the queue sizes, which can be used for monitoring and performance tuning.

At Netflix, we are working on tuning the PodController rate limiter and we need insights into the queue size.

PR #1162, opened 2023-09-15
Expose queue unprocessed and items being processed len

jfernandez:
This change is a follow-up to https://github.com/virtual-kubelet/virtual-kubelet/pull/1145 based on our experience using the exposed queue len for observability at Netflix.

The existing `Len()` func is useful for instrumentation, but its current implementation includes unprocessed items and those currently being processed. As a result, it's challenging to discern from our metrics whether we're hitting a rate limit or simply viewing the number of items being processed.

To address this I've added separate getter functions to expose the count of unprocessed items in the queue and the items currently under processing separately. We intend to create separate metrics for these values.

### DIFF
--- a/internal/queue/queue.go
+++ b/internal/queue/queue.go
@@ -288,6 +288,24 @@ func (q *Queue) Len() int {
 	return q.items.Len() + len(q.itemsBeingProcessed)
 }
 
+// UnprocessedLen returns the count of items yet to be processed in the queue
+func (q *Queue) UnprocessedLen() int {
+	q.lock.Lock()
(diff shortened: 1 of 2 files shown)


helayoty (approved):
LGTM 🚢 . Thanks for your contribution.

helayoty:
It seems this PR needs to be rebased.

jfernandez:
@helayoty rebased!

pires (approved):

jfernandez:
@helayoty rebased again. Could you please merge?

PR #1185, opened 2024-01-12
feat: support SPDY subprotocol version v5 (K8s API >= v1.29.0)

### LINKED ISSUE #1184: Remote Exec/Attach causes virtual-kubelet to panic with K8s API >=v1.29.0
### Environment summary

Provider: Liqo

Version: latest master, v1.10.0

K8s: 1.29.0

### Issue Details

Using K8s API version >=1.29.0 the virtual-kubelet panic (nil pointer dereference) when we run Exec or Attach in a remote container. 

The problem arises as the new API introduces a new SPDY subprotocol version ("v5.channel.k8s.io") which the [createHTTPStreamStreams](https://github.com/virtual-kubelet/virtual-kubelet/blob/ce8a0ee8bd68f9c4c46a60a3db5e8e6e6c247839/internal/kubernetes/remotecommand/httpstream.go#L150) function does not handle in the switch case. This causes the handler to be nil and panic when it invokes the `waitForStreams()` function.

Possible solutions:
- handle and implement the new v5 protocol
- fallback to an existing protocol version if there is no exact version match

fra98:
This PR bumps the K8s API to version 1.29.0 and fixes a related bug causing virtual-kubelet to panic (nil pointer dereference) when we run Exec or Attach in a remote container. More info about the issue on #1184.

Implementation: the PR adds the v5 protocol handler (it is implemented like v4, as no new changes concern HTTP streams).

Fixes #1184


helayoty:
@fra98  Thank you for your PR. Would you please rebase and solve the conflict?

go.mod
```diff
@@ -1,12 +1,12 @@
 module github.com/virtual-kubelet/virtual-kubelet
 
-go 1.20
+go 1.21
```
smritidahal653, line 3: Is the go version upgrade required for the fix? If not, it'd be better to separate into a different PR

helayoty (approved):
LGTM 🚀. Thanks for your contribution.

Issue #1032, opened 2022-09-01
Add an interface or a library to enables providers implement downward API


helayoty:
_Originally posted by @cpuguy83 in https://github.com/virtual-kubelet/virtual-kubelet/issues/999#issuecomment-1233336562_

pires:
Also see #123 and #998

tskillian:
I can take this - not super familiar with the code base (yet) but I can at least open a PR taking a stab at it and can iterate from there.

antoinetran:
Hi @tskillian , any news on if the interfaces for providers are created yet? :)
I do some dev for one provide (interlink), and have such a need to get the service account as projected volume.

tskillian:
No, but I have [this PR](https://github.com/virtual-kubelet/virtual-kubelet/pull/1258) open to at least let providers take ownership of downward API resolution. That should let you do whatever it is you need to do.

antoinetran:
> No, but I have [this PR](https://github.com/virtual-kubelet/virtual-kubelet/pull/1258) open to at least let providers take ownership of downward API resolution. That should let you do whatever it is you need to do.

@tskillian ok thanks for your answer! After reading the PR, I think up until now, virtual kubelet has parts of downward API implemented only for environment variables, and your PR is about skipping what virtual kubelet is actually doing. But nothing is done for downward API used in volume (like service account).

So my understanding is, there is no library nor helper (yet) related to downward API yet, but each provider can still do whatever in their own code.

tskillian:
Ah I understand now - yea my understanding is also that the VK will not attempt any resolution of the downward API in volumes. My main concern is that if there's a downward API that isn't supported or can't be handled by VK's internal logic, the provider never gets called. For example, if a pod uses a node IP (not supported by VK), it'll just sit pending while scheduling to the VK indefinitely.

PR #1255, opened 2024-11-12
Make field reference resolution best effort

tskillian:
Today, VK tries to resolve field references for the provider before calling CreatePod. However, if it doesn't know how to resolve a particular field reference, VK core errors out and never calls CreatePod. I think this is a bug - here is the behavior seen by application owners if they see this issue hit:

1) workloads are running fine
2) add unsupported field reference (e.g. status.hostIP)
3) workloads will schedule as usual on VK, but pods will now sit in pending state indefinitely since the pods are assigned a node

Ultimately I think [this](https://github.com/virtual-kubelet/virtual-kubelet/issues/1032) is the right thing to do to offload much of the effort of resolving these field references based on provider needs, but with or without that I still think the current behavior should be considered a bug and fixed.

### DIFF
--- a/internal/podutils/env_internal_test.go
+++ b/internal/podutils/env_internal_test.go
@@ -45,6 +45,8 @@ const (
 	envVarName3 = "CHO"
 	// envVarName4 is a string that can be used as the name of an environment value.
 	envVarName4 = "CAR"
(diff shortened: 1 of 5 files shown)


tkillian-sc:
Just FYI, waiting on a ticket with my company's legal team about signing the CLA

PR #1258, opened 2024-12-13
Allow providers to skip VK downward API resolution

tskillian:
Currently, VK attempts to resolve downward APIs before calling the provider's CreatePod or UpdatePod implementation. However, there are valid reasons a provider may want to do downward API resolution themselves.

This is a logical first step to expose downward resolution utilities to providers as discussed [here](https://github.com/virtual-kubelet/virtual-kubelet/issues/1032)


pires (commented):
Can you, please, add tests that validate the new logic?

tskillian:
@pires added a UT, basically just proved that no downward API resolution is attempted on the pod write path when the new configuration is set

pires (approved):

PR #1295, opened 2025-04-26
Track pods by UID instead of only namespace/name

saleemrashid:


node/uid.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (k *podUIDKey) ObjectName() string {
+	return objectName(k.Namespace, k.Name)
+}
+
+func (k *podUIDKey) String() string {
+	return k.ObjectName() + "/" + string(k.UID)
+}
+
+type uidProviderWrapper struct {
```
pires, line 39: Please, provide a comment explaining why this exists.

node/uid.go
```diff
@@ -0,0 +1,51 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type podUIDKey struct {
+	Namespace string
+	Name      string
+	UID       types.UID
+}
+
+func newPodUIDKey(pod *corev1.Pod) podUIDKey {
```
pires, line 16: I think we should protect this code from taking a nil pointer.

node/podcontroller.go
```diff
@@ -245,11 +274,18 @@ func NewPodController(cfg PodControllerConfig) (*PodController, error) {
 		return nil, pkgerrors.Wrap(err, "could not create resource manager")
 	}
 
+	var provider PodUIDLifecycleHandler
```
pires, line 277: Please, add a comment explaining why. It will help us navigate this code in the future, especially when working to remove the non-UID Pod lifecycle handler.

pires (commented):
LGTM!

node/podcontroller.go
dciangot: Also, related to this, I see the PodControllerConfig still require PodLifecycleHandler, so to my understanding (and also at my direct experience trying this), to use the UID version GetPod and GetPodStatus are still needed to be included although a dummy once. 

Is this the wanted behavior? We might need to document accordingly then.

dciangot (commented):
Just tried it on [interLink](https://github.com/interlink-hq/interLink) code base, and definetely looks good. 
Not breaking the old behavior. Just a small comment you can find inline for the new UID implementation flow.

node/podcontroller.go
saleemrashid: @dciangot This is addressed in the latest commit, there is a `UIDBasedProvider` function that can be used to wrap a `PodUIDLifecycleHandler` with the necessary stubs to be a `PodLifecycleHandler`. You can see it used in the mock provider (which no longer has `GetPod` and `GetPodStatus` methods)

In the future, we'd want to invert this and require a `PodUIDLifecycleHandler` and `PodNotifier` directly, but for now we will retain backwards compatibility.

pires:
e2e tests seem to be timing out. Were you able to run them successfully locally?

PR #1298, opened 2025-05-20
omit secret/configmap informers when `skipDownwardAPIResolution` is enabled

alec-rabold:
It appears these informers aren't used when `skipDownwardAPIResolution` is enabled. This would help save on unused memory. 

TODO: testing

### DIFF
--- a/internal/manager/resource.go
+++ b/internal/manager/resource.go
@@ -15,6 +15,8 @@
 package manager
 
 import (
+	"fmt"
+
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	corev1listers "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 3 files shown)


pires:
FYI I haven't reviewed because it's still in draft.

alec-rabold:
Thanks for the nudge @pires; made some updates and verified that the unit tests + e2e suite passes

internal/manager/resource.go
```diff
@@ -59,15 +76,24 @@ func (rm *ResourceManager) GetPods() []*v1.Pod {
 
 // GetConfigMap retrieves the specified config map from the cache.
 func (rm *ResourceManager) GetConfigMap(name, namespace string) (*v1.ConfigMap, error) {
+	if rm.configMapLister == nil {
+		return nil, fmt.Errorf("configmap lister is not enabled")
+	}
```
alec-rabold: Maybe it would be better to panic for these, open to either

pires:
Can you please rebase from the latest master branch?

PR #1309, opened 2025-06-16
allow pod controller config overrides

alec-rabold:
Allow overriding pod controller config options

Current use-case is for `SyncPodsFromKubernetesRateLimiter`+`SyncPodsFromKubernetesShouldRetryFunc` -- specifically to infinitely retry for certain errors instead of stopping at 20 attempts (e.g. similar to `ImagePullBackoff`)

### DIFF
--- a/cmd/virtual-kubelet/internal/commands/root/root.go
+++ b/cmd/virtual-kubelet/internal/commands/root/root.go
@@ -114,34 +114,37 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
 		return err
 	}
 
-	cm, err := nodeutil.NewNode(c.NodeName, newProvider, func(cfg *nodeutil.NodeConfig) error {
-		cfg.KubeconfigPath = c.KubeConfigPath
-		cfg.Handler = mux
-		cfg.InformerResyncPeriod = c.InformerResyncPeriod
+	cm, err := nodeutil.NewNode(c.NodeName, newProvider, func(o *nodeutil.NodeOptions) {
+		o.NodeConfigOpts = append(o.NodeConfigOpts,
+			func(cfg *nodeutil.NodeConfig) error {
+				cfg.KubeconfigPath = c.KubeConfigPath
+				cfg.Handler = mux
+				cfg.InformerResyncPeriod = c.InformerResyncPeriod
 
-		if taint != nil {
-			cfg.NodeSpec.Spec.Taints = append(cfg.NodeSpec.Spec.Taints, *taint)
-		}
-		cfg.NodeSpec.Status.NodeInfo.Architecture = runtime.GOARCH
-		cfg.NodeSpec.Status.NodeInfo.OperatingSystem = c.OperatingSystem
+				if taint != nil {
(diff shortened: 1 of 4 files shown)


pires:
This breaks code badly :-(

pires (dismissed):
This breaks code badly :-( Please, fix.

pires:
@alec-rabold did you end up using these changes in your provider? It seems the changes are incomplete and it would be great to consider this for inclusion soon 🙏🏻

alec-rabold:
> @alec-rabold did you end up using these changes in your provider? It seems the changes are incomplete and it would be great to consider this for inclusion soon 🙏🏻

hey @pires sorry for the delay, we ended up not using this specific config but I do think it would still be beneficial; I'll spend some time today fixing this up to pass the unit/e2e tests

alec-rabold:
updated! (and verified the unit tests and e2e suite passes) I don't love the nested options style I've added, but should work

njuCZ:
kindly ping, what's the status of this PR? This PR would be helpful to customize rate limit and decrease the latency when there is a lot of pods to handle. @pires

pires:
I need to digest this further as it breaks the current API.

cmd/virtual-kubelet/internal/commands/root/root.go
```diff
@@ -114,34 +114,37 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
...
-	},
-		nodeutil.WithClient(clientSet),
-		setAuth(c.NodeName, apiConfig),
-		nodeutil.WithTLSConfig(
-			nodeutil.WithKeyPairFromPath(apiConfig.CertPath, apiConfig.KeyPath),
-			maybeCA(apiConfig.CACertPath),
-		),
-		nodeutil.AttachProviderRoutes(mux),
-	)
+	cm, err := nodeutil.NewNode(c.NodeName, newProvider, func(o *nodeutil.NodeOptions) {
```
pires, line 117: An alternative would be to impl a `nodeutil.NewNodeWithOpts` that takes `*nodeutil.NodeOptions` while keeping `nodeutil.NewNode` and `*nodeutil.NodeConfig` marked deprecated. A ctor in `NodeOptions` could be able to take from a `NodeConfig` so to make it easier to migrate code. Thoughts?

tskillian (commented):
I think this is a good change. My one significant concern is that it's not backwards compatible: repurposing the public `NodeOpt` type breaks existing callers.

Would a non-breaking approach work instead, by threading the pod-controller overrides through `NodeConfig` and leaving `NodeOpt` and the existing helpers untouched? Roughly:

```go
// in NodeConfig
PodControllerConfigOpts []func(*node.PodControllerConfig) error

func WithPodControllerConfigOverrides(fns ...func(*node.PodControllerConfig) error) NodeOpt {
    return func(cfg *NodeConfig) error {
        cfg.PodControllerConfigOpts = append(cfg.PodControllerConfigOpts, fns...)
        return nil
    }
}
```

Then apply `cfg.PodControllerConfigOpts` in `NewNode` just before `node.NewPodController`. Same feature, no changes to `NodeOpt` or the existing helpers, and `root.go` wouldn't need touching.

PR #1423, opened 2026-05-16
e2e: optional bypass for kubectl context allowlist

### LINKED ISSUE #1003: Improve e2e setup
e2e tests currently have some hard-coded things for minikube and some specific docker contexts (e.g. docker-desktop) that makes it difficult to run in other environments.

It would be really awesome to have this be more generic and easier to get going.

singhvishalkr:
The e2e test suite required the kubectl context to be one of a few specific values (minikube, docker-for-desktop, docker-desktop, or kind-*). I ran into this when trying to run the tests against a k3d cluster and had to manually patch the Makefile.

This removes the hard-coded context restriction. Any kubectl context is now accepted by default. For users who want to restrict which contexts are allowed (to avoid accidentally running against a production cluster), there are two new environment variables:

- `ALLOWED_CONTEXTS` - comma-separated list of allowed context names (supports wildcards like `kind-*`)
- `SKIP_CONTEXT_CHECK` - set to 1 to bypass context validation entirely

I also updated the README to mention kind and other Kubernetes distributions, and documented the new environment variables.

Closes #1003


pires:
Please, ensure your fork branch is up-to-date.

Looking at your changes, there are reasons why we have been limiting to certain well-known context names, namely protecting oneself from pointing VK e2e at production clusters or clusters that simply don't support VK join mode(s). That said, defaulting to allow all is breaking w/ what's been in place for a few years now without solid reasoning. If anything, I think there could be an env var that ignores such a check.

singhvishalkr:
Addressed the above: default behavior again matches the legacy allowlist (minikube, docker-for-desktop, docker-desktop, and contexts whose name matches kind-style semantics). Opt out only with **VK_E2E_ALLOW_ANY_CONTEXT=1** for intentionally non-standard dev clusters. See pushed head on this branch plus Makefile.e2e / \	est/e2e/README.md\ notes.

pires:
You need to update your branch so we don't have all these commits.

singhvishalkr:
Updated the branch as requested.

It is now based on `master` with one focused commit instead of carrying unrelated main-branch history. The diff is limited to `Makefile.e2e` and `test/e2e/README.md`.

Verification: `git diff --check` passes locally. I could not run Go tests on this machine because the `go` executable is not installed.

PR #1424, opened 2026-05-22
fix: ignore resolved env var ordering when comparing pods

immanuwell:
Small fix for a noisy update path.

When VK resolves env from `envFrom`, the provider compare can treat the same env set as different if the order comes back shuffled. Then we call `UpdatePod` again for no real spec change, kinda noisy.

Repro:
1. Create a pod that gets env via `envFrom`.
2. Have `GetPod` return the same env vars in a different order.
3. Reconcile the same pod again.
4. Before this patch VK calls `UpdatePod`.
5. After this patch it does not.

Why this is real:
Providers can round-trip env through unordered forms. `virtual-kubelet/openstack-zun` already flattens container env into `map[string]string` on create, so order loss is a real class of behavior, not just a toy test.

Tests:
- added a regression test for the reordered env case
- `go test ./...`
- `go test -race ./node/... ./internal/podutils/...`

### DIFF
--- a/node/pod_test.go
+++ b/node/pod_test.go
@@ -306,6 +306,48 @@ func TestPodNoSpecChange(t *testing.T) {
 	assert.Check(t, is.Equal(svr.mock.updates.read(), 0))
 }
 
+func TestPodNoSpecChangeWithResolvedEnvOrderDifference(t *testing.T) {
+	svr := newTestController()
+	svr.resourceManager = testutil.FakeResourceManager(testutil.FakeConfigMap("default", "pod-env", map[string]string{
+		"ALPHA": "1",
+		"BETA":  "2",
+	}))
+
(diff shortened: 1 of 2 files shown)


node/pod_test.go
```diff
@@ -670,3 +712,7 @@ func newPodSpec() corev1.PodSpec {
 		},
 	}
 }
+
+func boolPtr(v bool) *bool {
```
pires, line 716: This is a nit but I still want to ask if https://pkg.go.dev/k8s.io/utils@v0.0.0-20260507154919-ff6756f316d2/ptr#To could be used instead - it's already a dependency.

pires: Ping 🙏🏻

pires:
Merged in #1431

PR #1425, opened 2026-06-09
fix: cancel exec/attach context when the streaming connection closes

Raikerian:
A kubectl exec/attach aborted by a hard client disconnect (a crash, network partition, or SIGKILL: anything without a graceful in-band stream half-close) never cancelled the server-side exec. `HandleContainerExec` rooted the context in `context.TODO()`, and since the streaming connection is hijacked, net/http does not cancel `req.Context()` when a hijacked connection drops. The remote process ran until pod deletion and kept its streaming session open the whole time.

This roots the exec/attach context in `req.Context()` and, once the streams are set up, watches the connection's `CloseChan()` (which fires on both a client disconnect and the stream idle-timeout), cancelling the context when it closes. The cancellation propagates into the provider's exec/attach call so it can stop the remote process. The watcher goroutine returns as soon as the exec/attach returns, so it doesn't leak.

`ServeExec`/`ServeAttach` keep their original signatures: cancellation is opt-in via a runtime `executor.(interface{ Cancel() })` / `attacher.(interface{ Cancel() })` assertion, so an adopter enables it by implementing `Cancel()` on the value it already passes, with no API break and no context threaded through the call. The connection's `CloseChan` is likewise read through a type assertion rather than by widening the `context.conn` field, so the WebSocket streaming path (whose connection has no `CloseChan`) still compiles. That path already observes client disconnects through stream closure, so leaving the watcher inert there is correct.

This fixes a leak in [macOS-vz-kubelet](https://github.com/agoda-com/macOS-vz-kubelet) (the project which I maintain), a virtual-kubelet provider that runs each pod inside a macOS VM and serves exec/attach by running the command in the guest over an SSH session. Without a cancel signal the provider had no way to know the client was gone, so an aborted exec left the guest process and its SSH session alive until the pod was deleted, and a run of aborts exhausted the guest sshd's session limit and blocked further exec and stats on that VM. With the context cancelling on disconnect, the provider signals the remote process and closes the session as soon as the client drops, which reaps the process and frees the session slot.

### DIFF
--- a/internal/kubernetes/remotecommand/attach.go
+++ b/internal/kubernetes/remotecommand/attach.go
@@ -41,13 +41,17 @@ type Attacher interface {
 // ServeAttach handles requests to attach to a container. After
 // creating/receiving the required streams, it delegates the actual attachment
 // to the attacher.
-func ServeAttach(w http.ResponseWriter, req *http.Request, attacher Attacher, podName string, uid types.UID, container string, streamOpts *Options, idleTimeout, streamCreationTimeout time.Duration, supportedProtocols []string) {
+func ServeAttach(w http.ResponseWriter, req *http.Request, attacher Attacher, cancel func(), podName string, uid types.UID, container string, streamOpts *Options, idleTimeout, streamCreationTimeout time.Duration, supportedProtocols []string) {
 	ctx, ok := createStreams(req, w, streamOpts, supportedProtocols, idleTimeout, streamCreationTimeout)
 	if !ok {
 		// error is handled by createStreams
 		return
 	}
 	defer ctx.conn.Close()
+
+	execDone := make(chan struct{})
+	defer close(execDone)
+	go watchClose(closeChan(ctx.conn), execDone, cancel)
 
 	err := attacher.AttachToContainer(podName, uid, container, ctx.stdinStream, ctx.stdoutStream, ctx.stderrStream, ctx.tty, ctx.resizeChan, 0)
 	if err != nil {
--- a/internal/kubernetes/remotecommand/attach.go
+++ b/internal/kubernetes/remotecommand/attach.go
@@ -41,7 +41,7 @@ type Attacher interface {
 // ServeAttach handles requests to attach to a container. After
 // creating/receiving the required streams, it delegates the actual attachment
 // to the attacher.
-func ServeAttach(w http.ResponseWriter, req *http.Request, attacher Attacher, cancel func(), podName string, uid types.UID, container string, streamOpts *Options, idleTimeout, streamCreationTimeout time.Duration, supportedProtocols []string) {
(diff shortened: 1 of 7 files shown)


internal/kubernetes/remotecommand/attach.go
```diff
@@ -41,14 +41,18 @@ type Attacher interface {
 // ServeAttach handles requests to attach to a container. After
 // creating/receiving the required streams, it delegates the actual attachment
 // to the attacher.
-func ServeAttach(w http.ResponseWriter, req *http.Request, attacher Attacher, podName string, uid types.UID, container string, streamOpts *Options, idleTimeout, streamCreationTimeout time.Duration, supportedProtocols []string) {
+func ServeAttach(w http.ResponseWriter, req *http.Request, attacher Attacher, cancel func(), podName string, uid types.UID, container string, streamOpts *Options, idleTimeout, streamCreationTimeout time.Duration, supportedProtocols []string) {
```
pires: This is an API breaking change that I'd prefer to avoid. Couldn't this entire change become implicit in, say, https://github.com/agoda-com/virtual-kubelet/blob/main/internal/kubernetes/remotecommand/httpstream.go#L75?

Raikerian: Fair request. Constraint though: the cancel has to reach the provider's `exec`/`attach` call, and that `ctx` lives on the `Executor`/`Attacher` (node/api), not the context stream bundle. Wiring it purely in the bundle won't propagate cancellation, the bundle never reaches the provider.

(These are all internal/ symbols, so not a downstream break either way, but agree on keeping this file close to the k8s copy)

Two ways to drop the `cancel func()` param and restore the original `ServeExec`/`ServeAttach` signatures:

1. Make Executor/Attacher ctx-aware (`ExecInContainer(ctx, ...)`), matching current upstream k8s `remotecommand. ServeExec` derives `WithCancel(req.Context())` internally and passes `ctx` down. Converges with upstream, but touches the two interface methods.
2. Keep every signature as-is, opt-in the cancel via `executor.(interface{ Cancel() })` in `ServeExec`, same type-assert idiom already used for `CloseChan`. Cancel stays fully implicit.

Leaning 1 since upstream already moved there, but happy with 2 if you'd rather not touch the interfaces at all. Preference?

pires: Decisions, decisions. Design-wise I prefer (1) but (2) seems to deliver all without requiring the adopter to pass the right `Context` to the respective calls. I'm divided but I will go with my first instinct and avoid breaking the API, so let's go w/ (2), please. Thank you!

Raikerian: Done

pires (approved):
<3 thanks a ton

PR #1428, opened 2026-07-02
Remove deprecated `wsstream` package from `k8s.io/apiserver`

enrichman:
While attempting to bump our provider's Kubernetes dependency to v1.36, we encountered an issue with `k8s.io/apiserver/pkg/util/wsstream`.

This package is currently used in `internal/kubernetes/portforward` and `internal/kubernetes/remotecommand`. However, it was deprecated in [v0.28](https://pkg.go.dev/k8s.io/apiserver@v0.28.0/pkg/util/wsstream) (!) and completely removed in v0.36. As the documentation states, the correct package to use instead is `k8s.io/streaming/pkg/httpstream/wsstream`. This package however needs Go 1.26 and to avoid too many changes we can still use the one in the `k8s.io/apimachinery/pkg/util/httpstream/wsstream`.

To keep the changes strictly to what is necessary and updates some minor issues, I have done the following:

- Replaced the `wsstream` import path from `apiserver` to the `apimachinery` package.
- Updated Go version handling in the CI workflow to use `go.mod`.
- Updated the Go base image from 1.24 to 1.25 in the `Dockerfile`.
- Updated the Minikube version to v1.38.1 and the Kubernetes version to 1.35 in the CI workflow.


enrichman:
@pires I have rebased/updated the PR, and provided a proper description.

Let me know if that's ok!

cpuguy83:
Is the go1.26 update a requirement for the newer streaming package?

enrichman:
> Is the go1.26 update a requirement for the newer streaming package?

Yes, but I was giving a look at this and it seems the same package is available in the older `apimachinery` package.

I'll give it a try, so we can maybe avoid all the other bumps.

enrichman:
@cpuguy83 I have now used the `wsstream` available in the `apimachinery` package. This doesn't require any bumps to the 1.36 dependencies, nor the Go version. This is safer, so it's probably ok to release a patch version with this.

I am running the tests on our provider to check if the bump to v1.36 of the Kubernetes deps is feasible: https://github.com/rancher/k3k/pull/979

pires:
I would argue Go 1.27 is around the corner and 1.25 EOL will follow closely. Also, if we tag a minor release that states the breaking change in build toolchain, and assuming no bug fixes get in the same release - and instead follow a prior patch release - we should be fine if we are to decide to drop 1.25 support.

.github/workflows/ci.yml
```diff
@@ -25,7 +22,7 @@ jobs:
         uses: actions/checkout@v7
       - uses: actions/setup-go@v6
         with:
-          go-version: ${{ env.GO_VERSION }}
+          go-version-file: go.mod
```
pires, line 25: TIL ❤️

pires (approved):

PR #1441, opened 2026-09-10
node: reject an unnamed node from the API server instead of caching it

tskillian:
## Problem

`NodeController` caches the node as the API server returned it in `serverNode`. That cache is the only source of the name used to address every subsequent node and lease request, and it is also the input from which the next status update is built:

```go
providerNode := n.serverNode.DeepCopy()        // node.go
nodes.Get(ctx, nodeFromProvider.Name, ...)     // node.go
nodes.Patch(ctx, nodeFromProvider.Name, ...)   // node.go
c.leaseClient.Get(ctx, node.Name, ...)         // lease_controller_v1.go
```

If an object without a name ever reaches `serverNode`, every request that could replace it is addressed by the name it is missing. There is no external entry point, so the controller cannot recover.

`updateStatus` assigned the result of `updateNodeStatus` unconditionally, and `updateNodeStatus` returns a nil error whenever `Patch` does — so a response that decodes into a zero-valued `Node` is cached as though it were valid.

## Why it is unrecoverable rather than noisy

The lease controller surfaces it first, and it fails closed:

1. An empty name is rejected client-side by `rest.Request.Name` before any request is sent.
2. That error is a plain `fmt.Errorf`, not a `StatusError`, so `ensureLease`'s `apierrors.IsNotFound(err)` check does not match and the branch that would recreate the lease is never reached. It returns the error instead.
3. `backoffEnsureLease`'s loop has no retry cap and `sync` has no timeout, so `sync` never returns — and because `Run` drives it through a sequential `wait.UntilWithContext`, **renewals stop entirely**.

Renewal is not merely degraded, because `newLease` copies the name from the cached `latestLease` rather than from the node. So a controller in this state keeps working for as long as some other writer renews the lease, and goes permanently stale the moment that stops.

## Fix

Validate at the boundaries rather than threading the name through every call site:

- `setServerNode` refuses a node that is nil or unnamed, so a malformed response is reported as an error and the caller retries against the last known good node.
- `NewNodeController` rejects an unnamed node up front — the name is the controller's identity and no later step can supply it. (`ensureNode` already could not work without it: its first `updateStatus` fails with a non-`NotFound` error and returns before reaching `Create`, so `GenerateName` was never usable here.)
- `getServerNode` refuses to hand out an unnamed node, so the lease controller cannot be given a node it can only fail on.

This restores self-healing: rejecting the bad value leaves a usable name in the cache, so the next status tick succeeds.

## Observed

A lease that stopped being renewed while the process stayed otherwise healthy, with the lease controller logging `Unexpected error getting lease: resource name may not be empty` every 14 seconds for three days without recovering. Node heartbeats continued only while a second writer renewed the lease.

## Tests

`go test ./node/...` passes. Three tests added:

- `TestNewNodeControllerRequiresNamedNode` — fails against `master`, which accepts a node with an empty name.
- `TestSetServerNodeRejectsUnusableNode` — a nil or unnamed node is rejected, the previous node stays cached and addressable, and a well-formed node still updates the cache.
- `TestGetServerNodeRejectsUnnamedCachedNode` — an unnamed cached node is never handed to the lease controller.


node/node.go
```diff
@@ -79,6 +79,14 @@ type NodeProvider interface { //nolint:revive
 // Note: When if there are multiple NodeControllerOpts which apply against the same
 // underlying options, the last NodeControllerOpt will win.
 func NewNodeController(p NodeProvider, node *corev1.Node, nodes v1.NodeInterface, opts ...NodeControllerOpt) (*NodeController, error) {
+	if node == nil {
+		return nil, pkgerrors.New("node must not be nil")
+	}
+	if node.Name == "" {
```
pires, line 85: [As per convention, Kubernetes resource names have to be valid DNS subdomain name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names), so we should use API machinery enforcement code instead, eg:
```go
import (
	"strings"

	apivalidation "k8s.io/apimachinery/pkg/util/validation"
)

(...)

if errs := **apivalidation.IsDNS1123Subdomain(name);** len(errs) > 0 {
	return pkgerrors.WithMessagef(
		pkgerrors.New("invalid hostname"),
		"name %q: %s",
		name,
		strings.Join(errs, "; "),
	)
}
```

This will catch malformed names, including empty.

node/node.go
```diff
@@ -424,9 +432,30 @@ func (n *NodeController) updateStatus(ctx context.Context, providerNode *corev1.
...
+// status update is built, so it can only be repaired by a request that is
+// itself addressed by it. An object that is nil or has no name is therefore
+// rejected rather than cached: the caller reports the error and retries against
+// the last known good node, instead of permanently addressing requests that
+// cannot be built.
+func (n *NodeController) setServerNode(node *corev1.Node) error {
+	if node == nil {
+		return pkgerrors.New("cannot cache a nil server node")
+	}
+	if node.Name == "" {
```
pires, line 452: Same comment as above.

node/node.go
```diff
@@ -437,6 +466,12 @@ func (n *NodeController) getServerNode(_ context.Context) (*corev1.Node, error)
 	if n.serverNode == nil {
 		return nil, pkgerrors.New("Server node does not yet exist")
 	}
+	if n.serverNode.Name == "" {
```
pires, line 469: Same comment as above.

pires (commented):
Nice catch 👏🏻 

I suggest two things:
- rely on API machinery DNS subdomain validation for node name
- create an helper to reduces duplication of code, that takes a node and an error message as an args:
  1. checks node nilness
  1. does the name validation
  ```go
  func validateNodeName(node *corev1.Node, errorMsg string) error {
    if node == nil {
		return nil, pkgerrors.New("node must not be nil")
	}
    if errs := **apivalidation.IsDNS1123Subdomain(name);** len(errs) > 0 {
	  return pkgerrors.WithMessagef(
		  pkgerrors.New(errorMsg),
		  "name %q: %s",
		  name,
		  strings.Join(errs, "; "),
	  )
    }
  }
  ```

node/node.go
tskillian: Done in dfc91046 — switched to `apivalidation.IsDNS1123Subdomain` and pulled the three checks into a single `validateNodeName(node, errorMsg)` helper, so all three sites share it.

Two notes on details:

- I used `pkgerrors.Errorf` rather than `WithMessagef(New(errorMsg), ...)`. The latter renders as `name "x": <errs>: invalid hostname`, which puts the context last; `Errorf` gives `<subject>: invalid node name "x": <errs>`. Happy to switch if you prefer the wrapped form.
- Worded it "invalid node name" rather than "invalid hostname", since a node's name and its hostname are related but not the same field.

Your suggestion also caught a real problem in the tests. `testNode` derived names from `t.Name()`, which contains a `/` for subtests, so `TestNodeRun/WithLease` was building a node named `testnoderun/withlease` — not a valid DNS subdomain. The fake clientset doesn't validate node names, so it had gone unnoticed; `TestNodeRun` failed as soon as the real validation went in. Flattened the separator in the same commit.

node/node.go
tskillian: Done in dfc91046 — this site now calls the shared `validateNodeName` helper.

node/node.go
tskillian: Done in dfc91046 — also on the shared `validateNodeName` helper now.

One thing worth naming explicitly for this site: `getServerNode` is called on every lease renewal and every status update, so it now runs the regex on each read, re-checking an invariant that `setServerNode` and the constructor already enforce at write time. A 253-char regex a few times a second is negligible, so I've left it in for defence-in-depth — it means a future write path that bypasses the setter still can't wedge the lease controller. Mentioning it only so the cost is a deliberate choice rather than an accident; say the word if you'd rather keep the read path a cheap nil/empty check.

pires (approved):
Thanks a lot!

